PR #2, opened 2023-11-03
fix typo in GravityModel

Ibrassow:


duncaneddy (changes requested):
Appreciate the PR! This is a good catch and fix.

I'd request one additional change before merging which is to update the variable on [Line 167](https://github.com/duncaneddy/brahe/blob/5525bcbf7119ba742b6afadf8b3459d637d6fff4/brahe/orbit_dynamics/gravity.py#L167) as well. That is where the initialization variable is set to `True` during load.

Ibrassow:
Done!

duncaneddy:
Thank you appreciate the contribution!

PR #4, opened 2024-04-30
Fix pydantic in master

northshaur:
Really a dirty fix to make it usable with new fast api


brahe/frames.py
```diff
@@ -89,7 +90,7 @@ def polar_motion(epc:Epoch) -> np.ndarray:
     # Compute transformation and return
     rpm = _sofa.Pom00(xp, yp, _sofa.Sp00(_constants.MJD_ZERO, epc.mjd(tsys="TT")))
```
duncaneddy, line 91: Comment out the SOFA calls in this and above so we don't do the computation, if we're doing the hack let's get the benefit of making it run faster.

brahe/frames.py
```diff
@@ -51,7 +51,8 @@ def bias_precession_nutation(epc:Epoch) -> np.ndarray:
     # Compute transformation and return
```
duncaneddy, line 51: See comment below

duncaneddy (changes requested):

PR #56, opened 2025-11-13
Add nalgebra-lapack dependency

duncaneddy:
# Pull Request

## Description

This PR adds nalgebra-lapack as a crate dependency with different built backends depending on platform. This PR is primarily experimental to confirm that the library still builds on all platforms with the dependency added.

### Added
- Added dependency on nalgebra-lapack to enable usage of linear algebra routines in crate.


duncaneddy:
Welp. Going to leave this open as a reminder to find a good way to bring in Lapack as part of the library, but this isn't the way if the windows build takes over 2 hours.

Issue #268, opened 2026-04-13
[Bug]: Wheel distributions omit bundled Blue Marble texture used by default Plotly 3D plots


Mtrya:
### Pre-submission Checklist

- [x] I have searched existing issues to ensure this bug hasn't already been reported
- [x] I have checked the [documentation](https://duncaneddy.github.io/brahe/) for relevant guidance
- [x] I can reproduce this bug with the latest version of Brahe

### Description

`load_earth_texture("blue_marble")` assumes the Blue Marble PNG is bundled with the Python package, but standard wheel builds do not include that asset.

### Language

Python

### Python Reproduction Code

```python
import brahe as bh
from brahe.plots.texture_utils import load_earth_texture

img = load_earth_texture("blue_marble")

# Or via the Plotly path (which defaults to "blue_marble" texture)
# bh.plot_trajectory_3d([...], backend="plotly")
```

### Rust Reproduction Code

```rust

```

### Expected Behavior

The packaged Blue Marble texture should be available in wheel installs, and `load_earth_texture("blue_marble")` should work

### Actual Behavior

```
> python main.py
Traceback (most recent call last):
  File "/path/to/main.py", line 4, in <module>
    img = load_earth_texture("blue_marble")
  File "/path/to/.venv/lib/python3.13/site-packages/brahe/plots/texture_utils.py", line 181, in load_earth_texture
    texture_path = get_blue_marble_texture_path()
  File "/path/to/.venv/lib/python3.13/site-packages/brahe/plots/texture_utils.py", line 56, in get_blue_marble_texture_path
    raise FileNotFoundError(
    ...<2 lines>...
    )
FileNotFoundError: Blue Marble texture not found at /path/to/.venv/lib/python3.13/site-packages/data/textures/world.topo.200410.3x5400x2700.png. This may indicate a corrupted installation.
```

### Brahe Version

1.3.3

### Python Version (if applicable)

3.13

### Operating System

Linux (Other)

### Additional Context

Relevant code:

- brahe/plots/texture_utils.py: get_blue_marble_texture_path() resolves data/textures/world.topo.200410.3x5400x2700.png
- brahe/plots/trajectory_3d.py: Plotly defaults to blue_marble
- pyproject.toml: maturin include only lists brahe/_brahe.pyi and brahe/py.typed

I built a wheel locally with the same maturin build flow used in CI/release and inspected the archive contents. The wheel
contained the Python modules and extension module, but did not contain the Blue Marble texture asset.

[...]

PR #269, opened 2026-04-13
Package Blue Marble texture for wheel installs

Mtrya:
# Pull Request

## Description

This change updates the texture lookup in `brahe.plots.texture_utils` to load the asset via `importlib.resources`, adds the packaged data module and image to the distribution configuration, and adds a test that checks the resource is present as an installed package asset.

## Changelog
- Package the Blue Marble texture in Python distributions so `get_blue_marble_texture_path()` works from installed wheels
- Load the bundled texture via `importlib.resources`

## Changed
- Add `brahe.plots.data` as the home for packaged plotting resources
- Add a test covering Blue Marble resource packaging

## Related Issue

#268 

## Note to Reviewers

The Blue Marble resource seems to be a quite unusual case that sits in the middle:
- it's not protected by Rust's embedding pattern, so it's not in `brahe/_brahe.so` (and it shouldn't)
- it's meant to be a built-in asset, not user-supplied or cache-download

It's probably the only instance of this specific failure mode, but I didn't check too closely.

### DIFF
--- a/brahe/plots/data/__init__.py
+++ b/brahe/plots/data/__init__.py
@@ -1,0 +1,1 @@
+"""Packaged plot resources."""
--- a/brahe/plots/texture_utils.py
+++ b/brahe/plots/texture_utils.py
@@ -5,6 +5,7 @@ Handles loading packaged textures and downloading/caching external texture data.
 """
 
 import zipfile
+from importlib.resources import files
 from pathlib import Path
 from typing import Optional
 import httpx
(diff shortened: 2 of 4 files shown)


duncaneddy:
Appreciate finding the issue. Makes sense given I only generally build from source when developing locally and the CI run does as well.

I could have sworn that I had setup a similar download-on-first-use approach to reduce package size, but guess I just imagined it.

Right now it looks like github CI is failing (though doesn't seem like something that changed in the repo). Need to investigate that to confirm we don't have regressions, but after I fix that I'll get this merged.

Appreciate the clear issue, description, and investigation as always.

PR #276, opened 2026-04-16
Make pyo3 and numpy optional dependencies behind the python feature

markusz:
## Motivation

`brahe` ships Python bindings via PyO3. Until now, `pyo3` and `numpy` were unconditional `[dependencies]`, so any downstream Rust project that added `brahe` to its `Cargo.toml` unconditionally pulled in the entire PyO3 + numpy stack even if it only wanted the pure-Rust API.

This has three practical costs for pure-Rust consumers:
- **Compile time:** PyO3 and its transitive dependencies add measurable build time.
- **Dep graph size:** Consumers inherit all PyO3 transitive deps in their lockfile.
- **Build environment fragility:** Some environments (cross-compilation targets, minimal containers, embedded toolchains) don't have Python headers available. A mandatory PyO3 dep breaks these builds entirely.

## What changed

**`Cargo.toml`**
- `numpy` is now `optional = true`.
- `pyo3` is now `optional = true`. The `features = ["extension-module"]` line is also removed from `Cargo.toml` — maturin already passes `pyo3/extension-module` explicitly via `pyproject.toml` `[tool.maturin] features`, so having it in both places caused linker failures when using `cargo build --features python` directly.
- The `python` feature (which already existed and already gated `mod pymodule` in `src/lib.rs`) is wired up: `python = ["dep:pyo3", "dep:numpy"]`.

**`pyproject.toml`** — no changes. Maturin already passes `features = ["pyo3/extension-module", "python"]`, so Python wheel builds are unaffected.

**CI** — no changes needed. `test_rust.yml` already runs without `--features python`, so it now correctly exercises the no-Python build path. Python CI continues through maturin as before.

## Impact

| Consumer | Before | After |
|---|---|---|
| Pure-Rust (`cargo add brahe`) | Pulled in pyo3 + numpy unconditionally | No pyo3/numpy unless `--features python` |
| Python (pip / maturin) | Unchanged | Unchanged — maturin passes `python` feature |
| Rust with Python bindings | `--features python` had no dep effect | `--features python` activates pyo3 + numpy |

The public Rust API surface is unchanged. No astrodynamics logic was modified.

## Verification

```
cargo build                          # ✅ succeeds, pyo3/numpy absent from dep tree
cargo tree | grep -E "pyo3|numpy"    # ✅ empty
cargo build --features python        # ✅ succeeds, pyo3/numpy present
cargo tree --features python | grep -E "pyo3|numpy"  # ✅ both appear
cargo test                           # ✅ 4256 pass (1 pre-existing network test failure unrelated to this change)
cargo test --features python         # ✅ 4256 pass (same pre-existing failure, no regressions)
```

## Known follow-up

`crate-type = ["cdylib", "lib"]` is still unconditional. The `cdylib` is only needed for maturin/Python builds; pure-Rust consumers build an unnecessary cdylib artifact. Cargo does not support feature-gating `crate-type`, so a clean fix would require splitting into a workspace (`brahe-core` as `lib`, `brahe-py` as `cdylib`). That's a larger structural change left for a future PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- towncrier release notes start -->
 
+## [Unreleased]
+### Changed
+
+- `pyo3` and `numpy` are now optional dependencies, activated via the `python` feature flag. Pure-Rust consumers that do not need the Python bindings no longer compile these crates. Existing Python users (pip/maturin installs) are unaffected; maturin already passes `--features python` via `pyproject.toml`.
+
 ## [1.3.4] - 2026-04-15
 ### Fixed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ crate-type = ["cdylib", "lib"]
 [dependencies]
 rsofa = "0.5.0"
 nalgebra = { version = "0.34.1", features = ["serde-serialize"] }
-numpy = "0.28.0"
+numpy = { version = "0.28.0", optional = true }
 once_cell = "1.21.3"
 num-traits = "0.2.19"
 regex = "1.12.3"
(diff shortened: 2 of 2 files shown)


duncaneddy:
Thank you for the PR. Definite improvement to avoid pulling in unnecessary dependencies into the Rust build.

Normally the `CHANGELOG.md` entry shouldn't be manually generated, but instead it gets automatically generated from PR's that follow the standard PR template (when opened via github) have the necessary fields and will auto-generate the changelog. I haven't been happy with how it's been working lately, so will accept this even though it might need some manual triaging to fix the release process. This is pretty straight forward, but for more complex issues, I would prefer a human-generated PR that follows the template.

PR #280, opened 2026-04-17
[DRAFT][POC] Heap allocations in propagation hot path limit performance

markusz:
**BLUF: Improved heap allocation in propagator improves performance by 5+%, other improvements also possible**

--

This branch explores whether reducing heap allocations in the numerical propagator yields measurable performance gains. 

**It is not intended to be merged as-is, I'm opening for visibility and discussion.**

### Changes
I explored three changes as a minimal PoC for the default Dopri propagator:

- `step_once()` in both propagators: use mem::replace to move x_curr into the integrator instead of cloning it — removes one DVector allocation per step
- WithSTMAndSensitivity mode: removed an unnecessary `result.phi.clone()` (the WithSTM mode already moved correctly here)
- DP54 stage loop: moved ksum and state_i scratch buffers outside the loop + replaced per-stage allocations with `fill(0.0)` + in-place axpy. This removes 12 DVector allocations per step attempt

### Benchmark

I used the `numerical_conservative_24h` benchmark with a 100 samples for increased statistical significance and achieved **a ~5% improvement**.

>   change: [−6.1% −5.2% −4.3%] (p = 0.00 < 0.05)

### Tests

All tests (expect unrelated flaky `space_weather::global::tests::test_initialize_sw`) pass

### Other improvements
This is likely a lower bound. Untouched areas with further potential:

- All other integrators (RK4, RKF45, RKN1210) have the same pattern in their stage loops
- Trajectory storage: `x_curr.clone()` on every stored step is unavoidable with the current ownership model but I feel that changing the allocator model could help
- STM/covariance matrix allocations inside the stage loop when propagating with WithSTM - this is likely expensive
- +Potentially more

Happy to discuss!

### DIFF
--- a/benchmarks/propagator_benchmarks.rs
+++ b/benchmarks/propagator_benchmarks.rs
@@ -14,6 +14,10 @@ use brahe::time::{Epoch, TimeSystem};
 use brahe::traits::{DStatePropagator, SStatePropagator};
 use nalgebra::DVector;
 
+#[cfg(feature = "dhat-heap")]
+#[global_allocator]
+static ALLOC: dhat::Alloc = dhat::Alloc;
+
 fn setup_providers() {
     let eop = FileEOPProvider::from_default_standard(true, EOPExtrapolation::Hold).unwrap();
     set_global_eop_provider(eop);
@@ -43,6 +47,9 @@ fn bench_sgp4_24hour(c: &mut Criterion) {
 }
 
 fn bench_numerical_conservative_24hour(c: &mut Criterion) {
+    #[cfg(feature = "dhat-heap")]
+    let _profiler = dhat::Profiler::new_heap();
+    
     setup_providers();
 
     let epoch = Epoch::from_datetime(2024, 1, 1, 0, 0, 0.0, 0.0, TimeSystem::UTC);
--- a/benchmarks/propagator_benchmarks.rs
+++ b/benchmarks/propagator_benchmarks.rs
@@ -58,7 +58,7 @@ fn bench_numerical_conservative_24hour(c: &mut Criterion) {
     let dstate = DVector::from_column_slice(state.as_slice());
 
     let mut group = c.benchmark_group("propagator_numerical");
-    group.sample_size(10);
+    // group.sample_size(10);
 
     group.bench_function("numerical_conservative_24h", |b| {
         b.iter(|| {
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,6 +63,7 @@ version = "0.28.2"
 optional = true
 
 [dev-dependencies]
+dhat = "0.3.3"
 approx = "0.5.1"
 rstest = "0.26.1"
 serial_test = "3.3.1"
@@ -78,6 +79,8 @@ criterion = { version = "0.8", features = ["html_reports"] }
 ci = []
 manual = []  # For tests requiring network downloads (run with: cargo test --features manual)
 python = ["dep:pyo3", "dep:numpy"]
+dhat-heap = []    # if you are doing heap profiling
+dhat-ad-hoc = []  # if you are doing ad hoc profiling
 
 # Build Profiles
 [profile.dev]
@@ -88,6 +91,10 @@ lto = "off"
 opt-level = 3
 codegen-units = 1
 lto = "fat"
+
+[profile.profiling]
+inherits = "release"
+debug = 1
 
 # Benchmark Profiles
 [profile.bench]
--- a/src/integrators/dp54.rs
+++ b/src/integrators/dp54.rs
@@ -500,23 +500,32 @@ impl DormandPrince54DIntegrator {
                 }
             }
 
-            // Stages 1-6
+            // Stages 1-6 — allocate scratch buffers once and reuse across stages
+            let mut ksum = DVector::<f64>::zeros(self.dimension);
+            let mut state_i = state.clone();
+            let mut k_phi_sum = if compute_phi {
+                DMatrix::<f64>::zeros(self.dimension, self.dimension)
+            } else {
+                DMatrix::zeros(0, 0)
+            };
+            let mut k_sens_sum = if compute_sens {
+                DMatrix::<f64>::zeros(self.dimension, num_params)
+            } else {
+                DMatrix::zeros(0, 0)
+            };
+
             for i in 1..7 {
-                let mut ksum = DVector::<f64>::zeros(self.dimension);
-                let mut k_phi_sum = if compute_phi {
-                    DMatrix::<f64>::zeros(self.dimension, self.dimension)
-                } else {
-                    DMatrix::zeros(0, 0)
-                };
-                let mut k_sens_sum = if compute_sens {
-                    DMatrix::<f64>::zeros(self.dimension, num_params)
-                } else {
-                    DMatrix::zeros(0, 0)
-                };
+                ksum.fill(0.0);
+                if compute_phi {
+                    k_phi_sum.fill(0.0);
+                }
+                if compute_sens {
+                    k_sens_sum.fill(0.0);
+                }
 
                 #[allow(clippy::needless_range_loop)]
                 for j in 0..i {
-                    ksum += self.bt.a[(i, j)] * k.column(j);
+                    ksum.axpy(self.bt.a[(i, j)], &k.column(j), 1.0);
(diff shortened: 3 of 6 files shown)


markusz:
Hm, I've just ran it again, but now I don't observe changes in performance anymore.

Avoiding `.clone()` would be the idiomatic Rust approach, but if it doesn't realiably yield improvements, it's probably not worth it

benchmarks/propagator_benchmarks.rs
```diff
@@ -58,7 +58,7 @@ fn bench_numerical_conservative_24hour(c: &mut Criterion) {
     let dstate = DVector::from_column_slice(state.as_slice());
 
     let mut group = c.benchmark_group("propagator_numerical");
-    group.sample_size(10);
```
duncaneddy, line 61: It's not clear to me why we wouldn't want to keep explicitly setting the sample size?

duncaneddy: To be clear, if we need more samples my preference would just be to increase the number, not leave it as a non-explicit value.

markusz: Oh yeah, I see how that change is misleading. I commented it out because otherwise it didn't pick up the `--sample-size 100` flag in the CLI. I wanted more than 10 samples to reduce noise in the results

This was only for the draft of this MR, if we want to look more closely into this, we'd do it properly

duncaneddy:
Appreciate the investigation and proposed improvement! Definitely see lots of opportunity for improving this area.

For some background the original clone-heavy implementation was an originally coming form translating the SVector implementation to DVectors and just keeping similar semantics in the API for ease of implementation. Another consideration with the original implementation, was trying to minimize dealing with `Option/Result` checking all over the place and try to keep the initial implementation close to the underlying math and maintaining a an API where the outputs (and results) generally look like:

```
x_{t+1}, phi_{t+1}, S_{t+1} = f(t, x, phi, S)
```

I'm actually OK making some breaking changes here to the Rust API to improve performance. For example changing the `state: DVector<f64>` to `state: &DVector<f64>` in the `step/step_with_varmat/step_with_sensmat/step_with_varmat_sensmat` methods. Also accepting incremental improvements if it moves the library closer to idiomatic rust (provide the above desirement for readability/understandability in the user-facing API is maintained) even if they don't yield statistically significant performance improvements.

I also really like the use of `axpy` to vectorize the computation. One thing I've considered is whether it would be worth it to do a more complete `v2.0.0` overhaul of brahe to support package-wide SIMD vectorization by default. That would almost certainly require breaking how Earth Orientation Parameters are handled (they would need to be explicitly passed as an input for all calls).

I'm sure there is a lot we can do still to improve performance, so all for investigating other methods and would be happy to accept the changes.

A few smaller nits just to mention now if building it out in a full PR:
1. It would be appreciated if you used the default PR template (with `### Added / ### Changed / etc` sections) since that is then able to auto-generate the release notes from the change.
1. As you mention, if we're making a change to one integrator, it would be best to make them to all (not just DP54 but the also RK45, and RKN1710 implementations so the standards and memory management practices (and benefits) remain consistent across all of them)

markusz:
Hey @duncaneddy,

thanks for the comprehensive reply (and thanks for building brahe in the first place). 

I spent a couple of hours profiling `compute_spherical_harmonics`, which I found to be one of the code hot paths. I saw a lot of `malloc` for that function, but after trying `Mutex`, `local_thread!` and a couple of other things, I came to the conclusion that `DMatrix` is fine and the compiler can efficiently reuse the allocated memory across calls.

I however found another promising optimization that I'd like to discuss with you. Currently, the C and S coeffs are in the hot path and are calculated for every propagation, when in reality they are constants that can be calculated once or retrieved from the respective tables.

In eac9206dcaee4336aee5a59dff15438e33f3b914, I moved them into precomputed matrixes in `GravityModel`. This change improves the times of the `propagator_numerical/numerical_conservative_24h` benchmark by ~70%

```
Benchmarking propagator_numerical/numerical_conservative_24h: Collecting 100 samples in estimated 70.300 s propagator_numerical/numerical_conservative_24h
                        time:   [689.81 ms 691.55 ms 693.44 ms]
                        change: [−70.463% −70.375% −70.294%] (p = 0.00 < 0.05)
                        Performance has improved.
```

All the existing unit test continue to pass, so I'm somewhat confident that this doesn't break functionality. I'll have to rely on your domain expertise here but mathematically the refactoring should be equivalent to the old code. I can't claim that I fully understand what `Fac` does, but doublechecked (ref. Vallado) the significance of `C` and `S`, which (unless I'm totally of) should indeed be constant across all `n` and `m` 

Let me know what you think

markusz:
In terms of other performance improvements, I have a couple of thoughts, but thus far mostly focused on the propagators as they are most relevant for my use case

- My profiling indicates that `clone()` likely is more a micro-optimization than a real bottleneck
- In the past, I've seen some SIMD success with matrixes that were flattened into flat 1D arrays, but that would require some closer investigation. AI enthusiastically agrees that potential is there, but I think it overstates the actual ROI
- A use case where SIMD could shine is propagating several orbits with fixed step integrators in parallel, but that seems like rather special scenario

It's less of a performance improvement and more like a feature, but I could imagine a third variant `EarthZonal` (or similar) in `GravityConfiguration` that only uses J-terms, but is handcrafted and optimized for performance over accuracy.

This is actually something that I do for my use case: I directly implement J2-J6 math like below. Since it's all just math, the compiler can do it's magic and optimize. It's very fast and accurate to within ~50km over a 24h horizon, compared to the full-fidelity propagator

```
// Values taken from [Vallado]
pub const J2: f64 = 1.08263e-3; // Oblateness

let j2_coeff = -1.5 * J2 * re_r.powi(2) * mu_r2;
let j2_xy_factor = 1.0 - 5.0 * z_r2;
let j2_z_factor = 3.0 - 5.0 * z_r2;

accel.x += j2_coeff * j2_xy_factor * x / r;
accel.y += j2_coeff * j2_xy_factor * y / r;
accel.z += j2_coeff * j2_z_factor * z / r;
```

I don't have a apples to apples comparison, but would expect this to be significantly faster again

duncaneddy:
Hey @markusz,

First, my pleasure on building brahe. I'm glad you're finding it useful! It's exciting to see the people start to find it, adopt it, and find enough value in it to contribute back improvements!

Thanks for the additional performance optimizations!

Everything makes sense on why it works, there's nothing related to the input spacecraft state that affects the computation of the C and S zonal harmonics, so we can just precompute them once when loading the model.

A 70% speedup in numerical propagation is huge! I'd love to get this landed in main and then a minor release. Would you mind terribly either 1) branching the gravity model changes and submitting them separately from the numerical integrator improvements or 2) also apply the numerical integrator improvements to the other integrator types to maintain consistency across implementations? Since they are consistent now I'd like to avoid introducing deviations.

In terms of SIMD, I agree, I think it would be a more complete overhaul of the package where every input is treated in an array-like way, so you can then vectorize fundamental functions (coordination / frame transformations) for speed up. However it does complicate things a lot in terms of worrying about memory layout for all functions, and would be a major overhaul of most things....so I haven't thought it would be worth it given that the package is generally pretty understandable and mirrors the physics today. Generally it seems to me that rayon-based parallelization of individual propagators / threads is the most direct acceleration approach for now. It's something still in the back of my head, I just don't think it's a clear enough win to pursue right now.

In terms of adding precompute J-type gravity acceleration models, I'm all for it! Incrementally building more tools in the brahe-modeling toolbox so that we have different options for different needs would be great. We can make an issue for it to track it as a feature. Though if you'd like to do it, I'd suggest adding the implementation to `orbit_dynamics/gravity.rs`, then modify `GravityConfiguration` to add an option for `J2/J4/J10/J*` gravity fields (then add python bindings and docs). 

This is the [CLAUDE.md](https://gist.github.com/duncaneddy/919d9c44df2f5e81b733398a73d7b361) file I use to help with most of the project now. It's pretty darn good at following the development guidelines and behavior with this structure.

I'll be traveling for the next week and a half so responses might be a bit slower than usual.

markusz:
Sounds good. I'll isolate changes into smaller MRs that follow your preferred structure and keep this branch as a kind of scratch pad to test and bounce ideas

markusz:
Closing this catch-all MR in favor of smaller scoped concrete implementations

PR #302, opened 2026-04-21
WIP: Fast zonal harmonics propagator

markusz:
## What this adds

  - `fast_spherical_zonal_harmonics_accel`: a hand-written implementation of the J2–J6 zonal perturbations (ref. Vallado, Wakker), designed to avoid the matrix operations in `compute_spherical_harmonics` so the compiler can optimize the hot path.
  - `GravityConfiguration::Zonal { degree: ZonalHarmonicsDegrees }` wired into `DNumericalOrbitPropagator`.
  - A regression test comparing the fast path against `SphericalHarmonic(degree=n, order=0)` over 24 h.

## Performance

- 1.7x faster compared to equivalent propagator through existing `compute_spherical_harmonics`
- Benchmarked using added benchmarks on MBP M4 48GB:

### Existing implementation
```
compute_spherical_harmonics(n=6, m=0)
----
propagator_numerical/numerical_spherical_harmonic_j6_equivalent_24hour
                        time:   [451.53 ms 462.30 ms 472.90 ms]
```

### New implementation
```
fast_spherical_zonal_harmonics_accel(n=6)
----
propagator_numerical/numerical_fast_j6_zonal_24hour
                        time:   [274.96 ms 275.87 ms 276.86 ms]
```

## Known open questions (reason this is WIP)

- The fast-zonal and degree-n-order-0 spherical-harmonic propagators agree to < 3 km over 24 h for J6. 
- Root cause not yet identified
- The tolerance in `test_spherical_harmonic_agrees_with_fast_zonal` is intentionally loose for now with a TODO comment.

----

# Pull Request                                                                                           
                                                             
## Description

Adds a fast, closed-form Earth zonal-harmonics gravity model (J2–J6) and wires it into `DNumericalOrbitPropagator` as a new `GravityConfiguration::EarthZonal` variant. Adds a propagator-wide `FrameTransformationModel` setting that lets the force model trade IAU 2006/2000A precision for a rotation-only ECI↔ECEF when speed matters. Tightens `J2_EARTH` and adds `J3_EARTH` - `J6_EARTH` to the public constants, all derived from the EGM2008 fully-normalized Stokes coefficients via `J_n = -C_n,0 * sqrt(2n + 1)`.

### Added                                                                                                
- Added `accel_earth_zonal_gravity` — a hand-rolled closed-form J2–J6 zonal acceleration that mirrors `accel_gravity_spherical_harmonics` with `m = 0` but evaluates ~1.5–2x faster, with agreement to <3 km over a 24 h LEO propagation when both are driven in the same Earth-fixed frame.
- Added `GravityConfiguration::EarthZonal { degree }` variant and supporting `ZonalHarmonicsDegree` enum (`J2`..`J6`) so the numerical propagator can use the fast zonal path directly.                           
- Added `FrameTransformationModel` (variants `FullEarthRotation`, `EarthRotationOnly`) and a `frame_transform` field on `ForceModelConfig`. `EarthRotationOnly` skips precession, nutation, and polar motion for ~1.5x faster ECI↔ECEF rotations at the cost of ~0.07° pole-tilt accuracy; `FullEarthRotation` (default) preserves prior behavior.                                                                      
- Added `J3_EARTH`, `J4_EARTH`, `J5_EARTH`, and `J6_EARTH` constants derived from EGM2008.
- Added Python bindings for `accel_earth_zonal_gravity`, `ZonalHarmonicsDegree`, `FrameTransformationModel`, `GravityConfiguration.earth_zonal(...)`, and the new `J3_EARTH`–`J6_EARTH` constants.                                                                                               
- Added coefficient-derivation tests verifying each `J_n_EARTH` constant matches `-C_n,0 * sqrt(2n + 1)` against the EGM2008 fully-normalized Stokes coefficients.                                                
   
### Changed                                                                                              
- Refined `J2_EARTH` to be derived from the EGM2008 `C_2,0` Stokes coefficient (`1.0826261738522227e-3`) instead of GGM05s. Downstream values change in the 5th significant digit.                                
- Reused a single ECI→body-fixed rotation per dynamics step across gravity, NRLMSISE-00 density, and drag inside `DNumericalOrbitPropagator` rather than recomputing the rotation per force term.


markusz:
@duncaneddy Coming back to the optimized zonal harmonics implementation we discussed in https://github.com/duncaneddy/brahe/pull/280#issuecomment-4274428041, please find a WIP version that shows that this can be significantly faster (~1.7x on my MBP for n=6, m=0)

It does not yet follow the preferred template for MRs and needs some more refinement in general, but I wanted to run it by you first. 

Also please note that currently the new `fast_spherical_zonal_harmonics_accel` implementation disagrees with `compute_spherical_harmonics` by up to 3km over a 24h window. I attribute this to rounding or similar, but am not entirely sure - happy to hear your thoughts as to why this is

duncaneddy:
Much appreciated for the PR! I think the idea is great. Life is going to be a bit scattered this week, but will look and provide thoughts and feedback as soon as able.

A 1.7x speed-up is great!

Mtrya:
I dug into the ~ 3km / 24 h disagreement, and I think the main mismatch appears to be frame handling. The current `GravityConfiguration::Zonal` path evaluates `fast_spherical_zonal_harmonics_accel` directly on the ECI position vector, while the existing spherical harmonic path rotates ECI -> ECEF, computes body-fixed gravity, then rotates the acceleration back to ECI.

So I think the `Zonal` branch should probably do the same frame transformation as the spherical harmonic branch:

1. compute `r_i2b = rotation_eci_to_ecef(epoch)`
2. evaluate the fast zonal formula on `r_i2b * r`
3. add `r_i2b.transpose() * a_body`

There is also a smaller coefficient issue: the hard-coded Wakker `J2..J6` constants are close but not identical to the denormalized EGM2008 zonals, but that only produced about ~18 m over 24 h, so it's not the main 3 km issue.

markusz:
Ha, funny that you just commented that - I was about to write a message with similar findings. 

I see two sources of the discrepancy:

1. **Frame transformation**: I believe you are right that this contributes somehow, but don't fully grok why. My thinking is: For zonal harmonics, only the z-axis should matter and that is the same for ECI and ECEF frames
2. **Legrende polynomial vs. J2**: The approaches should be mathematically equivalent, but there might be floating-point effects where we get numerical drift inside the calculations

Source #2 should be fixable, but for #1 I'm concerned that this negates the performance benefits. We need to benchmark this, but generally, I believe there's a case to be made to accept the discrepancy in the spirit of trading speed over accuracy. 

Overall, what I can say, for my use case a 3km/24h disagreement is well within the error budget, so the change as is would already be useful for me.

Happy to hear @duncaneddy take here

Mtrya:
Hi @markusz , Nice we're converging on the same suspects. On #1, I think your intuition is right that for a simplified case where ECI <-> ECEF is only an Earth-spin rotation about a shared z-axis, the frame transformation wouldn't matter.  But Brahe's `rotation_eci_to_ecef` actually includes the Earth orientation / precession-nutation machinery, so the Earth-fixed pole is tilted relative to the inertial z-axis.

On performance: agreed this needs benchmarking. If the full ECI -> ECEF rotation is too costly, maybe the optimized implementation can work directly with the z-axis vector expressed in ECI instead of rotating into a full body-fixed coordinate frame?

And yeah, I'd defer to @duncaneddy on the boundary here.

duncaneddy:
Hey @markusz @Mtrya!

Sorry for the slow follow up on this. Just getting caught up from my travel.

Super excited to see you both collaborating on this! The growing community of users and contributors is super exciting to see 😃!

@Mtrya's analysis of the error is spot-on here of both the explanation of the source of the error. J2 essentially comes from the equatorial bulge of the Earth, so variations in Earth's tilt due to precession, nutation, and polar motion effects do cause x- and y-axis rotations of the bulge (and other higher order terms) will rotate that mass, which in turn affects the resulting orbital motion.

My preference for path forward would be to add the rotation in. Generally, I've tried to make sure all the default implementations are "correct", and I think saying something is a J2 dynamics propgator (which should normally account for Earth rotation), but then not performing that rotation under the hood sets a risky precedent for having some implementations not be "default-correct" and require that the user understand that we've made certain trades.

I would not expect the Earth Orientation parameter calculation to have a massive performance impact, but still of course need to quantify and see if it does. I think if it does, the way the way I think would be best to handle it would be to introduce a new `ForcdeModelConfig` field that is `FrameTransformationModel` which would look something like:

```
pub enum FrameTransformationModel {
    /// Just model rotation about Earth-z axis
    EarthRotationOnly,
    /// Full rotation model (default)
    FullEarthRotation,
    /// Specific frame realization (supports other planets, new IAU frame resolutions, can be added later)
    FrameRealization(realization),
}
```

Where then `Realization` is something we can add later but should work how `GravityModelSource` works now. We could also condense it and just have `FullRotation(RealizationType)` be the only full rotation with the default `RealizationType` being `IAU2014`, which is the current default frame transformation the package uses (and the latest/most accurate).

This approach of adding the `FrameTransformationModel` allows addressing both considerations. You can turn off the full transformation if you want the added performance, but the default would still be the "correct" full-rotation of the J2-J6 harmonics.

markusz:
Thanks for the analysis @duncaneddy & @Mtrya 

Coming from a Computer Science background, those things are not always intuitively clear to me, but your explanation made it clear where my assumption was flawed.

With that in mind, I gave your proposed solution a pass. 

Here's the results on my local machine (MBP M4 48GB)

| propagator | benchmark | delta pos |
|--------|--------|--------|
| numerical_spherical_harmonic_j6_equivalent_24hour | 457ms | n/a (used as reference) |
| numerical_fast_j6_zonal_full_rotation_24hour | 382ms | <1mm |
| numerical_fast_j6_zonal_earth_rotation_only_24hour | 288ms | <3km |

In other words, the fast zonal harmonics propagator is still faster by ~20% (on my hardware, curios to see your numbers) while controlling for accuracy through `FrameTransformationModel::FullEarthRotation`. When relaxing the requirements for accuracy with `FrameTransformationModel::EarthRotationOnly`, its 59% faster.

LMK what you think

duncaneddy:
Appreciate the changes! Will look in depth tomorrow and benchmark on a couple machines (I can run on an M1 Max 64GB or an AMD EPYC 9554 64-Core Processor 512GB).

duncaneddy:
I just pushed a few changes to clean up this branch to get it merge-ready:

1. Moved zonal constants to constants/physical.rs so they can be available package-wide in the consistent location for global constants.
2. Moved the frame_transformation class out from
`GravityConfiguration::Zonal` to a top-level `ForceModelConfig` setting.
    - This has the added benefit that we can consolidate where we compute frame transformation and only do it once across a few different force models that need it.
    - The performance optimization also makes sense 
3. Added python bindings for zonal constants and new ForceModelConfig
options (gravity and frame transformations)
4. Renamed acceleration function to `spherical_zonal_harmonic_accel`. The "fast" is a nice side effect of the model, but IMO shouldn't be part of the name.
5. Cleaned up docstrings for `spherical_zonal_harmonic_accel` to meet package standards of having all argument / input fields defined.

Going to wait for tests to run/pass then if you're good with the changes merge this and cut a release.

duncaneddy:
These were the benchmark results I got on my M1 Max 64 GB:

| propagator | benchmark |
| ---------- | --------- |
| numerical_conservative_24h | 1316.1 ms | 
| numerical_spherical_harmonic_j6_equivalent_24hour | 834.98 ms |
| numerical_fast_j6_zonal_full_rotation_24hour | 734.24 ms |
| numerical_fast_j6_zonal_earth_rotation_only_24hour | 503.00 ms |

src/orbit_dynamics/gravity.rs
```diff
@@ -151,6 +152,145 @@ pub fn accel_point_mass_gravity<P: IntoPosition>(
...
+/// let a_grav = spherical_zonal_harmonic_accel(r_object, r_central_body, GM_EARTH, 2);
+///
+/// // Magnitude is dominated by the two-body pull (~9.8 m/s²) with a small J2 perturbation.
+/// assert!((a_grav.norm() - 9.8).abs() < 1e-1);
+/// ```
+///
+/// # References
+///
+/// - Vallado, *Fundamentals of Astrodynamics and Applications*, 4th ed., pp. 593.
+/// - Wakker, *Fundamentals of Astrodynamics*, Earth zonal-harmonic coefficients.,
```
markusz: The Wakker coefficients are actually not used anymore. The values are derived by the formula from the EGM2008 table

src/orbit_dynamics/gravity.rs
```diff
@@ -1061,6 +1240,154 @@ mod tests {
...
+        assert_abs_diff_eq!(s_state[4], z_state[4], epsilon = eps_v);
+        assert_abs_diff_eq!(s_state[5], z_state[5], epsilon = eps_v);
+    }
+
+    /// Regression test: with `FullEarthRotation` both propagators transform to the same
+    /// Earth-fixed frame before evaluating the harmonics, so the only remaining difference
+    /// is floating-point round-off between the explicit-formula and Legendre-recursion
+    /// implementations of J2–J6.  Agreement should be well under 1 m after 24 h.
+    ///
+    /// TODO: run and record the actual residual; tighten the tolerance to match.
```
markusz: I think I missed that TODO. I did tighten the tolerance to 1mm, so this is just a remainder

src/orbits/keplerian.rs
```diff
@@ -974,7 +974,7 @@ mod tests {
     #[test]
     fn test_sun_synchronous_inclination() {
         let inc = sun_synchronous_inclination(R_EARTH + 500e3, 0.001, DEGREES);
-        assert_abs_diff_eq!(inc, 97.40172901366881, epsilon = 1e-12);
+        assert_abs_diff_eq!(inc, 97.40179532653737, epsilon = 1e-12);
```
markusz, line 977: That change feels out of place. It's really minor, but might be an AI artifact? Maybe doublecheck

tests/orbits/test_keplerian.py
```diff
@@ -167,7 +167,7 @@ def test_sun_synchronous_inclination():
     vp = brahe.sun_synchronous_inclination(
         brahe.R_EARTH + 500e3, 0.001, angle_format=AngleFormat.DEGREES
     )
-    assert vp == pytest.approx(97.40172901366881, abs=1e-12)
+    assert vp == pytest.approx(97.40179532653737, abs=1e-12)
```
markusz, line 170: Same as in `keplerian.rs`

tests/test_constants.py
```diff
@@ -91,7 +91,23 @@ def test_ECC_EARTH():
 
 
 def test_J2_EARTH():
-    assert brahe.J2_EARTH == 0.0010826358191967
+    assert brahe.J2_EARTH == 1.0826261738522227e-03
```
markusz: Idea: compare them not to hardcoded values but to the result of the formula used to derive those values,i.e. `-1. * <respective C_n,0 value> * f64::sqrt((2. * n + 1.))`

src/constants/physical.rs
```diff
@@ -57,12 +57,53 @@ pub const GM_EARTH: f64 = 3.986004415e14;
...
+///
+/// Derived from the EGM2008 fully-normalized Stokes coefficient C_2,0 via
+/// `J_n = -C_n,0 * sqrt(2n + 1)`.
+///
+/// # References:
+///
+///  1. N. K. Pavlis, S. A. Holmes, S. C. Kenyon, J. K. Factor, *The development
+///     and evaluation of the Earth Gravitational Model 2008 (EGM2008)*, J. Geophys.
+///     Res., 117, B04406, 2012.
+pub const J2_EARTH: f64 = 1.0826261738522227e-03;
```
markusz, line 71: FYI: I was hoping to express those values as concrete, self-documenting formulas, but sqrt in a const context is only available in a [nightly build](https://doc.rust-lang.org/nightly/core/f64/math/fn.sqrt.html), which would be over the top.

Conceptually, I'd find it quite neat to write sth like `pub const J2_EARTH: f64 = -1. * <the C value> * f64::sqrt(2. * 2. + 1.)`

src/orbits/keplerian.rs
duncaneddy: I updated the J2_EARTH constant to match your EGM2008-derived value instead of the older Vallado value. The change has some knock-on effects for hard coded tests. This is one of those.

tests/orbits/test_keplerian.py
duncaneddy: This is a result of the previous change.

src/orbits/keplerian.rs
duncaneddy: I don't love the brittle tests, but there isn't a good way to test the correctness of these values otherwise, so I think the test is generally working to catch the change, but in this case it's intentional.

src/orbit_dynamics/gravity.rs
duncaneddy: Cool. I'll clean it up

src/orbit_dynamics/gravity.rs
duncaneddy: Good to know. I didn't know if you used it for the algorithmic formulation. I'll remove it.

tests/test_constants.py
duncaneddy: Ah, I like this. Confirms the derivation was correct. Will make that change.

src/constants/physical.rs
duncaneddy: Agree on having it be better to define it using sqrt in a const context. There are a few other places where it would be beneficial to use const math, but given I recently worked hard to get the package off nightly and back compiling on the stable branch so I don't particularly want to go back.

Can leave it as a TODO when const math lands in a release.

markusz:
@duncaneddy Thanks for polishing this!

I left some nits and remarks in the code, but overall this LGTM. 

The only thing I'd maybe add is to reflect the improved performance (and the additional ways to trade accuracy for speed) in the documentation. As it is now, it would be easy to miss and I believe there are many potential users out there that will be willing to make similar tradeoffs. 

Also, in the era of Agentic AI, those docs will be consumed by and encoded into all the future LLMs so I feel we should represent brahe well for future users (both humans and machines)

duncaneddy:
I'm going to make one more set of renaming changes to rename the function to `accel_earth_zonal_gravity`. The name `spherical_zonal_harmonic_accel` would imply the function could handle _any_ zonal harmonic model, while the current function with the hard-coded Earth J* coefficients limits the application to just Earth as the central body. Additionally, the accepted `gm` parameter is a bit unneeded since again we know the central body has to be earth. Will rename, fix the test failures, and address your feedback.

PR #310, opened 2026-04-27
Fix Starlink examples CelesTrak API

### LINKED ISSUE #309: [Bug]: Starlink examples still use removed CelesTrak datasets API
## Pre-submission Checklist

- [x] I have searched existing issues to ensure this bug hasn't already been reported
- [x] I have checked the [documentation](https://duncaneddy.github.io/brahe/) for relevant guidance
- [x] I can reproduce this bug with the latest version of Brahe

## Description

Several Starlink examples still call the removed `bh.datasets.celestrak.get_tles_as_propagators()` / `bh::datasets::celestrak::get_tles_as_propagators()` API. The CelesTrak client was moved to the standalone `brahe.celestrak` / `brahe::celestrak` module during the February 2026 CelesTrak client overhaul, and later JOSS Starlink examples were migrated to the new interface, but the regular Starlink examples were missed.

Affected files include:

- `examples/examples/visualizing_starlink.py`
- `examples/examples/starlink_propagation.py`
- `examples/common/starlink_propagation.py`
- `examples/common/starlink_propagation.rs`
- `docs/examples/visualizing_starlink.md`
- `docs/examples/visualizing_gps.md` still mentions the removed helper in prose, despite its embedded GPS code using the current API

The files are easy to miss because the Starlink examples are marked `FLAGS = ["IGNORE"]`, so they are skipped by default example testing.

## Language

Both

## Python Reproduction Code

```python
import brahe as bh

bh.initialize_eop()

# Raises AttributeError because datasets.celestrak no longer exists.
starlink = bh.datasets.celestrak.get_tles_as_propagators("starlink", 60.0)
```

## Rust Reproduction Code

```rust
use brahe as bh;

fn main() {
    bh::initialize_eop().unwrap();

    // Does not compile because bh::datasets::celestrak no longer exists.
    let propagators = bh::datasets::celestrak::get_tles_as_propagators("starlink", 60.0)
        .expect("Failed to fetch Starlink TLEs");
}
```

## Expected Behavior

The Starlink examples should use the current CelesTrak client API, matching the updated JOSS examples and other CelesTrak examples.

Python:

```python
client = bh.celestrak.CelestrakClient()
records = client.get_gp(group="starlink")
propagators = [record.to_sgp_propagator(60.0) for record in records]
```

Rust:

```rust
use brahe::celestrak::CelestrakClient;
use brahe::propagators::SGPPropagator;

let client = CelestrakClient::new();
let records = client.get_gp_by_group("starlink")?;
let propagators: Vec<_> = records
    .iter()
    .filter_map(|record| SGPPropagator::from_gp_record(record, 60.0).ok())
    .collect();
```

## Actual Behavior

Python fails with:

```text
AttributeError: module 'brahe.datasets' has no attribute 'celestrak'
```

Rust fails to compile because `bh::datasets::celestrak` does not exist.

## Brahe Version

1.4.2 local checkout; latest PyPI release checked as 1.4.0 on 2026-04-27

## Python Version (if applicable)

3.13.11

## Rust Version (if applicable)

1.92.0

## Operating System

Linux (Other)

## Additional Context

Git history suggests this is an oversight rather than an intentional compatibility path:

- `9ae3ffcc` moved CelesTrak out of `datasets` and updated many examples, including the GPS Python example, but did not update the regular Starlink example code.
- `f4955970` later updated JOSS Starlink examples from the old API to the new `CelestrakClient` API.
- The remaining stale references are now concentrated in the regular Starlink examples and docs prose.

Mtrya:
# Pull Request

## Description

Updates the remaining Starlink examples and related docs prose to use the current standalone CelesTrak client API instead of the removed `datasets.celestrak.get_tles_as_propagators` helper.

### Fixed
- Fixed Starlink examples and documentation prose that still referenced the removed `bh.datasets.celestrak` / `bh::datasets::celestrak` API.

## Related Issue

Closes #309

## Note to Reviewers

I did not run the ignored Starlink examples end-to-end because they fetch and propagate the full current Starlink catalog. I also did not change the `["IGNORE"]` flags because processing all Starlink satellites is probably too heavy to include in CI.


docs/examples/visualizing_gps.md
```diff
@@ -2,7 +2,7 @@
 
 In this example we'll show how to visualize the orbits of GPS satellites using Brahe. We'll download the latest TLE data for the GPS constellation from CelesTrak, propagate each satellite for one orbit, and create an interactive 3D plot showing their trajectories around Earth.
```
duncaneddy, line 4: This seems like an incorrect textual change since it references Starlink in the GPS visualization example. I think just leave the text as GPS given this is the GPS example then this is good to ship?

duncaneddy (dismissed):
Thank you for the fix given the example broke with the recent changes.

One small textual change then I think this is good to ship!

docs/examples/visualizing_gps.md
Mtrya: I read this sentence as comparing the GPS example to the existing Starlink example, since the link target is `visualizing_starlink.md`. Would you prefer we avoid mentioning Starlink here entirely, or should we keep the comparison and maybe word it more clearly?

duncaneddy: Nevermind. Sorry! Read this too quickly and too jet-lagged. This is right. Thanks for the fix!

PR #372, opened 2026-06-30
Add support for producing Linux arm64 wheels

hectcastro:
# Pull Request

## Description

- Add `ubuntu-24.04-arm` to the OS matrix so that ARM-based wheels are produced.
- Also, update artifact naming logic to ensure no artifact naming collisions exist.

## Changelog

**Fill in at least one section below.** At release time, these entries are aggregated across all merged PRs to produce `CHANGELOG.md` and the GitHub Release notes; each is attributed to its author and PR. Sections follow the [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) convention; see the [versioning policy](https://duncaneddy.github.io/brahe/latest/about/versioning/) for how deprecations and removals flow across releases.

### Added
- Added native Linux `arm64` wheel builds to the latest and release workflows, with platform-specific wheel artifact names to avoid artifact collisions.

### Changed
- 

### Deprecated / Future Warning
- 

### Removed
- 

### Fixed
- 

## Note to Reviewers

Not entirely sure of the best way to test these changes. Will review GitHub Actions workflows that are triggered by this PR to see what validation exists out-of-the-box.


duncaneddy:
Awesome—thanks for adding the additional wheels!

Agreed just adding it is the best (and only) way to really test it. It "just worked" when I added OSX and Windows wheels, so hopefully this is the same.

There's a separate breaking change in upstream dependencies that happened last night that's affecting us as well. I have a fix in https://github.com/duncaneddy/brahe/pull/373, once that works it's way through CI and merged to main, we can update this branch and test: https://github.com/duncaneddy/brahe/pull/373

duncaneddy:
Ok, got that fix in, merged it in and well see it run.

Normally for the rest of the CI process we also run the unit tests against all architectures, but given that we already cover arm architectures through MacOS-aarch64 and have separate Linux (Ubuntu24.04) tests, I'm not sure it's worth the added coverage, which comes at the cost of more CI minutes, pipeline time, and potential flakiness (CI already can cause issues for upstream dependencies though I'm trying to mitigate that as much as possible currently).

duncaneddy:
This worked without issue! I'll cut a release shortly to get out the EOP host fix, and it will get these platforms on pypi too.

hectcastro:
Excellent! Thanks, @duncaneddy!

PR #376, opened 2026-07-02
Native SPICE kernel implementation (replaces ANISE)

duncaneddy:
# Pull Request

## Description

Replaces the ANISE dependency with a brahe-native SPICE kernel implementation: a DAF container parser, Chebyshev segment evaluation (SPK Type 2/3, binary PCK Type 2), an `SPK` reader with cached signed-chain resolution, a `BPCK` orientation reader, and a global multi-kernel registry modeled on the EOP provider pattern. ANISE remains as a dev-dependency only, serving as the validation oracle and benchmark baseline.

**Motivation:** ANISE lacks Orekit-style segment caching and only offers full-state queries, making it slower than Orekit for third-body perturbation loops. The native implementation caches resolved segment chains per (target, center) pair, evaluates with O(1) record lookup over in-memory coefficients, and provides separate position/velocity/state accessors.

The API surface is typed end-to-end: `NAIFId`/`FrameId` enums for body and frame identifiers (raw integers still accepted everywhere), a `SPICEKernel` enum covering every downloadable kernel (DE, JPL planetary satellite ephemerides, binary PCK) with `KernelSource` for bring-your-own paths, and per-body accessors named `*_spice` (e.g. `sun_position_spice`) reflecting that the source is the SPICE kernel system. The outer-planet accessors return true planet body centers via a two-leg DE + satellite-ephemeris-kernel composition, auto-downloading the needed kernel on first use; `*_barycenter_*_spice` variants provide the planetary-system barycenter (used by third-body force models — DE-kernel-only, numerically identical to the classic formulation).

**Validation** (vs ANISE, de440s kernel, matched ET, 12 bodies x {Earth, SSB} x 100 epochs):
- Position: max |Δr| = 9.77e-4 m (< 1 mm; ~1 ULP at outer-body distances)
- Velocity: max |Δv| = 1.82e-11 m/s
- Lunar PCK orientation (moon_pa_de440, frame 31008): exact 0.0 elementwise DCM agreement
- End-to-end epoch path: max 0.81 m, root-caused to the TDB model difference between brahe's 7-term series and hifitime's single-sinusoid model (~20 µs)
- Mars body center vs system barycenter (mar099s two-leg): |Δr| = 0.192 m, |Δv| = 3.8e-5 m/s, consistent with the Phobos/Deimos mass ratio

**Benchmarks** (criterion, de440s, Apple Silicon):

| Case | Native | ANISE | Speedup |
|---|---|---|---|
| Single query, full state (Sun/Moon/Mars rel Earth) | 230–317 ns | 1290–1579 ns | 5.0–5.6x |
| Single query, position-only | 91–107 ns | n/a (no position-only API) | 14.2–15.2x vs ANISE state |
| Sequential 10k epochs, position-only | 1.10 ms | 15.91 ms | **14.5x** |
| Sequential 10k epochs, full state | 3.21 ms | 15.91 ms | 5.0x |
| SPK-direct (no registry) position | 57–77 ns | — | registry overhead ~30 ns |

### Added
- Native SPICE kernel support: `SPK` and `BPCK` readers (DAF container, SPK Type 2/3, binary PCK Type 2), usable standalone or via the global registry
- Global multi-kernel registry: `load_kernel`, `unload_kernel`, `clear_kernels`, `loaded_kernels`; kernels stay resident simultaneously with most-recently-loaded precedence and epoch-aware chain fallback
- `NAIFId` enum (planets, planetary-system barycenters, major moons, `Id(i32)` catch-all) and `FrameId` enum (`MoonPaDe440` = 31008 + `Id(i32)`); all query functions take `impl Into<NAIFId>` / `impl Into<FrameId>`, so raw integer IDs keep working. Python mirrors as `NAIFId`/`FrameId` IntEnums
- `SPICEKernel` enum covering every downloadable kernel — DE (de430–de442s), JPL satellite ephemeris kernels (`mar099`, `mar099s`, `jup365`, `sat441`, `ura184`, `nep097`, `plu060`), and binary PCK (`moon_pa_de440`) — plus `KernelSource` for bring-your-own kernel paths
- `load_common_kernels()` (de440s + moon_pa_de440, ~46 MB) and `load_all_kernels()` (~2.5 GB) pre-initialization helpers
- Generic NAIF-ID ephemeris queries: `spk_position`, `spk_velocity`, `spk_state` (pooled, cross-kernel chaining) and kernel-scoped `spk_position_from_kernel` / `spk_velocity_from_kernel` / `spk_state_from_kernel`
- Per-body kernel-backed accessors `{body}_{position,velocity,state}_spice` for Sun, Moon, Mercury, Venus, Mars, Jupiter, Saturn, Uranus, Neptune, and the SSB. The five outer planets return true body centers via a two-leg DE + satellite-ephemeris-kernel sum (kernel auto-downloaded on first use: mar099s ~68 MB, jup365 ~1.1 GB, sat441 ~662 MB, ura184 ~387 MB, nep097 ~105 MB); `{planet}_barycenter_{position,velocity,state}_spice` variants provide the planetary-system barycenter from the DE kernel alone
- PCK orientation queries with typed attitude returns: `pck_euler_angle` (`EulerAngle`, ZXZ), `pck_euler_angle_and_rates`, `pck_euler_rates`, `pck_quaternion` (`Quaternion`), `pck_rotation_matrix` (`RotationMatrix`), plus the raw `pck_euler_angles`
- `Epoch::spice_et()` — SPICE ephemeris time (TDB seconds past J2000) convenience accessor; `Epoch::seconds_past_j2000_as_time_system` for other time systems
- `datasets::naif::download_spice_kernel` for any known NAIF kernel (DE, satellite ephemeris, binary PCK), exposed in Python as `datasets.naif.download_spice_kernel`
- ANISE validation suite (mm-level matched-ET comparisons, network-gated lunar DCM validation) and native-vs-ANISE criterion benchmarks
- Learn/API documentation for the spice module and four standalone examples, enabled in CI with cached kernels

### Changed
- **Breaking:** third-body ephemeris accessors renamed from `*_de` to `*_spice` (e.g. `sun_position_de` → `sun_position_spice`, `accel_third_body_sun_de` → `accel_third_body_sun_spice`) — the ephemeris source is the SPICE kernel system (DE + satellite ephemeris kernels), not DE alone
- **Breaking:** the five outer-planet accessors return the **true planet body center** (previously the planetary-system barycenter); the barycenter is available via the new `*_barycenter_*_spice` functions, and third-body force models use those (numerically identical to before, no satellite-kernel downloads in propagation)
- **Breaking:** per-body position outputs no longer apply a J2000→ICRF frame-bias rotation. NAIF documents that SPICE "J2000" output from DE kernels is already ICRF-aligned, so the previous rotation introduced a ~23 mas systematic offset (~1.7e4 m at 1 AU)
- **Breaking:** `EphemerisSource::SPK` payload type is `SPICEKernel` (was `SPKKernel`); serde variant names for the DE kernels are unchanged, so serialized force-model configs remain compatible
- Switching between DE kernels no longer reloads global state; both kernels stay resident in the registry

### Removed
- **Breaking:** `SPKKernel` — use `SPICEKernel`
- **Breaking:** `datasets::download_de_kernel` — use `datasets::naif::download_spice_kernel`, which accepts any known kernel name (DE, satellite ephemeris, binary PCK)
- **Breaking:** `set_global_almanac`, `get_loaded_kernel_type`, and `brahe_epoch_to_anise` (ANISE types no longer appear in the public API)
- **Breaking:** ANISE as a runtime dependency (now dev-only). Downstream users relying on brahe's transitive `anise` dependency must add it directly

## Related Issue

Groundwork for #134 (propagation about other central bodies): generic target/center queries and binary PCK orientation are the ephemeris/orientation foundation that work needs.


tests/time/test_epoch.py
```diff
@@ -405,6 +405,26 @@ def test_epoch_to_mjd(eop):
     assert epc.mjd_as_time_system(bh.UTC) == bh.MJD_J2000 - 32.0 / 86400.0
 
 
+def test_seconds_past_j2000_as_time_system(eop):
+    epc = bh.Epoch.from_datetime(2000, 1, 1, 12, 0, 0.0, 0.0, bh.TimeSystem.TT)
+    assert epc.seconds_past_j2000_as_time_system(bh.TimeSystem.TT) == pytest.approx(
+        0.0, abs=1e-9
+    )
+    assert epc.seconds_past_j2000_as_time_system(bh.TimeSystem.TAI) == pytest.approx(
+        -32.184, abs=1e-9
```
duncaneddy: -32.184 should be replaced with the const TAI_TT (or TT_TAI)

tests/spice/test_registry.py
```diff
@@ -0,0 +1,121 @@
...
+    x = bh.spk_state_in_kernel("de440s", bh.NAIF_SUN, bh.NAIF_EARTH, epc)
+    assert x.shape == (6,)
+
+
+def test_spk_position_in_kernel_agrees_with_pooled():
+    """With only de440s loaded, the kernel-scoped query must agree exactly
+    with the pooled registry query."""
+    epc = bh.Epoch.from_date(2025, 1, 1, bh.TimeSystem.UTC)
+    r_pooled = bh.spk_position(bh.NAIF_MOON, bh.NAIF_EARTH, epc)
+    r_in_kernel = bh.spk_position_in_kernel("de440s", bh.NAIF_MOON, bh.NAIF_EARTH, epc)
```
duncaneddy: Instead of `spk_position_in_kernel` it might be a bit more intuitive to name this `spk_position_from_kernel` to make it clear we're getting it from a specific kernel)

src/spice/registry.rs
```diff
@@ -0,0 +1,1252 @@
...
+/// use brahe::spice::{NAIF_MOON, NAIF_EARTH, spk_position_in_kernel};
+/// use brahe::time::{Epoch, TimeSystem};
+///
+/// let epc = Epoch::from_date(2025, 1, 1, TimeSystem::UTC);
+/// let r_moon = spk_position_in_kernel("de440s", NAIF_MOON, NAIF_EARTH, epc).unwrap();
+/// ```
+pub fn spk_position_in_kernel(
+    kernel_name: &str,
+    target: i32,
+    center: i32,
```
duncaneddy: I'd like to reverse the ordering of these arguments to be (kernel name, center, target, epoch). It makes more sense to put the center first to declare "I'm in this position, where is X" as the next question. (unless there is a reason for the opposite.

This convention should be extended to all other similar functions calls that use a similar signature.

src/time/epoch.rs
```diff
@@ -2471,6 +2512,62 @@ mod tests {
...
+            0.0,
+            epsilon = 1.0e-9
+        );
+
+        // TT vs TAI differ by exactly 32.184 s. TT leads TAI, so the TAI
+        // clock reading of the J2000 instant (defined in TT) is earlier,
+        // i.e. negative seconds past J2000 as read on a TAI clock.
+        assert_abs_diff_eq!(
+            epc.seconds_past_j2000_as_time_system(TimeSystem::TAI),
+            -32.184,
```
duncaneddy: Use constant TT_TAI for test.

src/spice/validation.rs
```diff
@@ -0,0 +1,280 @@
...
+            ))
+        .norm();
+        max_dr = max_dr.max(dr);
+        assert!(dr < 1.2, "body {} end-to-end |Δr| = {} m", body, dr);
+    }
+    println!("end-to-end epoch path: max |dr| = {:.3} m", max_dr);
+}
+
+#[test]
+fn test_validation_frame_bias_removal_documented() {
```
duncaneddy: Given this was an incorrect error in the original understanding / implementation, I think we can just drop this validation /test from the repo and just move on with the correct implementation.

src/spice/validation.rs
```diff
@@ -0,0 +1,280 @@
+//! Validation of the native SPICE implementation against ANISE.
```
duncaneddy, line 1: Should the entire contents of this file be in a cfg(test) macro block so that it's not compiled / included in the package implementation since it's internal only and validation?

This should also avoid this file getting improperly counted against coverage requirements (needing coverage) when really it is just added testing.

src/spice/positions.rs
```diff
@@ -1,544 +1,277 @@
...
-
-    let r_m = Vector3::new(
-        r_j2000.radius_km[0] * 1.0e3,
-        r_j2000.radius_km[1] * 1.0e3,
-        r_j2000.radius_km[2] * 1.0e3,
-    );
-
-    Ok(j2000_to_icrf() * r_m)
-}
+body_de_functions!(
```
duncaneddy: We should have both body_*_de functions and body_barycenter_*_de functions. 

A user can reasonably expect that body_*_de will yield the exact position of that body, so for things like outer-planets we should download the relevant system kernel and correct barycenter to planet as well.

We should document that for third-body applications they should generally prefer the barycenter formulation, but we should support both definitions for accession position as well as third-body forces.

src/spice/kernels.rs
duncaneddy: I think we need to support additional Kernel names, mainly the ones listed here (https://ssd.jpl.nasa.gov/sats/ephem/sep.html) as additional options to download and load as well.

I don't think we can do precise body positioning without it.

We should also enable users to bring/define their own

src/spice/registry.rs
```diff
@@ -0,0 +1,1252 @@
...
+use super::spk::{
+    ChainLink, SPK, evaluate_chain_position, evaluate_chain_state, evaluate_chain_velocity,
+    evaluate_with_epoch_fallback, resolve_chain,
+};
+
+// ============================================================================
+// NAIF ID Constants
+// ============================================================================
+
+/// NAIF ID of the Solar System Barycenter.
```
duncaneddy: I think we need to expand the NAIF ID registry to support all the planets, their barycenters, and ideally major moons (https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/req/naif_ids.html#Planets%20and%20Satellites)

It also might make more sense to instead of having this as a set of constants have it as a enum with the barycenters, planets, and major moons (phobos/demios/ganymede/io/europa/etc) as definitions, but then also have the ability to define an arbitrary NAIF_ID to support other downloaded kernels. This also avoids needing to have EVERY object id defined.

src/spice/pck.rs
```diff
@@ -0,0 +1,271 @@
...
+    /// segment reference frame (ICRF for DE440-era kernels).
+    ///
+    /// # Arguments
+    /// - `frame_id`: Body-frame class ID (e.g. 31008 for MOON_PA_DE440)
+    /// - `et`: TDB seconds past J2000
+    ///
+    /// # Returns
+    /// - `(angles, rates)`: `angles = [phi, delta, w]` in [rad];
+    ///   `rates` are their time derivatives in [rad/s]
+    pub fn euler_angles(
```
duncaneddy, line 102: This should return a properly initialized attitude::EulerAngle attitude class to take advantage of those synergies.

We should also support a function to return the orientation as a attitude::Quaternion

src/spice/pck.rs
```diff
@@ -0,0 +1,271 @@
...
+    /// Rotation matrix from the segment reference frame (ICRF) to the
+    /// body-fixed frame: `R = Rz(w) · Rx(delta) · Rz(phi)`.
+    ///
+    /// # Arguments
+    /// - `frame_id`: Body-frame class ID (e.g. 31008 for MOON_PA_DE440)
+    /// - `et`: TDB seconds past J2000
+    ///
+    /// # Returns
+    /// - 3x3 rotation matrix (ICRF to body-fixed). Dimensionless.
+    pub fn rotation_matrix(&self, frame_id: i32, et: f64) -> Result<Matrix3<f64>, BraheError> {
```
duncaneddy: This should be returning a RotationMatrix class not just a matrix or it should be renamed to enable return as both RotationMatrix and Matrix3<f64>.

docs/learn/spice/index.md
```diff
@@ -0,0 +1,201 @@
...
+        --8<-- "./docs/outputs/spice/spice_kernel_registry.py.txt"
+        ```
+
+    === "Rust"
+        ```
+        --8<-- "./docs/outputs/spice/spice_kernel_registry.rs.txt"
+        ```
+
+Known DE kernel names (`de430`, `de432s`, `de435`, `de438`, `de440`,
+`de440s`, `de442`, `de442s`) and the binary PCK name `moon_pa_de440` are
```
duncaneddy: This implies that `moon_pa_de440` is different than the de440 kernel and not contained within it (which I think is wrong)?

docs/learn/spice/index.md
```diff
@@ -0,0 +1,201 @@
...
+downloaded and cached automatically; any other string is treated as a path
+to a local `.bsp` or `.bpc` file. See
```
duncaneddy: I similarly don't think this is right. We need the ability to download other SPK files per my other comment)

docs/learn/spice/index.md
```diff
@@ -0,0 +1,201 @@
...
+
+| Function | Returns | Units |
+|---|---|---|
+| `spk_position(target, center, epc)` | Position of `target` rel. `center` | m |
+| `spk_velocity(target, center, epc)` | Velocity of `target` rel. `center` | m/s |
+| `spk_state(target, center, epc)` | `[x, y, z, vx, vy, vz]` of `target` rel. `center` | m, m/s |
+
+Common NAIF IDs are exposed as constants:
+
+| Constant | NAIF ID | Body |
```
duncaneddy: Add reference to https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/req/naif_ids.html#Planets%20and%20Satellites for full list of NAIF IDs

docs/learn/spice/index.md
```diff
@@ -0,0 +1,201 @@
...
+
+SPK and PCK kernels are parsed once, at `load_kernel` time, into an
+in-memory segment table. Queries take a short-lived read lock on the shared
+registry only to look up the relevant segment chain; the Chebyshev
+evaluation itself runs outside any lock. Repeated queries for the same
+target/center pair reuse a cached chain, so sequential access patterns (e.g.
+a propagator stepping through epochs) avoid re-resolving kernel coverage on
+every call.
+
+## See Also
```
duncaneddy, line 274: Should add links to relevant SPICE Docs:

- https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/req/daf.html
- https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/req/spk.html
- https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/req/pck.html
- https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/req/frames.html

docs/learn/spice/index.md
```diff
@@ -0,0 +1,201 @@
...
+        ```
+
+Known DE kernel names (`de430`, `de432s`, `de435`, `de438`, `de440`,
+`de440s`, `de442`, `de442s`) and the binary PCK name `moon_pa_de440` are
+downloaded and cached automatically; any other string is treated as a path
+to a local `.bsp` or `.bpc` file. See
+[NAIF Ephemeris Kernels](../datasets/naif.md) for kernel sizes and caching
+details.
+
+Registry semantics:
```
duncaneddy: Are these "semantics" or are they behaviors (or properties)?

docs/learn/spice/index.md
```diff
@@ -0,0 +1,201 @@
...
+(e.g. ~290 km for Saturn from Titan, ~230 km for Jupiter from the Galilean
+moons). Sun, Moon, Mercury, and Venus have no
+significant satellites, so their functions' body-center and barycenter
+positions coincide.
+
+## PCK Orientation
+
+`pck_euler_angles` and `pck_rotation_matrix` query body orientation from a
+loaded binary PCK, given a frame class ID. The lunar principal-axis frame
+from the `moon_pa_de440` kernel is registered under frame class ID `31008`
```
duncaneddy: Another piece of feedback was adding an enum for NAIF_ID for bodies, I don't know if it would make sense to introduce a similar one for frames (just the `moon_pa_de440` for now but with support for using arbitrary IDs?)

crates/brahe-py/src/spice.rs
```diff
@@ -0,0 +1,403 @@
...
+
+/// Velocity of a target body relative to a center body from loaded SPK kernels.
+///
+/// The result is expressed in the kernel's inertial frame (ICRF axes; NAIF
+/// labels this "J2000"). If no kernels are loaded, DE440s is loaded
+/// automatically.
+///
+/// Args:
+///     target (int): NAIF ID of the target body (e.g. bh.NAIF_MOON)
+///     center (int): NAIF ID of the center body (e.g. bh.NAIF_EARTH)
```
duncaneddy: Neets to be updated to be (center, target, epoch)

docs/library_api/spice/index.md
```diff
@@ -0,0 +1,89 @@
...
+
+::: brahe.pck_euler_angles
+
+### pck_rotation_matrix
+
+::: brahe.pck_rotation_matrix
+
+## NAIF ID Constants
+
+| Constant | NAIF ID | Body |
```
duncaneddy: Provide as enum definition (?)

examples/spice/spice_kernel_registry.py
```diff
@@ -0,0 +1,47 @@
+# /// script
+# dependencies = ["brahe", "numpy"]
+# FLAGS = ["IGNORE"]
+# ///
+"""
+Load SPICE kernels and query ephemeris data from the global registry.
+
+This example demonstrates the generic NAIF-ID queries (spk_position/velocity/state),
+the kernel-scoped variants, and the per-body *_de convenience functions. Downloads
+the de440s kernel (~33 MB) on first run.
```
duncaneddy: The Github Actions cache should be caching and preserving the de440s kernel so it doesn't need to be downloaded between runs, so we shouldn't need to "ignore" this example.

Similarly for testing locally a one-time cost of the download should be fine.

examples/spice/spice_kernel_registry.rs
```diff
@@ -0,0 +1,75 @@
+//! Load SPICE kernels and query ephemeris data from the global registry.
+//!
+//! This example demonstrates the generic NAIF-ID queries (spk_position/velocity/state),
+//! the kernel-scoped variants, and the per-body *_de convenience functions. Downloads
+//! the de440s kernel (~33 MB) on first run.
+//!
+//! FLAGS = ["IGNORE"]
```
duncaneddy: Same comment as python example, this doesn't need to be ignored.

src/time/epoch.rs
```diff
@@ -1286,6 +1286,47 @@ impl Epoch {
...
+    /// use brahe::time::{Epoch, TimeSystem};
+    ///
+    /// let eop = FileEOPProvider::from_default_file(EOPType::StandardBulletinA, true, EOPExtrapolation::Zero).unwrap();
+    /// set_global_eop_provider(eop);
+    ///
+    /// let epc = Epoch::from_datetime(2000, 1, 1, 12, 0, 0.0, 0.0, TimeSystem::TT);
+    /// let et = epc.seconds_past_j2000_as_time_system(TimeSystem::TDB);
+    /// assert!(et.abs() < 1.0e-3); // TDB and TT agree to < 1.7 ms
+    /// ```
+    pub fn seconds_past_j2000_as_time_system(&self, time_system: TimeSystem) -> f64 {
```
duncaneddy, line 1316: It would also make sense to add a convenience alias `spice_et()` that wrapps this function with passing in the argument as `TimeSystem::TDB` so there is just a simple call to get the proper ET value for an epoch.

This can be used elsewhere in the implementaiton to reduce number of calls to the long name of `seconds_past_j2000_as_time_system`

The comment can be updated to the reflect that passing TimeSystem::TDB` as the argument is equivalent to `spice_et`

src/datasets/naif.rs
```diff
@@ -16,11 +16,21 @@ use std::path::PathBuf;
...
 const NAIF_BASE_URL: &str = "https://naif.jpl.nasa.gov/pub/naif/generic_kernels/spk/planets/";
 
-/// Supported DE kernel names
-const SUPPORTED_KERNELS: &[&str] = &[
+/// Base URL for NAIF generic PCK kernels
+const NAIF_PCK_BASE_URL: &str = "https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/";
+
+/// Supported DE kernel names. `pub(crate)` so `spice::registry` can consult
+/// this single list instead of maintaining its own copy.
+pub(crate) const SUPPORTED_KERNELS: &[&str] = &[
```
duncaneddy: As a similar idea to consolidating NAIF_IDs to an enum, it might make sense to consolidate SupportedKernels to an enum, so that we can use this set, but also users can bring their own or we can download additional kernels from (https://ssd.jpl.nasa.gov/ftp/eph/satellites/bsp/) which contains PCKs for other planetary bodies that would be needed to address the other feedback in this PR.

This could allow for both listing all the major DE/planetary PCK kernels brahe knows about and downloading along with defining a "from_file" or other option for loading more / bring-your-own.

src/datasets/naif.rs
```diff
@@ -40,6 +50,31 @@ fn validate_kernel_name(name: &str) -> Result<(), BraheError> {
...
 }
 
+/// Validate that a binary PCK kernel name is supported and return its filename
+///
+/// # Arguments
+/// * `name` - PCK kernel name to validate (e.g., "moon_pa_de440")
+///
+/// # Returns
+/// * `Result<&'static str, BraheError>` - The kernel's filename if valid, error otherwise
+fn validate_pck_kernel_name(name: &str) -> Result<&'static str, BraheError> {
```
duncaneddy: We we allow bring-your-own this can likely be dropped.

src/spice/daf.rs
```diff
@@ -0,0 +1,428 @@
...
+use crate::utils::BraheError;
+
+/// Size of a DAF physical record in bytes.
+const RECORD_BYTES: usize = 1024;
+/// Words (8-byte doubles) per record.
+const RECORD_WORDS: usize = 128;
+
+/// A single segment descriptor (summary) plus its name.
+#[derive(Debug, Clone)]
+pub(crate) struct DafSummary {
```
duncaneddy: We should `#[allow(non_snake_case)]` and name all types `DAF*` to it has proper capitalization

src/spice/daf.rs
```diff
@@ -0,0 +1,428 @@
...
+            return Err(BraheError::IoError(format!(
+                "Invalid DAF: ID word '{}' does not start with 'DAF/'",
+                id_word
+            )));
+        }
+
+        // Endianness from LOCFMT (bytes 88..96); fall back to an ND sanity
+        // check for pre-LOCFMT files.
+        let locfmt = String::from_utf8_lossy(&bytes[88..96]).trim().to_string();
+        let big_endian = match locfmt.as_str() {
```
duncaneddy: Should the name of this be `is_big_endian` as it's an indicator used elsewhere?

src/spice/mod.rs
```diff
@@ -1,17 +1,24 @@
...
-pub mod almanac;
+pub(crate) mod daf;
 pub mod kernels;
+pub mod pck;
 pub mod positions;
+pub mod registry;
+pub(crate) mod segments;
+pub mod spk;
+#[cfg(test)]
+mod validation;
```
duncaneddy, line 20: I see the entire mod is marked as cfg(test) but I think because the config is applied here coveralls / code coverage isn't properly picking up that this should be ignored

tests/time/test_epoch.py
duncaneddy: Replaced with `bh.TAI_TT` (bc057126).

src/time/epoch.rs
duncaneddy: Replaced with `TAI_TT` (bc057126).

tests/spice/test_registry.py
duncaneddy: Renamed to `spk_*_from_kernel` across Rust, Python, docs, and examples (d8aa492c).

src/spice/registry.rs
duncaneddy: Keeping (target, center, epoch): SPICE `spkezr`, spiceypy, and ANISE are all target-first, and with both args bare integers a swap would silently break code ported from SPICE. Confirmed offline; the docs now state the convention explicitly.

src/spice/validation.rs
duncaneddy: Test removed (76278829).

src/spice/validation.rs
duncaneddy: The module was already `#[cfg(test)]` via mod.rs; added `#[cfg_attr(coverage_nightly, coverage(off))]` and a `.codecov.yml` ignore entry so coverage stops counting it (76278829).

src/spice/positions.rs
duncaneddy: Done: `mars/jupiter/saturn/uranus/neptune_*_de` now return the true body center via a two-leg sum (DE kernel barycenter + satellite-system kernel body offset, auto-downloaded on first use), with new `*_barycenter_*_de` variants preserving the old behavior. Third-body forces use the barycenter variants and the docs state that preference (e83198f3). Measured Mars body-vs-barycenter: 0.192 m.

src/spice/kernels.rs
duncaneddy: Added downloadable satellite-system kernels (mar099, mar099s, jup365, sat441, ura184, nep097, plu060) via the `NAIFKernel` enum; bring-your-own kernels load from any local .bsp/.bpc path (9586c1f2).

src/spice/spk.rs
```diff
@@ -0,0 +1,947 @@
...
+
+    // Walk back target -> center, emitting one link per edge.
+    let mut chain = Vec::new();
+    let mut node = target;
+    while node != center {
+        let prev = parent[&node];
+        // Edge prev -> node. A stored segment (target=node, center=prev)
+        // gives node rel prev directly (sign +1); (target=prev, center=node)
+        // gives the reverse (sign -1).
+        if let Some(segs) = pair_segments.get(&(node, prev)) {
```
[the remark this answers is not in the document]
duncaneddy: Fixed: chain links now merge candidates from both segment orientations in global load order with per-candidate signs, so a later-loaded reversed segment wins as documented. Regression test added (faad4b28).

src/spice/registry.rs
duncaneddy: Added `NAIFId` enum (planets, barycenters, major moons + `Id(i32)` catch-all); all query functions take `impl Into<NAIFId>` so raw ints still work; `NAIF_*` constants removed; Python gets a matching IntEnum (54472d74, e7981091).

src/spice/pck.rs
duncaneddy: Added `pck_euler_angle` returning `attitude::EulerAngle` (ZXZ), `pck_euler_angle_and_rates`, `pck_euler_rates`, and `pck_quaternion`; the raw `pck_euler_angles` accessor is retained (413a3549).

src/spice/pck.rs
duncaneddy: `pck_rotation_matrix` (and `BPCK::rotation_matrix`) now returns `RotationMatrix` (413a3549).

docs/learn/spice/index.md
duncaneddy: The file split is real: `de440.bsp` carries only translational SPK segments, and NAIF distributes the lunar orientation from the same DE440 integration as a separate binary PCK (`moon_pa_de440_200625.bpc`). Added a clarifying sentence to the docs (7a3a308d).

docs/learn/spice/index.md
duncaneddy: Satellite-system kernels are now downloadable by name, and any local .bsp/.bpc path loads directly; docs updated with the full kernel table (9586c1f2, 7a3a308d).

docs/learn/spice/index.md
duncaneddy: Link added (7a3a308d).

docs/learn/spice/index.md
duncaneddy: All four Required Reading links added to See Also (7a3a308d).

docs/learn/spice/index.md
duncaneddy: Reworded to 'Registry behavior' (7a3a308d).

docs/learn/spice/index.md
duncaneddy: Added `FrameId` enum (`MoonPaDe440` + `Id(i32)`) with a Python IntEnum mirror; `pck_*` functions take `impl Into<FrameId>` (54472d74).

crates/brahe-py/src/spice.rs
duncaneddy: Keeping target-first per the registry.rs thread — SPICE convention.

docs/library_api/spice/index.md
duncaneddy: Constants table replaced with `NAIFId`/`FrameId` enum documentation (7a3a308d).

examples/spice/spice_kernel_registry.py
duncaneddy: IGNORE flag removed; CI caches ~/.cache/brahe/naif and the warm-up step pre-fetches de440s, moon_pa_de440, and mar099s (4b87a4cc).

examples/spice/spice_kernel_registry.rs
duncaneddy: Enabled, same as the Python example (4b87a4cc).

src/time/epoch.rs
duncaneddy: Added `Epoch::spice_et()` (Rust + Python), used internally in the registry, with the docstring cross-reference (bc057126).

src/datasets/naif.rs
duncaneddy: Added `NAIFKernel` enum consolidating DE, satellite-system, and binary PCK kernels, plus `KernelSource` for bring-your-own paths; the datasets download layer is table-driven off it via `download_kernel` (9586c1f2).

src/datasets/naif.rs
duncaneddy: Removed — unknown names fall through to path handling, and the download layer is enum-driven (9586c1f2).

src/spice/daf.rs
duncaneddy: Renamed to `DAFFile`/`DAFSummary` with `#[allow(clippy::upper_case_acronyms)]` (01506f6e).

src/spice/daf.rs
duncaneddy: Renamed to `is_big_endian` (01506f6e).

src/spice/mod.rs
duncaneddy: Added `#[cfg_attr(coverage_nightly, coverage(off))]` on the mod declaration and a `.codecov.yml` ignore entry for src/spice/validation.rs (76278829).

.github/workflows/test_examples.yml
```diff
@@ -56,6 +56,18 @@ jobs:
           name: brahe-cache
           path: ~/.cache/brahe/
 
+      # Restored before "Setup test assets in cache" below so the repo's
```
duncaneddy, line 59: It seems to me that the cached copy should win over test_assets/de440s.bsp since that isn't being updated and the github actions cache might be more recent.

brahe/spice.py
```diff
@@ -0,0 +1,122 @@
...
+    pck_euler_angles,
+    pck_euler_angle,
+    pck_euler_rates,
+    pck_euler_angle_and_rates,
+    pck_quaternion,
+    pck_rotation_matrix,
+)
+
+
+class NAIFId(IntEnum):
```
duncaneddy, line 50: Instead of having this enum defiend in python, shouldn't the Enum be implemented through PyO3 binding of a Rust Enum/class to ensure the rust implementation and python implementation stay in sync?

docs/learn/datasets/naif.md
```diff
@@ -1,29 +1,47 @@
...
+| `sat441`  | ~662 MB   | Saturn satellite ephemeris — Titan and mid-size moons (1750-2250) |
+| `ura184`  | ~387 MB   | Uranus satellite ephemeris — Miranda, Ariel, Umbriel, Titania, Oberon (1600-2399) |
+| `nep097`  | ~105 MB   | Neptune satellite ephemeris — Triton (1600-2400) |
+| `plu060`  | ~135 MB   | Pluto-system ephemeris — Charon (1800-2200) |
+| `moon_pa_de440` | ~13 MB | Lunar principal-axis binary PCK (orientation, not SPK) |
 </div>
 
 !!! tip "Choosing a Kernel"
-    For most applications, `de440s` provides a good balance between file size and accuracy. The "s" (small) variants cover a shorter time span but are significantly smaller files.
+    For most applications, `de440s` provides a good balance between file size and accuracy. The "s" (small) variants cover a shorter time span but are significantly smaller files. The satellite-system kernels (`mar099s`, `jup365`, `sat441`, `ura184`, `nep097`) are downloaded automatically the first time a planet body-center `*_de` function (e.g. `mars_position_de`) is called — see [Ephemerides](../../library_api/orbit_dynamics/ephemerides.md).
```
duncaneddy: You can add an admonition tip that using *_barycenter_de functions are preferred for most computations since the position difference from planetary bodies is small, but it eliminates the need to download secondary kernels or do cross-kernel chaining.

docs/learn/datasets/naif.md
```diff
@@ -68,30 +86,38 @@ The first call downloads and caches the kernel. Subsequent calls immediately ret
...
+        bh.datasets.naif.download_kernel("de999")
     except RuntimeError as e:
         print(e)
         # "Unsupported kernel name 'de999'. Supported kernels: de430, de432s, ..."
     ```
 
+In Rust, `download_kernel` takes a `NAIFKernel` enum value directly rather
+than a string, so an unsupported kernel name cannot be passed at all — the
+enum only has the 16 valid variants. To validate a name obtained at runtime
+(e.g. from user input), resolve it through `NAIFKernel::from_name` first:
```
duncaneddy: This statement seems incorrect given that NAIFKernel supports the from_name method and python does as well. I think both could just be consolidated into a single argument.

Similarly why isn't the NAIFKernel enum equivalently exposed in python?

crates/brahe-py/src/orbit_dynamics.rs
```diff
@@ -556,23 +725,1057 @@ fn py_solar_system_barycenter_position_de<'py>(
...
-fn py_ssb_position_de<'py>(
-    py: Python<'py>,
-    epc: &PyEpoch,
-    source: PyEphemerisSource,
-) -> PyResult<Bound<'py, PyArray<f64, Ix1>>> {
+#[pyo3(name = "jupiter_state_de")]
+fn py_jupiter_state_de<'py>(py: Python<'py>, epc: &PyEpoch, source: PyEphemerisSource) -> PyResult<Bound<'py, PyArray<f64, Ix1>>> {
     let kernel = de_kernel_from_py_source(source)?;
-    let r = spice::ssb_position_de(epc.obj, kernel)
+    let x = spice::jupiter_state_de(epc.obj, kernel)
```
duncaneddy: Instead of having all these functions be `state/position/velocity_de` (so `*_de`), be renamed to `*_spice` to make it clear the source of the ephemeris is the `spice` system not `de` since for planetary centers we need additional kernels?

crates/brahe-py/src/orbit_dynamics.rs
```diff
@@ -556,23 +725,1057 @@ fn py_solar_system_barycenter_position_de<'py>(
...
+    let kernel = de_kernel_from_py_source(source)?;
+    let x = spice::mars_barycenter_state_de(epc.obj, kernel)
+        .map_err(|e| exceptions::PyRuntimeError::new_err(e.to_string()))?;
+    Ok(vector_to_numpy!(py, x, 6, f64))
+}
+
+/// Calculate the position of the Jupiter system barycenter in the GCRF inertial frame using NAIF DE ephemeris.
+///
+/// Returns the planetary-system barycenter using only the DE kernel (DE440s or
+/// DE440); no satellite-system kernel is required. This is the position used by
```
duncaneddy: Is `satellite-system kernel` really the best term here for the secondary kernels of other celestial bodies? If so fine to leave it, but if it isn't standard then, "planetary-system kernel" might be more explanatory.

docs/learn/datasets/naif.md
```diff
@@ -1,29 +1,47 @@
 # NAIF Ephemeris Kernels
 
-[NAIF (Navigation and Ancillary Information Facility)](https://naif.jpl.nasa.gov/) is NASA JPL's archive for planetary ephemeris data. Brahe provides functions to download DE (Development Ephemeris) kernels, which contain high-precision position and velocity data for solar system bodies.
+[NAIF (Navigation and Ancillary Information Facility)](https://naif.jpl.nasa.gov/) is NASA JPL's archive for planetary ephemeris data. Brahe provides functions to download DE (Development Ephemeris), satellite-system, and lunar-orientation kernels from the NAIF archive.
```
duncaneddy: Same comment here, "planetary-system kernel" might be more appropriate and more informative a term.

docs/library_api/spice/index.md
```diff
@@ -0,0 +1,166 @@
+# SPICE Kernels
+
+Native SPICE kernel registry: loading/unloading SPK and binary PCK kernels,
+and generic NAIF-ID ephemeris/orientation queries. For the per-body `*_de`
+convenience functions, see the
+[Ephemerides reference](../orbit_dynamics/ephemerides.md).
+
+## Kernel Registry
+
+### load_kernel
```
duncaneddy: Instead of having subheaders for each function just define the function imports since the function headers get added to the TOC by default by mkdocstrings

docs/library_api/datasets/naif.md
```diff
@@ -1,14 +1,22 @@
 # NAIF Functions
 
-Functions for downloading planetary ephemeris kernels from NASA JPL's NAIF archive.
+Functions for downloading planetary DE, satellite-system, and lunar-orientation kernels from NASA JPL's NAIF archive.
 
 All functions are available via `brahe.datasets.naif.<function_name>`.
 
-## download_de_kernel
+## download_kernel
```
duncaneddy: Don't need a subheader for API docs.

docs/library_api/datasets/naif.md
```diff
@@ -1,14 +1,22 @@
...
 
-::: brahe._brahe.naif_download_de_kernel
+::: brahe._brahe.download_naif_kernel
+
+This function accepts any of the 16 kernels brahe knows how to download —
```
duncaneddy: This qualifications should be in the docstring of the function if it is not already. The Rust equivalence statement can be dropped.

docs/library_api/spice/index.md
```diff
@@ -0,0 +1,166 @@
...
+::: brahe.NAIFId
+    options:
+      show_root_heading: true
+      show_root_full_path: false
+
+Full NAIF integer ID listing: [NAIF Integer ID Codes](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/req/naif_ids.html).
+
+**Members:**
+
+| Member | NAIF ID | Body |
```
duncaneddy: It would be better if this could be autogenerated from mkdocstrings so it doesn't need to be manually maintained but I don't know if that's possible, but I think that TimeSystem does it properly in docs/library_api/constants/units.md

docs/library_api/spice/index.md
```diff
@@ -0,0 +1,166 @@
...
+## FrameId
+
+::: brahe.FrameId
+    options:
+      show_root_heading: true
+      show_root_full_path: false
+
+**Members:**
+
+| Member | Frame Class ID | Frame |
```
duncaneddy: Also should have members auto-generated.

src/datasets/naif.rs
```diff
@@ -1,146 +1,136 @@
...
 /// let output = PathBuf::from("/path/to/my_kernel.bsp");
-/// let kernel_path = download_de_kernel("de440s", Some(output.clone())).unwrap();
+/// let kernel_path = download_kernel(NAIFKernel::DE440s, Some(output.clone())).unwrap();
 /// assert_eq!(kernel_path, output);
 /// ```
-pub fn download_de_kernel(name: &str, output_path: Option<PathBuf>) -> Result<PathBuf, BraheError> {
-    // Validate kernel name
-    validate_kernel_name(name)?;
-
+pub fn download_kernel(
```
duncaneddy: This should be `download_spice_kernel` since `kernel` is a pretty generic term in scientific applications.

src/propagators/force_model_config.rs
```diff
@@ -17,7 +17,7 @@ use serde::{Deserialize, Serialize};
 use crate::orbit_dynamics::ParallelMode;
 use crate::orbit_dynamics::gravity::{GravityModelTideSystem, GravityModelType};
 use crate::orbit_dynamics::tides::SolidTideConfig;
-use crate::spice::SPKKernel;
+use crate::spice::NAIFKernel;
```
duncaneddy: Instead of NAIFKernel I think this should be SPICEKernel (but keep NAIFId since that's the term used for the Id lookup)

crates/brahe-py/src/orbit_dynamics.rs
duncaneddy: Renamed the full family to `*_spice` — per-body accessors, barycenter variants, `ssb_*`, and all `accel_third_body_*` wrappers (da5a92f4).

crates/brahe-py/src/orbit_dynamics.rs
duncaneddy: JPL's own term is "planetary satellite ephemerides", so adopted "satellite ephemeris kernel" throughout instead of the nonstandard "satellite-system" (da5a92f4).

docs/learn/datasets/naif.md
duncaneddy: Adopted "satellite ephemeris kernel" (JPL's term) here as well (da5a92f4).

docs/library_api/spice/index.md
duncaneddy: Per-function subheaders removed; the directives stand alone under the section groupings (12047815).

docs/library_api/datasets/naif.md
duncaneddy: Removed (12047815).

docs/library_api/datasets/naif.md
duncaneddy: The `download_spice_kernel` docstring already lists all 16 supported kernel names, so the page prose was redundant — deleted it along with the Rust-equivalence sentence (12047815).

docs/library_api/spice/index.md
duncaneddy: Now autogenerated via the mkdocstrings `members:` pattern from constants/units.md — all 35 NAIFId members render with their values (verified in the built HTML) (12047815).

docs/library_api/spice/index.md
duncaneddy: Same — FrameId members autogenerated (12047815).

src/datasets/naif.rs
duncaneddy: Renamed to `download_spice_kernel` on all three surfaces: Rust `datasets::naif::download_spice_kernel`, Python flat binding, and `bh.datasets.naif.download_spice_kernel` (da5a92f4).

src/propagators/force_model_config.rs
duncaneddy: Renamed `NAIFKernel` → `SPICEKernel`; `NAIFId` and `KernelSource` unchanged (da5a92f4).

PR #378, opened 2026-07-04
Clenshaw-summation spherical harmonic gravity kernel

duncaneddy:
# Pull Request

## Description

**WIP** — Replaces the default spherical-harmonic gravity kernel with a Clenshaw-summation evaluator (Holmes & Featherstone 2002, GeographicLib-style): a single backward sweep over fully-normalized coefficients with an overflow-guard scale factor, computing potential gradient in one pass with no intermediate matrices. The existing Cunningham V/W kernel remains available under `*_cunningham` names; the main `accel_gravity_spherical_harmonics` / `compute_spherical_harmonics` APIs keep their signatures and dispatch Clenshaw-first.

Measured on this branch (M-series host, EGM2008_360): serial Clenshaw is 1.30×/1.51×/1.55× faster than serial Cunningham at n = 50/90/120, with the advantage growing with degree. Accuracy was adjudicated against an independent 40-digit `mpmath` reference: Clenshaw matches it to ~1e-16 relative at high degree/low altitude where Cunningham degrades to ~1e-9 or overflows to NaN (an inherent ceiling of denormalized V/W recursions near degree ≳150 at LEO radii — now converted from silent NaN into a descriptive error).

Remaining before un-drafting: docs pass, full `just check` gate, external + final whole-branch reviews.

### Added
- Clenshaw-summation spherical-harmonic gravity kernel (`GravityModel::compute_spherical_harmonics_clenshaw`, `accel_gravity_spherical_harmonics_clenshaw`), with serial and parallel (order-parallel, bitwise-identical) execution — Rust + Python.
- `GravityModelCoefficients` load configuration (`Clenshaw` default / `Cunningham` / `Both`) with `from_model_type_with_coefficients` / `from_file_with_coefficients` constructors (Rust + Python) and `load_uncached_with_coefficients` (Rust only, matching `load_uncached`), plus `precompute_*` / `drop_*` / `has_*` coefficient-set lifecycle methods — Rust + Python.
- `GravityModel::get_c(n, m)` / `get_s(n, m)` single-coefficient accessors — Rust + Python.
- `accel_gravity_spherical_harmonics_cunningham` convenience function — Rust + Python.
- High-precision reference tests: Clenshaw kernel pinned against independent 40-digit mpmath evaluations of EGM2008_360 to 1e-13 relative at degrees 120–200, with the generator committed as `scripts/generate_clenshaw_gravity_reference.py`.
- The Cunningham kernel returns a descriptive error (instead of silent NaN) on denormalized-recursion overflow.
- `set_global_gravity_model` Python binding (mirrors the existing Rust function).
- Clenshaw-vs-Cunningham benchmark suite in `benchmarks/gravity_benchmarks.rs`.

### Changed
- `compute_spherical_harmonics` and `accel_gravity_spherical_harmonics` now evaluate via the Clenshaw kernel when its coefficients are present (the default), falling back to Cunningham coefficients when those are the only set loaded.
- Gravity model loading now precomputes only Clenshaw coefficients by default (use `GravityModelCoefficients::Both`/`Cunningham` for the V/W kernel); `set_max_degree_order` rebuilds whichever coefficient sets exist.
- The numerical orbit propagator evaluates spherical-harmonic gravity through the Clenshaw kernel (the per-propagator V/W workspace matrices were removed); its per-propagator rotation cache is now an `lru`-crate LRU keyed on stage time.
- `CLENSHAW_PARALLEL_THRESHOLD_NMAX` tuned to 180 from benchmarks (with a TODO to derive it from the execution environment); Cunningham's documented numerical limits added to its rustdoc.
- `GravityModel::compute_spherical_harmonics_with_workspace` and `accel_gravity_spherical_harmonics_with_workspace` — renamed to `compute_spherical_harmonics_cunningham_with_workspace` / `accel_gravity_spherical_harmonics_cunningham_with_workspace` (the Clenshaw kernel needs no workspace). The old allocating Cunningham entry point is now `GravityModel::compute_spherical_harmonics_cunningham` (`compute_spherical_harmonics` keeps its name and signature but dispatches). No deprecation aliases retained.
- Marked additional space-weather tests and integration to issues running locally if celestrak is unavailable.

## Note to Reviewers

A propagator explicitly configured with a Cunningham-only gravity model at high degree can panic mid-propagation (with the descriptive overflow error, not silent NaN) if the trajectory descends below the Cunningham overflow ceiling.

The kernel is a faithful port of GeographicLib v1.52 `SphericalEngine::Value<true, FULL, 1>` restructured into independent per-order inner Clenshaw sweeps (parallelizable, bitwise-deterministic) plus a sequential outer Clenshaw combine. The Clenshaw/Cunningham parity test compares per-position degree grids that omit six (position, degree) combinations where the Cunningham *reference* overflows or is mpmath-provably degraded; those points are pinned against the high-precision reference instead (generator script: `scripts/generate_clenshaw_gravity_reference.py`). A follow-up PR stacked on this branch fixes the repo-wide pre-commit debt (ruff, rustfmt drift, unmarked live-network tests) and is required before the pre-commit hooks can be enabled.


docs/learn/orbital_dynamics/tides.md
```diff
@@ -89,10 +89,40 @@ $\bar{C}_{20}$ with the conventional tide-free convention:
...
+Shared global gravity models (`GravityModelSource.Global`) are read-only across
+every propagator that references them, so their permanent-tide handling is
+resolved once, when the model is installed as the global — not per propagator.
+The propagator trusts the global model's tide system as-is and never mutates
+shared state, so the `PermanentTideConfig` in a `Global`-source force model has
+no effect. Set the tide system on the global model up front, either with the
+convenience setter `set_global_gravity_model_to_tide_system(model, target)` or
+by calling `GravityModel.convert_tide_system` before `set_global_gravity_model`:
+
+=== "Python"
```
duncaneddy, line 103: This should be implemented as code examples that are tested.

docs/library_api/orbit_dynamics/gravity.md
```diff
@@ -11,12 +11,36 @@ Gravity acceleration functions including point-mass and spherical harmonic model
...
 
+::: brahe.accel_gravity_spherical_harmonics_clenshaw
+
+::: brahe.accel_gravity_spherical_harmonics_cunningham
+
 ## Gravity Model Class
 
 ::: brahe.GravityModel
 
+## Gravity Tables
```
duncaneddy: We should have a discussion about whether these are called Gravity Tables or Gravity Model Coefficients. I think the latter is more intuitive.

tests/orbit_dynamics/test_gravity.py
```diff
@@ -466,6 +477,179 @@ def test_set_max_degree_order_validation_exceeds_max(self):
...
+            bh.set_global_gravity_model(original)
+
+    def test_set_global_gravity_model_to_tide_system_converts(self):
+        """Setter converts the model to the target tide system before installing it."""
+        original = bh.get_global_gravity_model()
+        try:
+            # GGM05S is a zero-tide model.
+            ggm = bh.GravityModel.from_model_type(bh.GravityModelType.GGM05S)
+            assert ggm.tide_system == bh.GravityModelTideSystem.ZeroTide
+            c20_zero = ggm.get(2, 0)[0]
```
duncaneddy: Not sure I like this API for accessing coefficients. Instead of knowing what the coefficient is, I think a `get_c(n,m)` could be more ergonomic

crates/brahe-py/src/propagators.rs
```diff
@@ -5177,6 +5177,35 @@ impl PyForceModelConfig {
...
+/// The Rust-side trial evaluation (`DNumericalOrbitPropagator::new`) already hard-errors
+/// if the Cunningham kernel overflows at the initial state, so construction reaching this
+/// point has already succeeded once. This warns about configurations (e.g. GEO altitudes)
+/// where accuracy degrades or a later state could still overflow.
+///
+/// `GravityModelSource::ModelType` is not checked: the underlying model always comes from
+/// the process-wide cache (`GravityModel::shared`), which is guaranteed Clenshaw-only, so
+/// the lacks-Clenshaw condition cannot occur for that source. Only `GravityModelSource::Global`
+/// is checked, via the public `get_global_gravity_model()` accessor.
+fn warn_if_cunningham_only_high_degree(py: Python<'_>, force_config: &propagators::ForceModelConfig) -> PyResult<()> {
```
duncaneddy: Not sure if I like all the complexity being added with warning multiple places. We were using Cuninngham previously without issues so both not sure this is needed and in fact might be overkill, at most this should be a check in NumericalOrbitPropagator or a cached called in the specific Cunningham accel method.

src/propagators/dnumerical_orbit_propagator.rs
```diff
@@ -226,33 +226,23 @@ impl RotationCache {
...
+/// Bundles the [`RotationCache`] under a single lock target so each dynamics
+/// invocation acquires only one mutex. Previously also held the V/W
+/// recurrence buffers for `compute_spherical_harmonics_cunningham_with_workspace`;
+/// the propagator now calls the Clenshaw-first `accel_gravity_spherical_harmonics`
+/// dispatcher directly, which needs no persistent per-propagator workspace
+/// (see the `GravityConfiguration::SphericalHarmonic` arm of `compute_dynamics`),
+/// so this struct is just the rotation cache. Kept as its own type (rather than
+/// using `RotationCache` directly) so future per-propagator scratch state has
+/// a home without another signature change.
 struct DynamicsWorkspace {
```
duncaneddy: It seems instead of having a whole new type for DynamicsWorkspace that we have the opportunity here to do the right thing and just make this a private (internal) property of DNumericalOrbitPropagator unless there is a good reason to keep this as it's own struct.

It seems we could cache (epoch -> transform w/ max eviction size large enough to accomodate the largest (RKN 17/12 stage interator, so something like a size of 30). Look at the LRU crate.

src/propagators/dnumerical_orbit_propagator.rs
```diff
@@ -826,6 +811,57 @@ impl DNumericalOrbitPropagator {
...
+        // Clenshaw tables (an explicit Cunningham-only opt-in via
+        // `GravityTables`) falls back to the Cunningham V/W kernel for
+        // every dynamics evaluation. At high degree and low altitude that
+        // kernel can overflow to a non-finite result (see
+        // `GravityModel::compute_spherical_harmonics_cunningham`'s
+        // `# Numerical limits`) — which would otherwise only surface deep
+        // inside the integrator, far from the configuration that caused it.
+        // Catch it here with one evaluation at the initial state instead.
+        // Models with Clenshaw tables (the default) skip this entirely.
+        if let GravityConfiguration::SphericalHarmonic {
```
duncaneddy: This seems like overly defensive programming. We never had such guards before and only used Cunningham recursion. This might make sense to just remove for now.

src/propagators/dnumerical_orbit_propagator.rs
```diff
@@ -1485,14 +1521,13 @@ impl DNumericalOrbitPropagator {
...
     ) -> SharedDynamics {
-        // Per-propagator scratch state: rotation cache + SH V/W work
-        // matrices, bundled under one Mutex so each dynamics invocation
-        // takes a single lock. Mutex (not RefCell) because
-        // `SharedDynamics: Send + Sync` — even though the dynamics function
-        // is only ever driven by one thread at a time (serialised through
-        // `&mut self` on the propagator), the type system needs the
-        // conservative bound. Lock acquisition on this uncontended Mutex is
-        // ~10 ns; the work it saves dwarfs that.
+        // Per-propagator scratch state: the rotation cache, held under a
```
duncaneddy: Again, given the fact that each dynamics invocation takes a single lock actually having a mutex for this seems overkill. Is there no way to just removed this?

I think there's a chance not since the Jacobian and Sensitive providers could be parallelized and would need shared access to the common rotation cache.

src/propagators/dnumerical_orbit_propagator.rs
```diff
@@ -1698,38 +1720,38 @@ impl DNumericalOrbitPropagator {
                 order,
                 parallel,
             } => {
-                // Use the per-propagator V/W workspace so the spherical-
-                // harmonic recurrence doesn't allocate two `DMatrix`es of
-                // size (degree+2)² on every integrator stage. Cost saved
-                // at 80×80: ~15 µs/stage = ~11 ms per 180-step propagation.
+                // Dispatches to the Clenshaw kernel (the default table
```
duncaneddy: This comment is overly defensive. Just need to say it dispatches to Clenshaw by default.

src/propagators/dnumerical_orbit_propagator.rs
```diff
@@ -8548,6 +8570,104 @@ mod tests {
...
+
+    #[test]
+    #[serial_test::serial]
+    fn test_dnumericalorbitpropagator_construction_cunningham_only_high_degree_leo_fails_trial() {
+        // A Cunningham-only global model at degree 160 and LEO altitude
+        // overflows the denormalized V/W recursion (see
+        // `test_cunningham_high_degree_overflow_errors` in gravity.rs).
+        // The trial-evaluation guardrail must surface that at construction,
+        // not mid-propagation.
+        use crate::orbit_dynamics::gravity::{GravityModelType, GravityTables};
```
duncaneddy: These use statements should be moved to the top of the test section and removed from redundant tests.

tests/propagators/test_numerical_orbit_propagator.py
```diff
@@ -4596,6 +4603,95 @@ def test_numericalorbitpropagator_force_gravity_degree_order():
...
+            GravityModelType.EGM2008_360, GravityTables.Cunningham
+        )
+        set_global_gravity_model(cunningham_only)
+
+        fc = ForceModelConfig.default()
+        fc.gravity = GravityConfiguration.spherical_harmonic(
+            degree=200, order=200, use_global=True
+        )
+
+        with pytest.warns(UserWarning, match="degree > 150"):
```
duncaneddy: Expect test o change/be unneeded given reductions to defensive validation.

src/orbit_dynamics/gravity.rs
```diff
@@ -439,6 +502,59 @@ pub(crate) fn should_parallelize(mode: ParallelMode, n_max: usize) -> bool {
     }
 }
 
+/// Benchmark-calibrated crossover degree for the Clenshaw kernel. At or above
+/// this `n_max`, parallel evaluation of the Clenshaw kernel beats serial on a
+/// typical multi-core host. Machine-approximate; see
+/// `benchmarks/gravity_benchmarks.rs`.
+///
+/// Measured on a 10-core Apple M1 Max. Serial vs parallel median times for
```
duncaneddy, line 510: We should add a TODO comment that this parallel threshold should really be better tuned / set. It will change with both clock speed and number of available threads. These initial settings are just heuristics (slightly informed by one local test), but really it should be set by calculations based on number of available threads and clock-speed.

No action needed aside from  adding TODO.

src/orbit_dynamics/gravity.rs
```diff
@@ -503,6 +619,153 @@ impl GravityModelType {
...
 }
 
+/// Selects which precomputed evaluation tables a gravity model builds at load.
+///
+/// The Clenshaw and Cunningham kernels require different precomputed values
+/// (normalized packed coefficients vs denormalized dense matrices), so each
+/// kernel can only run when its table set is present. The default is
+/// `Clenshaw` — the main evaluation APIs use the Clenshaw kernel.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum GravityTables {
```
duncaneddy: Suggest renaming to `GravityModelCoefficients` or `GravityModelConstants`

src/orbit_dynamics/gravity.rs
```diff
@@ -503,6 +619,153 @@ impl GravityModelType {
...
+/// is no high-degree overflow ceiling) in GeographicLib's triangular packing:
+/// column-major by order `m`, degree `n = m..=n_max` contiguous within a
+/// column. The only auxiliary table is the O(N) square-root table; the
+/// Legendre recurrence factors are computed on the fly from it, which keeps
+/// the per-evaluation memory traffic to one pass over the coefficients.
+///
+/// Reference: Holmes & Featherstone (2002); GeographicLib v1.52
+/// `SphericalEngine.cpp`.
+#[derive(Clone)]
+pub(crate) struct ClenshawTables {
```
duncaneddy: ClenshawTables and CunninghamTables should be similarly renamed.

src/orbit_dynamics/gravity.rs
```diff
@@ -1904,27 +3024,233 @@ mod tests {
...
+            // (position index, n, m)
+            (0, 160, 160), // mid-lat LEO: cunningham returns a non-finite-overflow error
+            (0, 200, 200), // mid-lat LEO: cunningham returns a non-finite-overflow error
+            (1, 120, 120), // equatorial LEO: cunningham loses precision near equator
+            (1, 160, 160), // equatorial LEO: cunningham returns a non-finite-overflow error
+            (1, 200, 200), // equatorial LEO: cunningham returns a non-finite-overflow error
+            (1, 200, 100), // equatorial LEO: cunningham loses precision near equator
+        ];
+
+        let mut worst_rel = 0.0_f64;
```
duncaneddy, line 3199: Instead of having a bunch of degress and orders to test the nskipping ones for specific reasons can we just drop the excluded values up front and never test them, then drop the commend and the continue branch?

src/orbit_dynamics/gravity.rs
```diff
@@ -1904,27 +3024,233 @@ mod tests {
...
+                );
+            }
+        }
+        eprintln!("test_clenshaw_matches_cunningham: worst-case relative residual = {worst_rel:e}");
+    }
+
+    #[test]
+    #[allow(clippy::excessive_precision)]
+    fn test_clenshaw_high_precision_reference() {
+        // Reference accelerations from an independent mpmath (40-digit)
```
duncaneddy, line 3222: Is the mpmath source implementation for these values committed somewhere? Maybe under /scripts? would be good to have as a reference to regenerate if needed.

src/orbit_dynamics/gravity.rs
```diff
@@ -2598,6 +4008,143 @@ mod tests {
...
+        let tables = model.clenshaw.as_ref().unwrap();
+        assert_abs_diff_eq!(
+            tables.c[tables.index(2, 0)],
+            -1.0826e-3 / 5.0_f64.sqrt(),
+            epsilon = 1e-18
+        );
+    }
+
+    #[test]
+    fn test_clenshaw_characterization() {
```
duncaneddy: What does this test get us that the previous mpmath didn't? Can it just be dropped?

src/propagators/dnumerical_orbit_propagator.rs
```diff
@@ -826,6 +811,57 @@ impl DNumericalOrbitPropagator {
...
+            degree,
+            order,
+            parallel,
+        } = &force_config.gravity
+        {
+            let r_i2b = match force_config.frame_transform {
+                FrameTransformationModel::FullEarthRotation => rotation_eci_to_ecef(epoch),
+                FrameTransformationModel::EarthRotationOnly => earth_rotation(epoch),
+            };
+            let r_body = r_i2b * Vector3::new(state_eci[0], state_eci[1], state_eci[2]);
```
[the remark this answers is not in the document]
duncaneddy: Resolved by removal: the trial evaluation itself was removed as overly defensive (see https://github.com/duncaneddy/brahe/pull/378#discussion_r3560487602), so the eager rotation computation no longer exists. Default construction no longer touches the EOP provider.

docs/learn/orbital_dynamics/tides.md
duncaneddy: Converted to tested standalone examples: `examples/numerical_propagation/tides_global_tide_system.py` / `.rs`, included in `tides.md` via `--8<--` and covered by `just test-examples`.

docs/library_api/orbit_dynamics/gravity.md
duncaneddy: Renamed to `GravityModelCoefficients` throughout (Rust enum, Python class, docs section renamed to "Gravity Model Coefficients"), matching the stated preference. `ClenshawTables`/`CunninghamTables` are now `ClenshawCoefficients`/`CunninghamCoefficients`, and the `*_tables` methods (`from_model_type_with_tables`, `precompute_*`, `drop_*`, `has_*`) are now `*_coefficients`.

tests/orbit_dynamics/test_gravity.py
duncaneddy: Added `get_c(n, m)` and `get_s(n, m)` to `GravityModel` (Rust + Python, with parity tests). The tuple-returning `get(n, m)` remains for callers that want both. This test now uses `get_c(2, 0)`.

crates/brahe-py/src/propagators.rs
duncaneddy: Removed entirely — both the Python `UserWarning` helper and the Rust construction-time trial evaluation are gone, along with their tests. The remaining behavior is the descriptive error from the Cunningham kernel itself on overflow (which already lives in the accel method), so there is no extra validation layer anywhere.

src/propagators/dnumerical_orbit_propagator.rs
duncaneddy: Done. `DynamicsWorkspace` is deleted — the dynamics closure now captures `Arc<Mutex<RotationCache>>` directly, and the cache is an `lru::LruCache` (lru crate) with capacity 30 (covers RKN1210's 17 stages plus FSAL headroom), keyed on the stage time's bit pattern (`t.to_bits()`, equivalent to epoch since epoch = epoch_initial + t, and avoids hashing an Epoch). One note: it's captured closure state rather than a named struct field, since the closure is the only consumer — adding a propagator field would just be a never-read duplicate of the same Arc.

src/propagators/dnumerical_orbit_propagator.rs
duncaneddy: Removed. The trial-evaluation block, both Rust tests, and the Python mirror tests are gone. A Cunningham-only configuration that overflows now surfaces the kernel's descriptive error mid-propagation, same as before this PR's guardrails.

src/propagators/dnumerical_orbit_propagator.rs
duncaneddy: Kept the Mutex, for the reason anticipated: `SharedDynamics` is `Send + Sync` and the same closure is shared with the Jacobian and sensitivity providers, so the type system requires `Sync` interior mutability (`RefCell` doesn't compile here). Two changes shrink its footprint though: the lock is now held only for the cache lookup itself (previously it was held across the entire `compute_dynamics` evaluation), and the comment is trimmed to the actual justification. Uncontended acquisition is ~10 ns vs the ~170 µs rotation it saves.

src/propagators/dnumerical_orbit_propagator.rs
duncaneddy: Trimmed to a single line: "Dispatches to the Clenshaw kernel by default."

src/propagators/dnumerical_orbit_propagator.rs
duncaneddy: The gravity-related `use` statements (`GravityModelType`, `ParallelMode`, `set_global_gravity_model`, `set_global_gravity_model_to_tide_system`, `setup_global_test_gravity_model`, `SolidTideConfig`) are hoisted to the top of `mod tests` and all 14 function-local duplicates removed. (The two tests this comment was on were themselves removed with the trial-evaluation guardrail.)

tests/propagators/test_numerical_orbit_propagator.py
duncaneddy: All three tests removed (`warns_on_cunningham_only_high_degree_global`, `construction_fails_cunningham_only_high_degree_leo`, `no_warning_for_default_clenshaw_model`) along with their `_create_geo_state` helper and now-unused imports, matching the removal of the validation they exercised.

src/orbit_dynamics/gravity.rs
duncaneddy: TODO added above `CLENSHAW_PARALLEL_THRESHOLD_NMAX` (also naming `PARALLEL_THRESHOLD_NMAX`): threshold should be derived from thread count/clock speed rather than fixed; current value is a single-machine heuristic.

src/orbit_dynamics/gravity.rs
duncaneddy: Renamed to `GravityModelCoefficients` (picked over `GravityModelConstants` to match the "Gravity Model Coefficients" preference stated on the docs thread — and because the stored data are coefficient representations: normalized packed vs denormalized dense).

src/orbit_dynamics/gravity.rs
duncaneddy: Renamed: `ClenshawCoefficients` / `CunninghamCoefficients`, with prose, error messages ("Clenshaw coefficients not precomputed…"), and test names updated to match.

src/orbit_dynamics/gravity.rs
duncaneddy: Restructured: the test now iterates per-position degree lists (`MID_LAT_LEO_DEGREES`, `EQUATORIAL_LEO_DEGREES`, `ALL_DEGREES`) that simply omit the six invalid points — no exclusion set, no `continue`, and the long rationale comment is reduced to a five-line pointer at `test_clenshaw_high_precision_reference`.

src/orbit_dynamics/gravity.rs
duncaneddy: It wasn't committed — now it is: `scripts/generate_clenshaw_gravity_reference.py` (mpmath at 40 digits, runnable with `uv run`, cases matching the test). Verified it reproduces all six committed reference vectors digit-for-digit.

src/orbit_dynamics/gravity.rs
duncaneddy: Dropped. Its coverage exists elsewhere: `test_gravity_model_compute_spherical_harmonics` pins the main dispatch API with Clenshaw goldens, `test_clenshaw_matches_cunningham` covers the polar-axis geometry against the independent Cunningham kernel, and the mpmath test pins high degrees at 1e-13. The only thing it uniquely held was self-captured goldens at degrees 5/20/80, which the independent-reference tests supersede.

PR #379, opened 2026-07-04
Pre-commit debt: rustfmt, ruff, network-test markers

duncaneddy:
## Description

Clears the pre-commit debt that accumulated because git hooks were never installed in this clone. Five mechanical, tool-separated commits:

1. `style: apply cargo fmt across crate` — `cargo fmt --all` on the 9 drifted files.
2. `style: fix ruff errors and apply ruff format` — fixes all 35 `ruff check` errors (15 auto-fixed unused imports, 20 manual `E702` semicolon-statement splits in `benchmarks/comparative/implementations/gmat/propagation.py`), then `ruff format .` (73 files, pure reflow).
3. `test: mark live-network tests as integration` — adds `@pytest.mark.integration` to every test that reaches a live network call (CelesTrak, space-weather download) in `tests/cli/test_access.py`, `tests/space_weather/test_sw_caching_provider.py`, `tests/space_weather/test_global_sw.py`, and `tests/earth_models/test_nrlmsise00.py`. `tests/celestrak/`, `tests/cli/test_celestrak.py`, and `tests/propagators/test_gp_record_propagators.py` were audited and were already correctly marked or already network-free — no changes needed there.
4. `fix: add safety docs to brahe-py attitude bindings; lint whole workspace in pre-commit` — the pre-commit config's clippy hook ran without `--workspace`, so `crates/brahe-py` was never linted; adding `--workspace` surfaced 10 pre-existing `brahe-py` errors. Fixed all of them: `# Safety` doc sections for 5 `pub unsafe fn` numpy-conversion methods in `crates/brahe-py/src/attitude.rs` (verified none actually perform an unsafe operation — `into_pyarray`/`reshape` are safe in the `numpy` crate version this repo pins, and the `unsafe` marker predates the file's earliest commit, so it's documented rather than removed to keep this change signature-preserving), plus `Default` impls and `is_empty()` methods for 3 `PyEOPProvider` structs in `crates/brahe-py/src/eop.rs` (clippy's own suggested idiomatic fix for `new_without_default`/`len_without_is_empty`).
5. `test: make access CLI validation tests hermetic against EOP cache age` — a reviewer proved empirically (aging the real EOP cache file to 9+ days and timing a "validation-only" test) that all 26 tests in `tests/cli/test_access.py`, not just the 17 marked in commit 3, reach live network: `compute()` calls `set_cli_eop()` → `brahe.initialize_eop()` **before any argument validation**, and `CachingEOPProvider::new` unconditionally re-downloads the cached EOP file whenever it's stale, independent of which code path runs afterward. Rather than marking all 26 integration (losing fast local coverage of the CLI's validation logic), added an autouse `tmp_path`-backed `BRAHE_CACHE` fixture so every test in the file cold-seeds from the bundled EOP data (always fresh) instead of reading the real on-disk cache. Verified by replicating the reviewer's exact experiment against the real cache file: with it artificially aged to 9+ days, the fixture keeps the 9 unmarked tests fast (0.02s each, same as before) and leaves the real cache file's mtime and content completely untouched.

None of these commits change behavior: `ruff format` and `cargo fmt` are pure formatters, the manual ruff fixes are dead-import removals plus semicolon-to-newline statement splits, the `integration` marker only gates test *selection* (`addopts = "-m 'not integration'"` already existed and already gated 83 tests before this PR), and the safety-doc/lint fixes in commit 4 add no new unsafe behavior (`is_empty()` is a new but trivial Python-facing method delegating to an existing trait default; `Default` is Rust-side only and doesn't change the PyO3 `__new__` constructor).

After merge, hooks should be enabled with `pre-commit install` so this doesn't drift again.

**Rebased** onto the final `de/clenshaw-gravity` tip (`c7fffa75`) once that branch was done. Commit 1 (`cargo fmt`) was regenerated fresh against the new base rather than rebased (formatter commits don't rebase cleanly); commits 2-5 cherry-picked cleanly with no conflicts. All gates below were re-run against the merged state.

### Fixed
- Repo-wide `cargo fmt` and `ruff` lint/format debt that had accumulated since pre-commit hooks were never installed locally.
- Live-network tests (CelesTrak CLI/API calls, space-weather cache downloads) that previously hung or failed in sandboxed/CI environments are now marked `@pytest.mark.integration` and deselected by default.
- `crates/brahe-py` clippy errors (missing `# Safety` docs, missing `Default`/`is_empty`) that were silently exempted from the pre-commit gate because the clippy hook lacked `--workspace`; the hook now covers the whole workspace.
- `tests/cli/test_access.py`'s 9 non-`integration`-marked tests could still silently hit the network (and rewrite the real on-disk EOP cache) via `set_cli_eop()`'s unconditional cache-age check; they now run against an isolated, always-fresh temp cache.

### Added
- `is_empty()` method on `StaticEOPProvider`, `FileEOPProvider`, and `CachingEOPProvider` (Python bindings), matching the existing `Trajectory.is_empty()` convention.

## Related Issue

N/A

## Note to Reviewers

The diff is large but entirely mechanical — each commit is scoped to one tool/fix so they can be reviewed independently. Commit 4 was added after an independent reviewer found that `cargo clippy --all-targets --all-features` (without `--workspace`) never lints `crates/brahe-py`, which let 10 clippy errors accumulate there undetected by this repo's own pre-commit hook. Commit 5 was added after a second reviewer found that commit 3's network-test audit had a gap: it only traced the CelesTrak fetch inside `compute()` and missed that `set_cli_eop()` reaches a network-capable code path (EOP cache refresh) earlier, before any validation runs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QJNzZnqrG2yp2RzGfNpimG


duncaneddy:
Superseded by #380 — same five commits regenerated directly against main (d9b23538) so the cleanup can land independently of #378, avoiding a force-push of this branch. All gates re-verified on the main-based stack.

PR #386, opened 2026-07-11
Fix estimation module defects and rework estimator construction API

duncaneddy:
# Pull Request

## Description

Fixes the defects found in an in-depth review of the `estimation` submodule and reworks estimator construction for composability with the propagators. The headline items: consider-parameter batch least squares panicked on first use, the `StackedObservationMatrix` solver silently mis-weighted correlated measurement noise, a failed observation could permanently desynchronize a filter's state and covariance, the UKF paid hidden STM-integration cost on every sigma-point propagation, and filters accumulated unbounded trajectory data. Estimator `new()` constructors keep the full orbit-propagator pass-through (force model parameters, additional dynamics, control inputs); `from_propagator(propagator, models, config)` accepts any pre-built propagator and takes its initial covariance from the propagator's `initial_covariance`, with covariance managed by the estimator internally.

Verified with 4,445 Rust lib tests, 3,020 Python tests, and all estimation examples in both languages (estimation suite grew from 102 to 116 Rust tests plus new covariance-validation tests and error-path coverage: terminal-event guards, transactional rollback, model output shape validation, and sigma-parameter validation in both languages).

### Added

- `disable_stm_propagation()` on `DNumericalOrbitPropagator` and `DNumericalPropagator` (Rust and Python): turns off variational-equation integration after construction (covariance auto-enables it), clearing the STM and covariance which can no longer be propagated. The UKF uses this internally so `from_propagator` with a covariance-carrying propagator does not pay STM-integration cost during sigma-point propagation

- The propagator's consider / force-model parameter vector is now passed through to measurement models: `DNumericalOrbitPropagator` and `DNumericalPropagator` expose a `params()` getter, and EKF/UKF/BLS forward it to every `MeasurementModel::predict`/`jacobian` call, so consider values (e.g., measurement biases) can affect predicted measurements. The Python `MeasurementModel` interface gains the same optional third argument: `predict(self, epoch, state, params)`

- `ExtendedKalmanFilter`, `UnscentedKalmanFilter`, and `BatchLeastSquares` expose `dynamics()` and `into_dynamics()` accessors for the underlying propagator
- `DynamicsSource` implements `From<DNumericalOrbitPropagator>` and `From<DNumericalPropagator>`; `from_propagator` accepts `impl Into<DynamicsSource>` so propagators can be passed directly
- End-to-end consider-parameter coverage for `BatchLeastSquares` (both solver methods, Rust and Python) and a correlated-noise solver-equivalence regression test
- Python `BatchLeastSquares.total_covariance()` returning the formal-plus-consider covariance
- Estimators validate measurement-model output shapes (`predict`, `jacobian`, `noise_covariance`) and return structured errors instead of dimension panics, including for Python-defined models
- Estimators error when propagation stops short of an observation epoch (e.g., a terminal event fired) instead of silently processing a stale state

### Changed

- **Breaking**: `from_propagator(propagator, measurement_models, config)` accepts propagators directly via `impl Into<DynamicsSource>` and takes the filter's initial covariance from the propagator's `initial_covariance`, erroring if the propagator was not initialized with one; `new()` signatures are unchanged and keep the full propagator pass-through (`params`, `additional_dynamics`, `control_input`)
- **Breaking**: covariance is managed by the estimator internally — the UKF no longer integrates variational equations during sigma-point propagation (6 ODE dimensions instead of 42 for a 6D state), through both `new()` and `from_propagator` construction paths
- **Breaking**: Python `BatchLeastSquares.current_covariance()` now returns the formal covariance (mirroring Rust); use `total_covariance()` for the consider-inclusive total. Python `ExtendedKalmanFilter.current_covariance()` no longer returns `Optional`
- `current_covariance()` returns `&DMatrix<f64>` uniformly across all three estimators
- Estimators disable trajectory recording on their propagator, eliminating unbounded trajectory growth during filtering (UKF previously interleaved 2n+1 sigma-point passes per interval; BLS re-recorded the arc every iteration)
- `DynamicsSource` is `#[non_exhaustive]`
- `validate_covariance` additionally rejects non-finite entries and non-positive-definite matrices; `covariance_from_upper_triangular` validates its result
- `BLSIterationRecord.cost` and `final_cost()` are evaluated at the post-correction state so the recorded cost describes the recorded state
- All numerical differentiation shares one engine: a new `numerical_jacobian()` in `math::jacobian` (rectangular, fallible-function finite differences) now backs `DNumericalJacobian`, `DNumericalSensitivity`, and `measurement_jacobian_numerical`, and the default `MeasurementModel::jacobian` delegates to it
- Estimator rustdoc examples are compiled code instead of commented-out fragments; solver-method and process-noise documentation corrected to match the implementations

### Removed

- `UKFConfig.state_dim` — the state dimension is derived from the propagator at filter construction

### Fixed

- Consider-parameter `BatchLeastSquares::solve()` panicked with a dimension mismatch on first use: the n_solve-sized formal covariance was fed to the full-state propagator's covariance propagation
- `StackedObservationMatrix` silently dropped off-diagonal measurement-noise correlations during whitening, producing mis-weighted estimates for correlated R; it also formed the normal equations despite documenting a QR solve — it now whitens with the Cholesky factor of R and solves via QR
- A failed observation (erroring measurement model, non-positive-definite innovation covariance) left EKF/UKF filters with state and covariance from different epochs — the UKF was left holding a propagated sigma point as its state; `process_observation` now rolls the propagator back on every error path
- UKF sigma-point parameter validation accepted NaN `alpha`/`beta`/`kappa` and underflow-sized `alpha`, silently producing NaN estimates instead of a construction error
- Covariance updates are symmetrized and Kalman gains computed via Cholesky solve rather than explicit inversion, preventing roundoff-asymmetry-induced Cholesky failures in long filter runs

## Related Issue

N/A — follows from a code review of the estimation submodule.


crates/brahe-py/src/estimation.rs
```diff
@@ -1413,23 +1411,21 @@ impl MeasurementModel for MeasurementModelHolder {
         &self,
         epoch: &brahe::time::Epoch,
         state: &DVector<f64>,
-        params: Option<&DVector<f64>>,
     ) -> Result<DVector<f64>, brahe::utils::errors::BraheError> {
         match self {
-            MeasurementModelHolder::RustNative(m) => m.predict(epoch, state, params),
```
duncaneddy, line 1419: It seems to me that dropping the additional parameters from the calls that should pass through to an underlying DNumericalOrbitPropagator or NumericalPropagator is a large mistake. We should be able to leverage the full range of propagator initializations and configurations with estimations (e.g. control inputs, additional parameters, additional dynamics etc. This seems like a breaking change for that.

brahe/estimation.py
```diff
@@ -45,9 +45,9 @@
     InertialPositionMeasurementModel,
     InertialVelocityMeasurementModel,
     InertialStateMeasurementModel,
-    ECEFPositionMeasurementModel,
-    ECEFVelocityMeasurementModel,
-    ECEFStateMeasurementModel,
+    EcefPositionMeasurementModel,
```
duncaneddy: These should remain capitalized ECEF* since that is a acronym for a standard frame and consistent with capitalization of other frames throughout the module.

src/estimation/ekf.rs
duncaneddy, line 138: This seems like we're improperly dropping the ability to define arbitrary numerical orbit propagators.

src/estimation/ekf.rs
```diff
@@ -951,23 +1198,11 @@ mod tests {
...
-            true_state,
-            cfg,
-            ForceModelConfig::two_body_gravity(),
-            None,
-            None,
-            None,
-            Some(p0),
-        )
-        .unwrap();
+        let prop = create_stm_propagator(epoch, true_state);
```
duncaneddy: Also dropping needed arguments here.

src/estimation/ukf.rs
```diff
@@ -220,17 +264,18 @@ impl UnscentedKalmanFilter {
             weights_cov[i] = w_i;
         }
 
-        let covariance = initial_cov.clone();
+        dynamics.set_trajectory_mode(TrajectoryMode::Disabled);
```
duncaneddy, line 305: Should this be mirrored in the EKF implementation?

src/estimation/ekf.rs
duncaneddy: Restored in 3fa58e4b: `new()` on all three estimators takes `params`, `additional_dynamics`, and `control_input` again and passes them through to the internal orbit propagator, so the full range of propagator configuration is available directly. `from_propagator` remains the path for generic propagators (`DNumericalPropagator`) and any pre-built propagator — nothing is lost there since the propagator carries its own dynamics/params/control configuration.

crates/brahe-py/src/estimation.rs
duncaneddy: Restored in 3fa58e4b — the Rust constructors take the full pass-through again (the Python constructors had kept these kwargs throughout). One related note: the `params` argument that was removed from `MeasurementModel::predict`/`jacobian` is separate from propagator configuration — no estimator ever passed it to a measurement model and the Python `MeasurementModel` interface never exposed it, so it was unreachable dead surface. Happy to rewire it end-to-end instead if you want measurement models to receive the force-model parameter vector.

brahe/estimation.py
duncaneddy: Reverted in 3fa58e4b — Python keeps the ECEF* capitalization (class names, tests, examples, docs, plots, and stubs all restored). Rust keeps `Ecef*` per Rust type-naming convention, matching how the original code had it.

src/estimation/ekf.rs
duncaneddy: Restored in 3fa58e4b along with the other constructor call sites.

src/estimation/ukf.rs
duncaneddy: Yes — it's already mirrored: `ExtendedKalmanFilter::from_propagator` and `BatchLeastSquares::from_propagator` both call `set_trajectory_mode(TrajectoryMode::Disabled)` on their dynamics at construction, for the same reason (filters re-propagate overlapping spans, which would otherwise accumulate unbounded interleaved trajectory data — worst in the UKF where every interval is propagated 2n+1 times). `test_ekf_trajectory_recording_disabled` covers the EKF side.

duncaneddy:
**params restoration (74c6654b):** the `params` argument is restored on `MeasurementModel::predict`/`jacobian` and `measurement_jacobian_numerical`, and is now wired end-to-end rather than being dead surface — both propagators expose a `params()` getter, and EKF/UKF/BLS forward the propagator's consider / force-model parameter vector to every measurement model call, so consider values (e.g., measurement biases) can affect predicted measurements. The Python `MeasurementModel` interface gains the same optional third argument (`predict(self, epoch, state, params)`, receiving `None` when the propagator has no parameters). Threading is asserted by tests in all three estimators and through the Python wrapper; docs and stubs updated.

PR #394, opened 2026-07-15
Add per-force-model body attribution

### LINKED ISSUE #387: Per-force-model body attribution: non-central-body gravity fields and drag (EMB/cislunar configs)
Follow-up from PR #377 review discussion (comments on dnumerical_orbit_propagator.rs).

## Problem

Force models are currently evaluated about the propagation's central body. Two configurations this cannot express:

1. **Zonal/spherical-harmonic gravity attached to a non-central body.** For an EMB- or Moon-centered propagation, one may want Earth J2-J6 (or an Earth spherical-harmonic field) applied as a perturbation even though Earth is not the central body. Today `GravityConfiguration::EarthZonal` is validated Earth-central-only, and third-body perturbers are point masses.

2. **Earth-centered drag under an EMB central body.** A cislunar trajectory passing through LEO altitudes would need an atmospheric drag model evaluated at the object's position relative to Earth, while the integration state is EMB-centered. Today drag uses the central body's radius/spin, and `validate()` rejects drag for barycenters.

## Sketch

Attach an optional `body: CentralBody` (or NAIF ID) to individual force-model configurations (gravity beyond the central field, drag), evaluated at the object's position relative to that body (via the ephemeris offset), with the resulting acceleration expressed in the propagation frame. Validation rules move from "central body must be Earth" to "attributed body must be Earth" for Earth-specific models (NRLMSISE-00, Harris-Priester, EarthZonal).

This generalizes the third-body machinery (which already resolves per-body ephemeris offsets kernel-scoped) from point masses to extended force models.

duncaneddy:
# Pull Request

## Description

Adds per-force-model body attribution so gravity fields and atmospheric drag can be attributed to a body other than the propagation's central body — e.g. an EMB-centered cislunar trajectory with an Earth spherical-harmonic field and Earth-attributed NRLMSISE-00 drag. Third-body configuration is flattened to one entry per body (each with its own ephemeris source and gravity model), `ThirdBody` distinguishes planet centers from system barycenters, and the planetary GM constants are split into planet-only and system values sourced from NAIF `gm_de440.tpc`.

Correctness gates: an EMB-centered propagation with an Earth spherical-harmonic (or J2 zonal) third body matches the Earth-centered formulation to 1 m / 1 mm/s over one day, and the drag-induced trajectory delta under an EMB center matches the Earth-centered drag path to 0.1 m per orbit (Rust + Python integration tests).

### Added
- Per-body third-body configuration: `ThirdBodyConfiguration` pairs each perturber with its own `ephemeris_source` and `gravity` model; `SphericalHarmonic` and `EarthZonal` fields are evaluated at the object's position relative to the perturbing body in that body's fixed frame, with a point-mass indirect term.
- Public `accel_third_body_field_for_body` (Rust + Python) evaluating a configured gravity model (point-mass, spherical-harmonic, or Earth-zonal) for a perturbing body about any central body.
- `ThirdBody` planetary-system barycenter variants `MarsBarycenter`–`NeptuneBarycenter` (NAIF 4–8, system GMs), used by the default Earth force models; `accel_third_body` accepts both barycenter and planet-center variants (the latter auto-load their satellite-system kernels).
- `DragConfiguration.body` to attribute drag to a non-central body: density and relative wind are evaluated at the object's state relative to that body (enables Earth drag on EMB-centered propagations).
- `GravityConfiguration::Zero` — explicit no-central-gravity term for barycentric propagation centers (used by `cislunar_default`).
- ECI↔EMBI frame translation helpers `position_eci_to_emb`/`position_emb_to_eci` and `state_eci_to_emb`/`state_emb_to_eci` (Rust + Python).
- Constants `GM_MARS_SYSTEM`–`GM_NEPTUNE_SYSTEM` and `GM_PLUTO_SYSTEM` (planetary-system barycentric GMs, NAIF `gm_de440.tpc`).
- `ThirdBody::as_central_body()` and `ThirdBody::body_fixed_frame()` helpers (Rust + Python).
- Standalone example `numerical_propagation/cislunar_earth_forces` (EMB-centered propagation with Earth-attributed forces).

### Changed
- **Breaking:** `ForceModelConfig.third_body` becomes `Option<Vec<ThirdBodyConfiguration>>` with per-body entries. The field deserializes from bare bodies, a single entry, a list, or the legacy pre-flattening object shape, which migrates on load with its original semantics (pre-split planet names map to the `*Barycenter` variants they then denoted).
- **Breaking:** `ThirdBody::Mars`–`Neptune` now denote planet centers (NAIF 499–899, planet-only GMs, resolved through satellite-system kernels) instead of system barycenters; use the new `*Barycenter` variants for the previous behavior.
- **Breaking:** `GM_MARS`–`GM_NEPTUNE` and `GM_PLUTO` now hold planet-only GMs from NAIF `gm_de440.tpc`; the previous DE430-era system values moved to `GM_*_SYSTEM` and were updated to DE440 (full-precision tpc digits).
- Force-model validation keys the Earth-atmosphere-model and radius/spin drag requirements on the *attributed* body rather than the central body, and validates per-third-body gravity models (`EarthZonal` requires `ThirdBody::Earth`; `SphericalHarmonic` requires a body with a known body-fixed frame). An attributed drag body's ephemeris is verified at propagator construction.

## Related Issue

Closes #387

## Note to Reviewers

- The physics equivalence tests (`test_emb_earth_spherical_harmonic_matches_earth_centered`, `test_emb_earth_zonal_matches_earth_centered`, `test_emb_attributed_earth_drag_matches_earth_centered`, `test_emb_combined_earth_field_and_drag_matches_earth_centered`) are the main correctness gates and run in the default suite (all required kernels/models are in the CI data-cache manifest or packaged). The drag test compares drag-on/off deltas because the two gravitational formulations carry a ~5 m/orbit baseline difference that would otherwise mask the drag path.
- The legacy `third_body` shape migration (the `deserialize_third_bodies` field deserializer) is worth a close look — it exists so pre-branch serialized configs keep their physics instead of silently losing all third-body forces.
- Earth-centered configurations are bit-identical to `main` (asserted by existing tests at 1e-15–1e-18 tolerances).


docs/learn/orbit_propagation/numerical_propagation/cislunar_lunar_propagation.md
```diff
@@ -18,14 +18,18 @@ The propagator integrates in the central body's inertial frame (`LCI` for `Moon`
...
 
 This handling is automatic: selecting `CentralBody::EMB` (via `cislunar_default()` or directly) routes Earth and Moon perturbations through the direct form and all others through the differential form.
 
+## Earth-Attributed Force Models
+
+An EMB-centered trajectory that passes through low Earth altitudes can carry Earth-fidelity forces without switching central bodies. A third-body entry's `gravity` accepts a `SphericalHarmonic` or `EarthZonal` field evaluated at the object's position relative to that body, and `DragConfiguration.body` attributes atmospheric drag to a body other than the central body (density and relative wind are evaluated at the object's Earth-relative state). See [Force Models: Body Attribution](force_models.md#body-attribution) for the configuration details and a complete EMB-centered example with an 8x8 Earth field and Earth-attributed NRLMSISE-00 drag.
```
duncaneddy: "(density and relative wind are evaluated at the object's Earth-relative state)" this seems incorrect since the evaluation method should work for other bodies (like mars). Specifically if you're propagating around the Mars-barycenter but define the drag relative to Mars".

This should be "density and relative wind are evaluated at the object's body-relative state"

docs/learn/constants.md
```diff
@@ -53,15 +53,21 @@ be well known and do not change frequently.
...
+| `GM_URANUS`   | Gravitational constant of Uranus (planet only, without its satellites).                                       | $5.793951256527211 \times 10^{15}$   | $\frac{m^3}{s^2}$ | NAIF `gm_de440.tpc` [^6]   |
+| `GM_URANUS_SYSTEM` | Gravitational constant of the Uranus system barycenter (NAIF ID 7).                                      | $5.7945564 \times 10^{15}$           | $\frac{m^3}{s^2}$ | NAIF `gm_de440.tpc` [^6]   |
+| `GM_NEPTUNE`  | Gravitational constant of Neptune (planet only, without its satellites).                                      | $6.835103145462294 \times 10^{15}$   | $\frac{m^3}{s^2}$ | NAIF `gm_de440.tpc` [^6]   |
+| `GM_NEPTUNE_SYSTEM` | Gravitational constant of the Neptune system barycenter (NAIF ID 8).                                    | $6.836527100580399 \times 10^{15}$   | $\frac{m^3}{s^2}$ | NAIF `gm_de440.tpc` [^6]   |
 | `GM_PLUTO`    | Gravitational constant of the Pluto.                                                                          | $9.770 \times 10^{11}$               | $\frac{m^3}{s^2}$ | Montenbruck and Gill [^1]  |
```
duncaneddy: GM_PLUTO seems it could be updated from gm_d3440.tpc to match `8.696138177608748` which is the value from de440. Simiarly could separate out the planet and the barycenter.

docs/library_api/constants/physical.md
```diff
@@ -244,7 +244,19 @@ Lunar radius.
...
+
+#### GM_MARS_SYSTEM
+
+::: brahe.GM_MARS_SYSTEM
+    options:
+      show_root_heading: true
+      show_root_full_path: false
+      heading_level: 5
+
+**Value**: `4.283e13` m³/s²
```
duncaneddy: Drop the "Value" from the docs here, unneeded, covered by `learn` docs.

examples/numerical_propagation/cislunar_earth_forces.py
```diff
@@ -0,0 +1,73 @@
...
+bh.initialize_sw()
+
+epoch = bh.Epoch.from_datetime(2024, 3, 1, 0, 0, 0.0, 0.0, bh.TimeSystem.UTC)
+
+# EMB-centered force model: point-mass "central" gravity (the barycenter has
+# no mass of its own); Earth carries a spherical-harmonic field and the
+# atmosphere; the Moon and Sun are point-mass perturbers.
+force_config = bh.ForceModelConfig.for_body(
+    bh.CentralBody.EMB,
+    bh.GravityConfiguration.point_mass(),
```
duncaneddy: Instead of having `GravityConfiguration.point_mass()` represent barycenter propagation (without a direct attracting body). I think `GravitationalConfiguration.barycenter()` or `GravityConfiguration.zero()` would be more informative and useful an approach. This was it is explicit that there is no gravity applied by the barycenter.

examples/numerical_propagation/cislunar_earth_forces.py
```diff
@@ -0,0 +1,73 @@
...
+    ],
+    mass=bh.ParameterSource.value(1000.0),
+)
+force_config.validate()
+
+# Start from a 500 km Earth orbit, re-expressed about the EMB by adding
+# Earth's barycentric state from the DE ephemeris.
+oe = np.array([bh.R_EARTH + 500e3, 0.001, 51.6, 15.0, 30.0, 45.0])
+x_earth = bh.state_koe_to_eci(oe, bh.AngleFormat.DEGREES)
+x_emb = x_earth + bh.spk_state(399, 3, epoch)
```
duncaneddy: It looks like we're missing a `state_eci_to_emb` transformation that would make this more intuitive and direct. This should be added to the frames module. (could also be named `state_gcrf_to_emb`. (Also need similar `position_*` method.) Rotation is unity so can be omitted.

examples/numerical_propagation/cislunar_earth_forces.py
```diff
@@ -0,0 +1,73 @@
...
+    epoch,
+    x_emb,
+    bh.NumericalPropagationConfig.default(),
+    force_config,
+    None,
+)
+
+# Propagate for one day
+epoch_end = epoch + 86400.0
+prop.propagate_to(epoch_end)
```
duncaneddy, line 63: This example could be a useful consistency test (in python and rust) to setup the same force model in the EMB center and one with the default ECI center and compare the propagation outputs (which should be identical).

src/constants/physical.rs
```diff
@@ -169,45 +169,185 @@ pub const GM_MERCURY: f64 = 22031.780000 * 1e9;
...
+///  1. NAIF `gm_de440.tpc` (`BODY4_GM`), JPL DE440 "ASTRO-VALUES";
+///     R.S. Park, W.M. Folkner, J.G. Williams, and D.H. Boggs, *The JPL
+///     Planetary and Lunar Ephemerides DE440 and DE441*, The Astronomical
+///     Journal, 161:105, 2021. doi:10.3847/1538-3881/abd414
+///
+/// # Examples
+/// ```
+/// assert_eq!(brahe::constants::GM_MARS_SYSTEM, 42828.375815756102 * 1e9);
+/// ```
+pub const GM_MARS_SYSTEM: f64 = 42_828.375_815_756_1 * 1e9;
```
duncaneddy: Drops the 02 at the end.

Should be 4.2828375815756102

src/constants/physical.rs
```diff
@@ -169,45 +169,185 @@ pub const GM_MERCURY: f64 = 22031.780000 * 1e9;
...
+/// # References:
+///
+///  1. NAIF `gm_de440.tpc` (`BODY7_GM`), JPL DE440 "ASTRO-VALUES";
+///     R.S. Park, W.M. Folkner, J.G. Williams, and D.H. Boggs, *The JPL
+///     Planetary and Lunar Ephemerides DE440 and DE441*, The Astronomical
+///     Journal, 161:105, 2021. doi:10.3847/1538-3881/abd414
+///
+/// # Examples
+/// ```
+/// assert_eq!(brahe::constants::GM_URANUS_SYSTEM, 5794556.4 * 1e9);
```
duncaneddy: Should be 5.7945563999999985

src/constants/physical.rs
```diff
@@ -169,45 +169,185 @@ pub const GM_MERCURY: f64 = 22031.780000 * 1e9;
...
+/// # References:
+///
+///  1. NAIF `gm_de440.tpc` (`BODY6_GM`), JPL DE440 "ASTRO-VALUES";
+///     R.S. Park, W.M. Folkner, J.G. Williams, and D.H. Boggs, *The JPL
+///     Planetary and Lunar Ephemerides DE440 and DE441*, The Astronomical
+///     Journal, 161:105, 2021. doi:10.3847/1538-3881/abd414
+///
+/// # Examples
+/// ```
+/// assert_eq!(brahe::constants::GM_SATURN_SYSTEM, 37940584.8418 * 1e9);
```
duncaneddy: Should be 3.7940584841799997

src/constants/physical.rs
```diff
@@ -169,45 +169,185 @@ pub const GM_MERCURY: f64 = 22031.780000 * 1e9;
...
+///  1. NAIF `gm_de440.tpc` (`BODY5_GM`), JPL DE440 "ASTRO-VALUES";
+///     R.S. Park, W.M. Folkner, J.G. Williams, and D.H. Boggs, *The JPL
+///     Planetary and Lunar Ephemerides DE440 and DE441*, The Astronomical
+///     Journal, 161:105, 2021. doi:10.3847/1538-3881/abd414
+///
+/// # Examples
+/// ```
+/// assert_eq!(brahe::constants::GM_JUPITER_SYSTEM, 126712764.1 * 1e9);
+/// ```
+pub const GM_JUPITER_SYSTEM: f64 = 126712764.1 * 1e9;
```
duncaneddy: Should be 1.2671276409999998

tests/propagators/test_numerical_orbit_propagator.py
```diff
@@ -5960,3 +5960,157 @@ def test_state_in_frame_lci_equals_central_inertial_for_lunar_propagator(
...
+        CentralBody.EMB,
+        GravityConfiguration.point_mass(),
+        third_bodies=[
+            bh.ThirdBodyConfiguration(bh.ThirdBody.EARTH, gravity=sh),
+            bh.ThirdBody.MOON,
+            bh.ThirdBody.SUN,
+        ],
+    )
+
+    earth_wrt_emb = spk_state(399, 3, epoch)
```
duncaneddy: Should use EMB transform function

tests/propagators/test_numerical_orbit_propagator.py
```diff
@@ -5960,3 +5960,157 @@ def test_state_in_frame_lci_equals_central_inertial_for_lunar_propagator(
         x_lci = prop.state_in_frame(ReferenceFrame.LCI, epoch)
         x_central = prop.state_bci(epoch)
         assert np.array_equal(x_lci, x_central)
+
+
+@pytest.mark.integration
```
duncaneddy: Why does this need to be integration? I believe we assume (though `just setup`) the the requisite caches and downloaded for unit testing

tests/propagators/test_numerical_orbit_propagator.py
```diff
@@ -5960,3 +5960,157 @@ def test_state_in_frame_lci_equals_central_inertial_for_lunar_propagator(
...
+    prop_emb.propagate_to(epoch_end)
+
+    xf_earth = prop_earth.current_state()
+    xf_emb_in_earth = prop_emb.current_state() - spk_state(399, 3, epoch_end)
+
+    assert np.allclose(xf_emb_in_earth[:3], xf_earth[:3], atol=1.0)
+    assert np.allclose(xf_emb_in_earth[3:], xf_earth[3:], atol=1e-3)
+
+
+@pytest.mark.integration
```
duncaneddy: Same. What is the need to integration.

tests/propagators/test_numerical_orbit_propagator.py
```diff
@@ -5960,3 +5960,157 @@ def test_state_in_frame_lci_equals_central_inertial_for_lunar_propagator(
...
+
+    delta_earth = run(CentralBody.Earth, drag_earth) - run(CentralBody.Earth, None)
+    delta_emb = run(CentralBody.EMB, drag_attributed) - run(CentralBody.EMB, None)
+
+    assert np.linalg.norm(delta_earth[:3]) > 1.0
+    assert np.allclose(delta_emb[:3], delta_earth[:3], atol=0.1)
+    assert np.allclose(delta_emb[3:], delta_earth[3:], atol=1e-4)
+
+
+@pytest.mark.integration
```
duncaneddy: Also what is the need for integration testing?

src/orbit_dynamics/third_body.rs
```diff
@@ -120,60 +133,68 @@ pub fn accel_third_body<P: IntoPosition>(
...
 
-        // Invalid: bodies only supported through `accel_third_body_for_body`
-        // (Earth, Phobos, Deimos, and Custom only make sense relative to a
+        // Invalid: bodies only supported through `accel_third_body_for_body`.
+        // Earth, Phobos, Deimos, and Custom only make sense relative to a
         // non-Earth central body, which this Earth-centered function does
-        // not model).
+        // not model; the planet-center variants (Mars..Neptune) additionally
+        // require their satellite-system ephemeris kernels, which only the
+        // kernel-scoped resolution in `accel_third_body_for_body` handles.
```
duncaneddy: Isn't by using the *_spice calls this ultimately doing the same kernel resolution so the planetary variants could be directly accepted here?

src/orbit_dynamics/third_body.rs
```diff
@@ -308,65 +329,191 @@ pub fn accel_third_body_for_body<P: IntoPosition>(
...
+///
+/// # Returns
+///
+/// * `Ok(Vector3<f64>)` - Acceleration in the inertial frame centered on
+///   `central_body`. Units: [m/s²]
+/// * `Err(BraheError)` - If the body coincides with the central body, the
+///   ephemeris cannot be queried, or a `ModelType` spherical-harmonic entry
+///   is missing its resolved model
+#[allow(clippy::too_many_arguments)]
+pub(crate) fn accel_third_body_extended(
```
duncaneddy: I actually disagree this is a well structured function and could be made public, but I dislike the name of `*_extended`, _for_body is good ending, just not sure I love *_extended for this.

src/propagators/force_model_config.rs
```diff
@@ -172,17 +180,15 @@ impl Default for ForceModelConfig {
...
                 area: ParameterSource::ParameterIndex(3),
                 cr: ParameterSource::ParameterIndex(4),
                 eclipse_model: EclipseModel::Conical,
                 occulting_bodies: vec![OccultingBody::Earth],
             }),
-            third_body: Some(ThirdBodyConfiguration {
-                ephemeris_source: EphemerisSource::DE440s,
-                bodies: vec![ThirdBody::Sun, ThirdBody::Moon],
-            }),
+            third_bodies: Some(vec![ThirdBody::Sun.into(), ThirdBody::Moon.into()]),
```
duncaneddy: Nit: the name of the field should still be `third_body` since that is generally name of the force, it semantically follows that each field is the force configuration for that type of force.

duncaneddy:
Review feedback addressed in d1762ce6 — resolution map:

| Comment | Resolution |
|---|---|
| `third_bodies` → `third_body` field name | Renamed everywhere (Rust, Python kwarg/getter/setter, tests, examples, docs, benchmarks). The legacy pre-flattening object shape now migrates through the same key's deserializer, so the both-fields ambiguity case disappears. |
| `GravityConfiguration.barycenter()`/`zero()` | Added `GravityConfiguration::Zero` (`GravityConfiguration.zero()` in Python) — an explicit no-central-gravity variant. `cislunar_default()` and the cislunar example now use it. A `Zero` gravity model on a *third-body entry* is rejected by validation (a disabled entry is a config mistake). |
| Missing `state_eci_to_emb` transform | Added `position/state_eci_to_emb` and `position/state_emb_to_eci` to the frames module (pure translation via the DE Earth↔EMB state; rotation is identity as noted), with Python bindings and tests. The example and equivalence tests now use them. |
| Example as consistency test | Added `test_emb_combined_earth_field_and_drag_matches_earth_centered` (Rust + Python) — the example's exact configuration (Earth SH 8×8 third body + Earth-attributed NRLMSISE-00 drag about EMB) vs the Earth-centered equivalent, agreeing within the gravitational formulation baseline. The SH-only, zonal-only, and drag-delta tests remain as sharper individual gates. |
| System GM precision (206/242/276/312) | Restored full `gm_de440.tpc` digits with `#[allow(clippy::excessive_precision)]`. Note Jupiter/Uranus shorthand values were 1 ulp off the tpc text — doc examples and Python tests updated to the faithful values. |
| GM_PLUTO update | Split like the others: `GM_PLUTO` = 869.6138177608748 km³/s² (`BODY999_GM`, planet-only) and new `GM_PLUTO_SYSTEM` = 975.5 km³/s² (`BODY9_GM`). |
| Drop **Value** lines in API docs | Removed for the planetary GM block (they were also stale). |
| "Earth-relative state" wording | Corrected to "body-relative state" in the cislunar doc. |
| Planet variants in `accel_third_body` | Accepted directly: planet-center variants route through the body-center `*_position_spice` functions (auto-loading their satellite-system kernels) with planet-only GMs; verified to match `accel_third_body_for_body` exactly. |
| `accel_third_body_extended` naming/visibility | Now public as `accel_third_body_field_for_body` (Rust + Python), resolving the body-fixed rotation and gravity model internally; the propagator keeps a crate-internal `_with_model` variant for its precomputed hot path. |
| Why `@pytest.mark.integration`? | Removed — the EMB equivalence tests, field-acceleration tests, and planet-center tests now run in the default suites in both languages (everything they need is in the CI data-cache manifest or packaged, and space weather uses the local test asset). |

Also fixed the four failing `just test-examples` targets (missing `DragConfiguration.body` field in struct literals), migrated the two Python profiling scripts flagged by Codex, and addressed the codecov gaps: the formerly integration-gated tests now provide default-suite coverage of the attributed-body paths, new unit tests cover the point-mass/zero/global/missing-model branches of the field acceleration and the Earth-central satellite-body routing arm, and the propagator test module gained the missing `#[cfg_attr(coverage_nightly, coverage(off))]`.

One structural note on the "cfg(test) sections reported as missed" confusion: the `coverage(off)` attribute pattern used across the repo only takes effect when coverage runs on a nightly toolchain (`cargo-llvm-cov` sets `--cfg coverage_nightly` only there), but all coverage workflows run stable — so ignored (feature-gated) test bodies count as uncovered source. Un-gating the tests fixes both the branch coverage and the phantom-missed test code for this PR; if you want test modules excluded from coverage universally, the coverage jobs would need to run on nightly. Happy to make that workflow change in a follow-up if desired.

duncaneddy:
Follow-up on the `GM_MARS`/system-value question (investigated in `d254de8e`): the mismatch is a provenance split inside `gm_de440.tpc` itself. The file's header sources the barycenter GMs (`BODY1`–`BODY9`) from the DE-440 NAVIO "ASTRO-VALUES" (for Mars: Park et al. 2021 Table 2, adopting Konopliv et al. 2016), but the planet/satellite GMs (`BODY499/401/402`) from "natural satellite file release forms" — the older satellite-ephemeris solutions. So `BODY4_GM` (42828.3758157561) ≠ `BODY499+401+402` (42828.3744256094) by ~1.39×10⁻³ km³/s² (~3×10⁻⁸ relative) because they are different estimation epochs, with no internal-consistency requirement (the 2025 mar099 solution carries yet a third planet value). brahe's pairing is the self-consistent one — `GM_MARS_SYSTEM` with DE-kernel barycenter positions (DE440 was integrated with that value), `GM_MARS` with satellite-kernel body-center dynamics — now documented on the constants with guard tests locking in the known offset.

PR #395, opened 2026-07-15
Align docs navs, document TimeSystem, rename IGNORE flag to NETWORK

duncaneddy:
# Pull Request

## Description

Documentation and tooling maintenance, in three independent workstreams.

**Navigation.** The User Guide and Python API Reference navs had drifted into two different orderings and two vocabularies for the same subjects — a reader moving between them saw "Access Computation" in one and "Access" in the other. Both now follow a single section order and a single set of names. Ordering and titles only: no page content, no file moves, no URL changes.

**TimeSystem.** `TimeSystem` was documented inside `constants/units.md`, a page about unit and format enumerations, and its `members:` list enumerated only five of the ten variants — so `TDB`, `TCG`, `TCB`, `BDT`, and `GST` rendered nowhere at all. It now has its own page under `library_api/time/` covering all ten, and the Learn time-scale table and the enum's own overview rustdoc were completed to match.

While documenting it, two things turned out to be wrong in the source rather than the docs. `Epoch::time_system` told readers to "convert between systems using `to_time_system()`" — a method that had never been implemented. Rather than delete the promise, this PR honours it: because every constructor stores its instant in TAI and `time_system` only selects the display scale, the conversion is a single field change, and the new method is covered by parity tests asserting it agrees with the `*_as_time_system` projections across all ten scales. Separately, the TCG secular drift was stated as `~0.7 s/year`; IAU 2000 Resolution B1.9 gives L_G = 6.969290134e-10, or ~22 ms/year — off by a factor of ~32.

**Example flags.** The `IGNORE` flag named no reason to skip. Every flag means "skipped by default", so `IGNORE` said only "skipped by default, because it is skipped by default" and collided with `MANUAL`, `CI-ONLY`, and `SLOW` instead of standing beside them. #366 already identified it as "an overloaded bucket" and split `MANUAL` out; this finishes that work. Every remaining file carrying it depends on a live third-party service, which is what the changelog for #366 already called "the live-network set" — so the flag is now `NETWORK`, and each of the four flags names exactly one reason:

| Flag | Reason to skip |
|---|---|
| `MANUAL` | Never runs automatically, not even with a flag |
| `CI-ONLY` | Needs a CI environment |
| `SLOW` | Long runtime |
| `NETWORK` | Needs a live third-party service |

#366 swept only `examples/`, so the two files under `plots/` still carried the pre-#366 flag despite not being network-driven: `fig_comparative_benchmarks.py` reads committed artifacts and is already driven directly by the benchmark recipe, and `fig_access_benchmark.py` shells out to a compiled Rust binary. Both are now `MANUAL`, and their docstrings no longer point at a runner that is not how either is actually driven.

The rename also surfaced a latent bug worth calling out on its own. The runner parsed only the first ten lines of a file for `FLAGS`, and `lunar_orbit.rs` and `mars_orbit.rs` declare theirs on line 13 — so their flag had never taken effect. Both were running on every default `just test-examples`, downloading the GRGM660PRIM gravity model and the `moon_pa_de440` kernel, which is precisely what the flag exists to prevent. Their `.py` twins, with the flag on line 3, were correctly skipped; that split pair was the tell. The runner now parses the whole leading comment block, and a new fixture pair locks the fix in.

`CHANGELOG.md` is untouched, since the template records that entries are aggregated from merged PR bodies at release time.

## Changelog

**Fill in at least one section below.** At release time, these entries are aggregated across all merged PRs to produce `CHANGELOG.md` and the GitHub Release notes; each is attributed to its author and PR. Sections follow the [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) convention; see the [versioning policy](https://duncaneddy.github.io/brahe/latest/about/versioning/) for how deprecations and removals flow across releases.

### Added
- `Epoch.to_time_system()` (Rust and Python) returns a new `Epoch` for the same instant expressed in a different time system. The instant is unchanged, so the returned epoch compares equal to the original; only the scale it reports in differs. The method had been referenced by the `Epoch::time_system` documentation but was never implemented.
- `TimeSystem` API reference page under `library_api/time/`, documenting all ten time systems (`GPS`, `TAI`, `TT`, `UTC`, `UT1`, `TDB`, `TCG`, `TCB`, `BDT`, `GST`). It previously lived under `constants/units.md` and listed only five, so `TDB`, `TCG`, `TCB`, `BDT`, and `GST` did not render at all.
- "Time System" section in the Learn Epoch guide, covering how an epoch's time system is set and read, and the distinction between converting an epoch and projecting one, with runnable Rust and Python examples.
- Example fixtures covering `FLAGS` declared past the tenth line of a file.

### Changed
- Renamed the `IGNORE` example flag to `NETWORK`, completing the split begun in #366. `IGNORE` named no reason to skip, while every file carrying it depends on a live third-party service. `--ignore` becomes `--network` on `just test-examples` and `just make-plots`.
- Reclassified `plots/fig_access_benchmark.py` and `plots/fig_comparative_benchmarks.py` from `IGNORE` to `MANUAL`. Neither is network-driven: both regenerate committed artifacts and require a custom environment, and neither is driven by `just make-plots`.
- Aligned the User Guide and Python API Reference navigation to a single section ordering and vocabulary.
- The Learn time-scale table and the `TimeSystem` overview rustdoc now document all ten time systems rather than five.

### Deprecated / Future Warning
- 

### Removed
- 

### Fixed
- `FLAGS` declared past the tenth line of an example were silently ignored, because the runner parsed only the first ten lines. `lunar_orbit.rs` and `mars_orbit.rs` declare theirs on line 13, so both ran on every default `just test-examples` — downloading the GRGM660PRIM gravity model and the `moon_pa_de440` kernel — despite being flagged, while their `.py` twins were correctly skipped. The runner now parses the whole leading comment block.
- Corrected the TCG secular drift from `~0.7 s/year` to `~22 ms/year` (IAU 2000 Resolution B1.9, L_G = 6.969290134e-10).
- Corrected the `Epoch::time_system` documentation, which listed five of the ten time systems and directed readers to a `to_time_system()` method that did not exist.
- Corrected typos in the `TimeSystem` rustdoc ("supposed" to "supported", "expeccted" to "expected").


docs/library_api/time/time_system.md
```diff
@@ -0,0 +1,58 @@
...
+        - UT1
+        - TDB
+        - TCG
+        - TCB
+        - BDT
+        - GST
+      show_bases: false
+      heading_level: 3
+
+### Time System Descriptions
```
duncaneddy: This should be dropped from the library_api docs. Just need the actual API documentation.

docs/library_api/time/time_system.md
```diff
@@ -0,0 +1,58 @@
...
+- **TDB** (Barycentric Dynamical Time): Time scale for solar system barycentric ephemerides. Differs from TT by small periodic terms (< 1.7 ms) from the relativistic effects of Earth's orbital motion. TDB seconds past J2000 is SPICE ephemeris time (ET).
+- **TCG** (Geocentric Coordinate Time): Coordinate time for geocentric reference systems. Differs from TT by a secular drift (~22 ms/year) from Earth's gravitational time dilation.
+- **TCB** (Barycentric Coordinate Time): Coordinate time for the solar system barycenter. Differs from TDB by a secular drift from relativistic effects.
+- **BDT** (BeiDou Navigation Satellite System Time): Atomic time scale aligned with UTC at inception (January 1, 2006). Fixed offset from TAI of 33 seconds (BDT = TAI - 33s).
+- **GST** (Galileo System Time): Atomic time scale for the Galileo navigation system. Steered to GPS time, sharing the same fixed offset from TAI of 19 seconds (GST = TAI - 19s).
+
+### Usage Example
```
duncaneddy: Shouldn't have usage examples in library API docs

docs/library_api/time/time_system.md
```diff
@@ -0,0 +1,58 @@
...
+      show_root_heading: true
+      show_root_full_path: false
+      members:
+        - GPS
+        - TAI
+        - TT
+        - UTC
+        - UT1
+        - TDB
+        - TCG
```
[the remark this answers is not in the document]
duncaneddy: Fixed in d8db77c5 — but the stated mechanism doesn't hold, so flagging it for the record.

The page never rendered the PyO3 member docstrings. Because of the `#[classattr]` enum pattern, the `///` docs attach to `fn TCG() -> Self`, and `#[classattr]` evaluates that to an instance — so the docstring is discarded and each member inherits the *class* docstring. Checking the built HTML before this change, all ten members rendered the same text ("Enumeration of supported time systems...") and `~0.7 s/year` appeared nowhere on the page. It's the same reason the stub generator flattens every member to the class docstring.

That said, the value was genuinely wrong for anyone reading the binding source, so it's corrected to `~22 ms/year` (IAU 2000 Res. B1.9, L_G = 6.969290134e-10 x Julian year = 21.99 ms/yr). That was the last surviving copy — `0.7 s/year` no longer appears anywhere in the repo.

docs/library_api/time/time_system.md
duncaneddy: Done in d8db77c5 — descriptions removed. The page is now just the `::: brahe.TimeSystem` directive plus See Also, matching the `CLAUDE.md` convention ("Library API: Only `::: brahe.ClassName` directives + See Also"). Carrying them over from `constants/units.md` was my mistake.

Worth knowing what this leaves, though: because of the `#[classattr]` pattern, the directive renders the same generic class docstring for all ten members rather than per-member text, so the page no longer carries any per-system detail. That detail now lives only in the Learn time-scale table (which this PR completes to all ten systems) and is linked from See Also — which I think is the right Diataxis split, but it's a real change in what the page tells a reader.

docs/library_api/time/time_system.md
duncaneddy: Done in d8db77c5 — usage example removed.

Heads up that four other library_api pages carry the same pattern: `constants/units.md`, `eop/static_provider.md`, `eop/functions.md`, and `propagators/sgp_propagator.md`. All pre-existing, so I've left them alone rather than widen this PR — happy to clean them up separately if you want.

examples/test_network_late_flag.py
```diff
@@ -0,0 +1,18 @@
+# /// script
```
duncaneddy: Removed in d8db77c5.

For the record on what's still covered: the parser fix itself stays verified in place, since `lunar_orbit.rs` and `mars_orbit.rs` carry their `FLAGS` on line 13 and both now classify as `(True, 'network')` — they exercise the past-line-10 path as real examples. What goes away with these fixtures is the *guard*: they were the thing that would fail if someone re-narrowed the parse window, whereas the lunar/mars files would just silently start hitting the network again, which is exactly how this went unnoticed the first time. Easy to reinstate if you'd rather have that tripwire.

duncaneddy: Correcting myself here — you're right, and my "tripwire" framing was wrong.

The fixture's body was a bare `println!`. If the parser regressed and it stopped being skipped, it would just run, print, exit 0, and **pass**. Nothing would have failed. There was no trip in the tripwire — it was only ever a one-time demonstration (revert the parser by hand, eyeball the classification), not a regression test. A real guard would be a unit test asserting `check_flags(late_flag_file) == (True, "network", None)`, which is a different thing and not something `scripts/` has infrastructure for.

And on the broader point: the fix replaced an arbitrary constant (`[:10]`) with the structurally correct rule (parse the leading comment block), which isn't the kind of thing that silently reverts. Inventing files to guard against it was speculative. They stay deleted.

docs/library_api/time/time_system.md
duncaneddy: Done in 2c0d8ea6 — swept all of library_api, not just the pages I listed earlier (my original count was too low; it only matched literal "Usage Example" headings). Removed from `constants/units.md`, `constants/index.md`, `eop/index.md`, `eop/file_provider.md`, `eop/static_provider.md`, `eop/functions.md`, and `propagators/sgp_propagator.md` — 334 lines.

This turned out to be more than a style fix. The examples were never executed, so they had rotted:

- `sgp_propagator.md` called `prop.propagate(...)` and `prop.propagate_multiple(...)` — **neither method exists** on `SGPPropagator`. Two of the example's four operations would raise `AttributeError`.
- `static_provider.md` passed a velocity vector to `position_ecef_to_eci`, which is the wrong transform for velocity.

Examples under `examples/` can't drift like this because `just test-examples` runs them.

Two things I left, happy to remove if you'd rather: the two-line `from brahe.spacetrack import operators as op` snippets in `ephemeris/shared_types.md` and `ephemeris/spacetrack/operators.md` (those functions render under their `brahe._brahe.*` paths but are used via `brahe.spacetrack.operators`, so the snippet is the only thing connecting the two), and the `MJD_ZERO = JD - MJD` formula in `constants/time.md`, which is a definition rather than an example.

PR #397, opened 2026-07-15
Add FES2004 ocean tides and pole tides

duncaneddy:
# Pull Request

## Description

Implements ocean tides and pole tides per IERS Conventions (2010), TN36 Chapter 6, completing the tidal force model alongside the existing solid Earth tides:

- **FES2004 ocean tides** (§6.3, Eq. 6.15): 18 main constituents plus 63 secondary waves completed by linear admittance interpolation (Eq. 6.16, Table 6.7), configurable to degree/order 100. Coefficients come from a one-time cached download of the IERS `fes2004_Cnm-Snm.dat` file into `$BRAHE_CACHE/tides/` with integrity validation (truncated downloads are rejected before caching; corrupt cache files are removed with a re-download hint).
- **Solid Earth pole tide** (§6.4) and **ocean pole tide** dominant (2,1) term (§6.5, Eq. 6.24), driven by wobble parameters relative to the IERS 2018 linear secular pole (updated Conventions §7.1.4, Eqs. 21/25). Both require initialized EOP data, checked at propagator construction.
- **Unified fold-in evaluation**: all tidal ΔC̄nm/ΔS̄nm corrections and the static gravity coefficients are folded into a single propagator-owned Clenshaw coefficient table and evaluated in one pass per dynamics call. With spherical-harmonic gravity this replaces the separate static evaluation, so enabling tides costs only the per-epoch delta computation. Exact by linearity of the geopotential in its coefficients.
- To keep the crates.io package under its 10 MB limit (it was at 9.5 MB), the bundled EGM2008 model is truncated to degree/order 120 (`GravityModelType::EGM2008_120`), reducing the package to ~7.7 MB. Coefficients through degree 120 are bit-identical to the previous file; higher degrees remain available through the ICGEM download path.

Two transcription errata in the printed TN36 Table 6.7 (pivot Doodson numbers `135.455` and `237.755`) were identified by visual verification of the source PDF and corrected to `135.655` (Q1) and `235.755` (2N2); every pivot is test-asserted to be one of the 18 FES2004 file constituents. The Eq. 6.15 real-form expansion and Doodson→Delaunay conversion were cross-checked against the Orekit implementation.

### Added
- FES2004 ocean tide model (`OceanTideModel`, `OceanTideConfig`) with admittance-wave expansion per IERS TN36 §6.3, configurable degree/order up to 100, and one-time cached download of the IERS coefficient file
- Solid Earth pole tide (`SolidTideConfig.pole_tide`) and ocean pole tide (`OceanTideConfig.pole_tide`) per IERS TN36 §6.4/§6.5 with the IERS 2018 linear secular pole model (`secular_pole`, `wobble_parameters`, `solid_earth_pole_tide_deltas`, `ocean_pole_tide_deltas`)
- `TidesConfiguration.ocean` field enabling ocean tides in the numerical propagator force model; `ForceModelConfig::high_fidelity()` now enables solid tides with frequency-dependent corrections, both pole tides, and 30×30 ocean tides
- Python bindings for `OceanTideConfig` and the new `pole_tide` / `ocean` configuration fields, with regenerated type stubs

### Changed
- All tidal corrections (solid Earth Step 1/2, pole tides, ocean tides) and the static gravity field are now evaluated in a single fold-in Clenshaw pass per dynamics call, replacing the separate solid-tide evaluation path
- Bundled EGM2008 gravity model truncated from degree/order 360 to 120 and renamed `GravityModelType::EGM2008_360` → `EGM2008_120` to keep the crates.io package within size limits (bit-identical coefficients through degree 120; higher degrees available via ICGEM)
- Ocean tide force-model documentation (`docs/learn/orbital_dynamics/tides.md`) expanded to cover ocean tides, admittance waves, pole tides, and the download/cache behavior

### Removed
- `GravityModelType::EGM2008_360` (renamed to `EGM2008_120`; not deprecated — the variant was removed outright since the underlying packaged file changed)
- Internal fixed-size (degree ≤ 4) tidal coefficient evaluator (`TideCoefficients`, `accel_low_degree_harmonics`), superseded by the dynamically sized `TideDeltas` accumulator evaluated through the shared Clenshaw kernel

## Note to Reviewers
- Physics anchors worth spot-checking: the ocean pole tide ΔS̄21 leading factor is **−1.7232×10⁻¹⁰**, deliberately different from the ΔC̄21 factor (−2.1778×10⁻¹⁰) per Eq. 6.24 — easy to "correct" wrongly. The FES2004 file contains no T2 constituent despite the TN36 prose listing it; T2 arrives via admittance.
- Unit tests run fully offline: FES2004 parse/propagation tests use a committed degree-30 fixture under `test_data/`; the real download is exercised by an `integration`-gated test.
- `just check` is green (Rust 4746 unit tests + 721 doctests, Python 3169 tests, clippy/fmt/stubs/docs). Space-weather examples were failing at the time of development due to a CelesTrak outage unrelated to this branch.
- Serde note: previously serialized `TidesConfiguration` values load with `ocean: None` (missing `Option` keys deserialize to `None`); this back-compat behavior is pinned by test.


examples/numerical_propagation/force_model_tides.rs
```diff
@@ -1,6 +1,12 @@
 //! Solid Earth tidal corrections to the geopotential.
 //! Propagates one LEO orbit with tides ON and OFF, and reports the peak
 //! position difference to show the tidal perturbation magnitude.
+//!
+//! Brahe also supports FES2004 ocean tides (`OceanTideConfig`) and both the
```
duncaneddy: use ocean-tides but keep to 30x30 to avoid the download.

examples/numerical_propagation/tides_full.py
```diff
@@ -0,0 +1,30 @@
+"""Force model configuration with the full tide model: solid Earth tides
```
duncaneddy, line 4: This appears to be missing the required dependencies and flags header.

examples/numerical_propagation/tides_full.rs
```diff
@@ -0,0 +1,40 @@
+//! Force model configuration with the full tide model: solid Earth tides
```
duncaneddy, line 1: This appears to be missing the required dependencies and flags header.

tests/orbit_dynamics/test_ocean_tides.py
```diff
@@ -0,0 +1,100 @@
...
+        [
+            bh.R_EARTH + 800e3,
+            0.001,
+            np.radians(98.7),
+            np.radians(30.0),
+            np.radians(40.0),
+            0.0,
+        ]
+    )
+    return bh.state_koe_to_eci(oe, bh.AngleFormat.RADIANS)
```
duncaneddy: Don't convert with `np.radians` then use `bh.AngleFormat.RADIANS`, just use `bh.AngleFormat.DEGREES` initially.

tests/orbit_dynamics/test_gravity.py
```diff
@@ -515,10 +515,26 @@ def test_clenshaw_matches_cunningham(self):
...
-        """Cunningham kernel raises on V/W overflow at high degree, low altitude."""
-        model = bh.GravityModel.from_model_type_with_coefficients(
-            bh.GravityModelType.EGM2008_360, bh.GravityModelCoefficients.Both
+    def test_cunningham_overflow_raises(self, tmp_path):
+        """Cunningham kernel raises on V/W overflow at high degree, low altitude.
+
+        The overflow comes from the V/W recursion itself (a function of
+        geometry and degree only, not the coefficient values), so a synthetic
+        degree-160 model is used: the packaged EGM2008_120 model is truncated
+        to degree 120 and can no longer exercise this degree at all.
```
duncaneddy: Can we just use the bundled GGM05S model which goes up to degree 180 instead of introducing synthetic data?

src/propagators/dnumerical_orbit_propagator.rs
```diff
@@ -204,6 +204,286 @@ impl RotationCache {
...
+/// gravity coefficients (when spherical-harmonic gravity is configured) plus
+/// all enabled tidal ΔC̄nm/ΔS̄nm corrections, folded into one packed Clenshaw
+/// table and evaluated in a single pass. Exact by linearity of the
+/// geopotential in its coefficients.
+///
+/// Shared behind `Arc<Mutex<...>>` by the dynamics closure, like
+/// [`RotationCache`]; per-epoch delta computation is skipped when consecutive
+/// integrator stages share an epoch (`last_t`, keyed on the relative time's
+/// bit pattern like the rotation cache).
+struct TideField {
```
duncaneddy: Instead of including this in this file which grows it even more, would be helpful and self-contained to break out it's it's own file with `TideField` used here.

src/propagators/dnumerical_orbit_propagator.rs
```diff
@@ -204,6 +204,286 @@ impl RotationCache {
...
+use crate::orbit_dynamics::ParallelMode;
+use crate::orbit_dynamics::gravity::{ClenshawCoefficients, clenshaw_acceleration};
+use crate::orbit_dynamics::ocean_tides::OceanTideModel;
+use crate::orbit_dynamics::tides::{
+    TideDeltas, doodson_delaunay_args, ocean_pole_tide_deltas, solid_earth_pole_tide_deltas,
+    solid_earth_tide_deltas, wobble_parameters,
+};
+use crate::propagators::force_model_config::OceanTideConfig;
+
+/// Propagator-owned tidal geopotential field (IERS TN36 Ch. 6): the static
```
duncaneddy: This sentence "the static gravity coefficients (when spherical-harmonic gravity is configured)" makes it read like this object contains the static gravity field coefficients, which I don't think is the case because those should still be coming from the defined Spherical Harmonics gravity model (which should just be loaded once, not have its tabled duplicated inside the class and inside the gravity model.

I do see the benefit of adding the two together to just do a single pass through the coefficients though.

It also seems that combining them makes no sense because you can't separate out the time-dependent tide part (which changes each evaluation) from the static part. So you need to keep them separate anyways. If this is just working memory and the true static tables are propagator owned and these are just for combining for doing corrections that seems fine.

src/propagators/dnumerical_orbit_propagator.rs
```diff
@@ -1438,7 +1721,7 @@ impl DNumericalOrbitPropagator {
         params: DVector<f64>,
         additional_dynamics: Option<DStateDynamics>,
         gravity_model: Option<Arc<GravityModel>>,
-        solid_tide: Option<SolidTideConfig>,
+        tide_field: Option<Arc<Mutex<TideField>>>,
```
duncaneddy, line 1589: Is there a reason this needs a Mutex at all? While `par_propagate_to` parallelizes over threads Each propagator should be primarily executed by a single thread. While the gravity model computation can be parallelized that is internal to that single call. Given that the tide_field is owned by a single propgator this appears overly defensive and would be improved by just using an Option<Arc<TideField>>.

src/propagators/dnumerical_orbit_propagator.rs
```diff
@@ -1643,60 +1926,67 @@ impl DNumericalOrbitPropagator {
...
             let r_ecef = r_i2b * r;
-            // TODO: when third_body uses a high-precision ephemeris (DE440/DE440s), source Sun/Moon here from the same provider for per-step consistency. Low-precision is acceptable for the tidal perturbation magnitude.
-            let r_sun_eci = sun_position(epoch);
-            let r_moon_eci = moon_position(epoch);
-            let r_sun_ecef = r_i2b * r_sun_eci;
-            let r_moon_ecef = r_i2b * r_moon_eci;
-            let (gm_t, rad_t) = match gravity_model {
-                Some(m) => (m.gm, m.radius),
-                None => (GM_EARTH, R_EARTH),
+            // TODO(kept from previous implementation): when third_body uses a
```
duncaneddy: We should add `EphemerisSource` option here (same from ThirdBody) to the Tidal Configuration to enable selection of high-precision/low-precision in how the position is computed with default for low-precision.

Should also look at implementing caching of sun/moon position if enabled for both third-bodies and tides from the same source so that they only need be computed once and then reused.

src/orbit_dynamics/ocean_tides.rs
```diff
@@ -0,0 +1,1061 @@
...
+///
+/// * `PathBuf` - Location of `fes2004_Cnm-Snm.dat` inside `$BRAHE_CACHE/tides/`.
+///
+/// # Errors
+///
+/// Returns `BraheError` if the tides cache directory cannot be created, or if
+/// no cached copy exists and the download from IERS fails. The error message
+/// names the URL and the target cache path so the file can be fetched
+/// manually to proceed offline.
+pub fn fes2004_coefficients_path() -> Result<PathBuf, BraheError> {
```
duncaneddy, line 50: This seems like it's reinventing a lot of utils::download which could just be used instead.

src/orbit_dynamics/ocean_tides.rs
```diff
@@ -0,0 +1,1061 @@
...
+    /// # Example
+    ///
+    /// ```no_run
+    /// use brahe::OceanTideModel;
+    ///
+    /// // Downloads FES2004 coefficients into the cache on first use.
+    /// let model = OceanTideModel::from_cache(20, 20, true).unwrap();
+    /// assert_eq!(model.n_max(), 20);
+    /// ```
+    pub fn from_cache(
```
duncaneddy: Instead of "from cache" this seems like it should just be the default `new` behavior. And `from_file` allows for loading other files.

src/orbit_dynamics/ocean_tides_admittance.rs
```diff
@@ -0,0 +1,539 @@
+/*!
+IERS TN36 Table 6.7: astronomical amplitudes Hf of the FES2004 main waves and
+the secondary (admittance) waves with their linear-interpolation pivot waves.
+
+Source: IERS Conventions (2010), TN36, Chapter 6, pp. 92-93.
+URL: <https://iers-conventions.obspm.fr/content/chapter6/icc6.pdf>
+
+Transcription notes (verified against the PDF at 600 DPI):
```
duncaneddy: Drop comment of (Verified against PDF at 600 DPI), it's extraneous.

examples/numerical_propagation/force_model_tides.rs
duncaneddy: Done in 0376996e — ocean tides enabled at 30×30 with admittance and both pole tides. One note: the FES2004 download is triggered by any ocean-tide propagator construction regardless of degree (the degree only truncates the parse), so the example now fetches the 3.7 MB file once into $BRAHE_CACHE/tides/. Since this example already requires network for initialize_sw(), it carries the NETWORK flag and its doc note now states the download behavior.

examples/numerical_propagation/tides_full.py
duncaneddy: Fixed in 0376996e — added the `# /// script` dependency header matching the sibling examples, and the docs include offset was updated accordingly.

examples/numerical_propagation/tides_full.rs
duncaneddy: Fixed in 0376996e. Note the sibling convention: Rust examples only carry a `//! \`\`\`cargo` block when they need crates beyond brahe itself; tides_full.rs uses only brahe, so it gets the standard doc-comment header without a cargo block (same as the other brahe-only examples).

tests/orbit_dynamics/test_ocean_tides.py
duncaneddy: Fixed in 0376996e — elements are passed in degrees with bh.AngleFormat.DEGREES directly; the np.radians conversion and numpy usage are gone.

tests/orbit_dynamics/test_gravity.py
duncaneddy: Done in 0376996e — both the Python test and its Rust twin (test_cunningham_high_degree_overflow_errors) now use the bundled GGM05S (n_max 180, asserted ≥160 in the test) instead of the synthetic header-only model; the non-finite-error and success-branch assertions are unchanged.

src/propagators/dnumerical_orbit_propagator.rs
duncaneddy: Done in bf49a06c — TideField (struct, impl, and its unit tests) moved to src/propagators/tide_field.rs as a pub(crate) module; the propagator file shrank by ~450 lines and only consumes the type.

src/propagators/dnumerical_orbit_propagator.rs
duncaneddy: Confirmed — it is exactly the working-memory case you describe: the gravity model stays the single source of truth (loaded once, never mutated; global models are read under the lock only at construction), and TideField holds a one-time copy of the truncated static coefficients as the fold baseline so each epoch's static+delta evaluation is a single Clenshaw pass. Memory is bounded by the evaluation truncation, not the model's full degree. The rustdoc was rewritten in e3c4aefe to say this explicitly instead of implying ownership of the static field.

src/propagators/dnumerical_orbit_propagator.rs
duncaneddy: Checked this and the Mutex is load-bearing rather than defensive: the shared dynamics closure type is Arc<dyn Fn .. + Send + Sync> (it is also captured by the Jacobian provider for STM/sensitivity), so everything it captures must be Sync — and TideField::acceleration needs &mut self for the per-epoch fold-in (tables + last_t). A plain Option<Arc<TideField>> can't provide &mut through an Arc, and RefCell isn't Sync, so it doesn't compile without some lock type. This is the same constraint that makes RotationCache Arc<Mutex<..>> (see its comment); in the single-thread-per-propagator case the lock is uncontended (~tens of ns next to a Clenshaw pass). Expanded the comment at the construction site in e3c4aefe to record this. Happy to revisit if we ever relax the Sync bound on DStateDynamics.

src/propagators/dnumerical_orbit_propagator.rs
duncaneddy: Implemented in e3c4aefe — TidesConfiguration gains ephemeris_source: EphemerisSource with serde default LowPrecision (missing-key back-compat pinned by test), wired so tidal Sun/Moon positions resolve through the configured source (SPICE sources validate/resolve at construction, fail-fast). Also added the shared epoch-keyed Sun/Moon cache: both the tides block and the third-body block consult it when their sources match, so the positions are computed once per epoch and reused (third-body results are bitwise-identical when sources differ — the cache is simply bypassed). high_fidelity() now sets DE440s for tides to match its third-body config so the two share the evaluation. Python binding, stubs, and docs updated.

src/orbit_dynamics/ocean_tides.rs
duncaneddy: Done in 0376996e — the hand-rolled ureq call is replaced with utils::download::download_bytes (which adds retries), keeping the pieces it doesn't cover: the in-process download mutex with post-lock existence recheck, the ≥59,000-line pre-promotion validation of the payload, atomic promotion, and the URL+cache-path error message for manual fallback.

src/orbit_dynamics/ocean_tides.rs
duncaneddy: Renamed in 0376996e — OceanTideModel::new is the default cached-download constructor and from_file loads arbitrary files. No alias kept; call sites, rustdoc, and the learn page updated.

src/orbit_dynamics/ocean_tides_admittance.rs
duncaneddy: Dropped in 0376996e — the module doc keeps the factual erratum note (misprinted pivots corrected against the pivot-must-be-a-main-wave invariant) without the verification-process aside.

src/orbit_dynamics/ocean_tides.rs
```diff
@@ -0,0 +1,1061 @@
...
+        Self::validate_truncation_bounds(n_max, m_max)?;
+        let path = fes2004_coefficients_path()?;
+        // A truncated download is rejected before caching (see
+        // `fes2004_coefficients_path`), but a pre-existing cache file may have
+        // been corrupted out-of-band (partial write from an older build,
+        // manual tampering, disk error). Validate it and, on failure, remove
+        // it so the next call re-downloads a fresh copy.
+        Self::from_file(&path, n_max, m_max, include_admittance)
+            .and_then(|model| {
+                model.validate_main_waves()?;
```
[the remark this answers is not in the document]
duncaneddy: Fixed in 0376996e — the cached-file validation now also checks degree coverage: after parsing, the maximum parsed degree must equal the requested n_max (the genuine IERS file covers degree 100, so any shortfall means a truncated cache). On failure the cache file is deleted with a re-download hint, same as the other corruption paths. Regression test added: seeding the cache with a degree-30 file and requesting (50, 50) errors and removes the file, while requests within the cached coverage keep working.

src/propagators/dnumerical_orbit_propagator.rs
duncaneddy: Other configurations don't use support Serde, so we shouldn't introduce it as a default, make the breaking change now.

duncaneddy: Removed in 1f232581 — ephemeris_source now carries no serde default, matching the rest of the configuration family. Loading a config serialized before the field existed errors until an explicit key is added; the breaking behavior is pinned by test (test_tides_configuration_missing_ephemeris_source_errors) and noted in the tides guide. The Rust Default impl and the Python constructor kwarg still default to LowPrecision (those are construction defaults, not serde).

examples/numerical_propagation/tides_full.rs
duncaneddy: Correction to my earlier reply: 0376996e did not touch tides_full.rs — no change was needed, since the sibling convention is that brahe-only Rust examples carry the plain //! doc header without a cargo block (verified against tides_static.rs, tides_permanent_only.rs, force_model_overview.rs). The 'Fixed in 0376996e' prefix was wrong for this file; only the .py twin needed the header.

PR #399, opened 2026-07-16
Add planetary examples, textures, and generic synodic frames

duncaneddy:
# Pull Request

## Description

Adds four long-form worked examples — LRO lunar orbit, an Earth-Moon free-return trajectory, MRO Mars orbit, and Dawn-at-Ceres as a fully user-defined rotating body — together with the infrastructure they need: a generic `ReferenceFrame::Synodic` variant for arbitrary two-body rotating frames, downloadable Solar System Scope planet textures (CC BY 4.0, credited in the license page), a `central_body`/`additional_bodies` generalization of `plot_trajectory_3d`, and a `plot_synodic_3d` plotter with a `plot_earth_moon_rotating_3d` alias.

All examples construct their initial states in body-equatorial inertial bases (rotated into the ICRF-aligned propagation frames), and each page explains that construction. Two small deviations from the original design: the planned `load_planet_texture` helper was folded into the more general `load_body_texture`, and the Ceres example uses point-mass gravity because ICGEM's only Ceres product is a degree-2519 crust-only forward model unsuitable for propagation (the page explains this).

### Added
- `ReferenceFrame.Synodic(origin, primary, secondary)` and `SynodicOrigin` for arbitrary two-body rotating frames, with `synodic_origin`/`synodic_primary`/`synodic_secondary` accessors (also populated for EMR/SER/GSE)
- `synodic_barycenter_id` and generalized synthetic-barycenter resolution in the frame router (Sun-Earth barycenter is now a special case of the generic encoding)
- Solar System Scope planet texture registry with `download_planet_texture` and generalized `load_body_texture` (CC BY 4.0, credited in the license page)
- `central_body` and `additional_bodies` parameters on `plot_trajectory_3d` for Moon/Mars/custom-body scenes, backed by a body-visuals registry
- `plot_synodic_3d` and `plot_earth_moon_rotating_3d` for rotating-frame trajectory visualization
- Worked examples: LRO lunar orbit, Earth-Moon free return, MRO Mars orbit, Dawn at Ceres (user-defined central body), with learn/API documentation for the new frames and plotting
- Standalone `Rx`/`Ry`/`Rz` matrix functions
- Planetary radius constants (`R_MERCURY` .. `R_NEPTUNE`)
- Tested learn examples for generic `ReferenceFrame.Synodic` frames and `plot_earth_moon_rotating_3d`

### Changed
- `plot_trajectory_3d`: `earth_texture` renamed to `texture`, `show_earth` renamed to `show_body`, and all parameters after `trajectories` are now keyword-only (breaking; all in-repo usages updated)
- `load_earth_texture` removed in favor of `load_body_texture`
- `just download-resources` now also warms the moon/mars/ceres textures
- `state_koe_to_eci_for_body`/`state_eci_to_koe_for_body` renamed to `state_koe_to_inertial_for_body`/`state_inertial_to_koe_for_body`, taking a `CentralBody` instead of a bare `gm` and referencing elements to the body's mean equator at J2000 instead of the ICRF axes
- Live-API examples now `NETWORK`-flagged


docs/learn/frames/synodic_frames.md
```diff
@@ -24,6 +24,41 @@ Primaries: Sun → Earth. Origin: the Sun-Earth barycenter. The SEB has no NAIF
...
+\text{GSE} \equiv \texttt{Synodic(Primary, 399, 10)}
+$$
+
+A `Barycenter` origin is computed analytically as the $GM$-weighted combination of the primary and secondary SPK states — the same scheme used for the Sun-Earth barycenter above — rather than read from a dedicated SPK ephemeris entry. Both bodies must therefore have packaged $GM$ constants (the Sun, the planets, the planetary barycenters, and the Moon). For the Earth-Moon pair, this computed barycenter differs from the SPK Earth-Moon Barycenter (NAIF ID 3) by well under a meter.
+
+The origin, primary, and secondary of any synodic frame (`Synodic`, `EMR`, `SER`, or `GSE`) are recoverable via the `frame.synodic_origin`, `frame.synodic_primary`, and `frame.synodic_secondary` properties, which return `None` for non-synodic frames.
+
+The generic frame has no dedicated `rotation_gcrf_to_synodic`-style functions; it is used through the frame router (`rotation_frame_to_frame`, `position_frame_to_frame`, `state_frame_to_frame`). The following example transforms a low Earth orbit state from GCRF into a Sun-Mars rotating frame:
+
+```python
```
duncaneddy: This should be implemented as a tested code block in both rust and python.

docs/learn/plots/synodic_plots.md
```diff
@@ -0,0 +1,72 @@
...
+
+See [Synodic Reference Frames](../frames/synodic_frames.md) for the axis construction and physical definition of each frame. Passing a non-synodic frame raises `ValueError`.
+
+## Reference Epoch
+
+`reference_epoch` sets the epoch at which the primary and secondary spheres are placed. It defaults to the first epoch of the first trajectory in `trajectories`, and must be supplied explicitly if `trajectories` is empty.
+
+## Example
+
+```python
```
duncaneddy: Should be tested code block / maybe a plots/* code block to generate and show the plot.

examples/examples/dawn_ceres_orbit.py
```diff
@@ -0,0 +1,239 @@
...
+ceres = bh.CentralBody.Custom(
+    "Ceres",
+    CERES_NAIF_ID,
+    GM_CERES,
+    radius=R_CERES,
+    omega=ceres_omega(),
+    fixed_frame=ceres_fixed,
+)
+
+# Gravity-only, point-mass force model: no third bodies or SRP, since the DE
```
duncaneddy: This is actually not fully true and a short-cut. We can look up Ceres ephemeris and generate an SPK file with the Horizons system (https://ssd-api.jpl.nasa.gov/doc/horizons.html). and load it, that way we can also get perturbation forces which are important for doing proper force simulation.

For now update the text and make a ticket to add a Horizons Client (simple small scope for just generating / downloading / loading specific SPK files for targeted ephemeris)

examples/examples/dawn_ceres_orbit.py
```diff
@@ -0,0 +1,239 @@
...
+CERES_W_RATE = 952.1532635  # deg/day
+# --8<-- [end:body_definition]
+
+# --8<-- [start:body_fixed_frame]
+# A body-fixed frame from the IAU-style pole/prime-meridian model, registered
+# as a custom frame: rotation(epoch) -> ICRF-to-body DCM.
+_J2000_TT_MJD = 51544.5
+
+
+def _rot_x(angle):
```
duncaneddy: We should not have to define these. This is a capability gap to address in the Utilities module. Rx/Ry/Rz shouldn't just be methods on RotationMatrix, but top level stand-alone functions in the rotation_matrix.rs file that are exported functions as well. This should just return an nalgrebra/numpy array (the RotationMatrix initializes should be kept for gettin a proper rotation matrix

examples/examples/dawn_ceres_orbit.py
```diff
@@ -0,0 +1,239 @@
...
+
+# --8<-- [start:orbit_setup]
+# Dawn LAMO (Low Altitude Mapping Orbit): ~375 km circular polar orbit.
+epoch = bh.Epoch.from_datetime(2016, 1, 1, 0, 0, 0.0, 0.0, bh.TimeSystem.UTC)
+
+# Ceres-equatorial inertial basis expressed in the ICRF-aligned propagation
+# axes. The spin pole comes straight from the IAU pole constants; x_eq is the
+# ascending node of the Ceres equator on the ICRF equator (ICRF pole x spin
+# pole) and y_eq completes the right-handed triad. B_eq_to_bci rotates
+# equator-referenced vectors into the propagation frame.
```
duncaneddy: Would be great for a source on how this is achieved. I want to better understand what the "ascending node of Ceres equator on ICRF" equator better means.

examples/examples/dawn_ceres_orbit.py
```diff
@@ -0,0 +1,239 @@
...
+_pole_ra = np.radians(CERES_POLE_RA)
+_pole_dec = np.radians(CERES_POLE_DEC)
+ceres_pole = np.array(
+    [
+        np.cos(_pole_dec) * np.cos(_pole_ra),
+        np.cos(_pole_dec) * np.sin(_pole_ra),
+        np.sin(_pole_dec),
+    ]
+)
+x_eq = np.cross([0.0, 0.0, 1.0], ceres_pole)
```
duncaneddy: I don't quite get how the geometry works out such that this cross-product defines x_eq.

examples/examples/dawn_ceres_orbit.py
```diff
@@ -0,0 +1,239 @@
...
+1. Define an entirely user-supplied central body (GM, radius, spin pole,
+   prime meridian) for a body brahe has no built-in support for
+2. Register a custom body-fixed frame from an IAU-style pole/prime-meridian
+   rotation model
+3. Propagate the Dawn spacecraft's LAMO (Low Altitude Mapping Orbit) around
+   Ceres using that body and frame
+4. Report the body-fixed state through the custom frame and visualize the
+   trajectory in 3D around a textured Ceres
+
+Unlike Earth, the Moon, and Mars, Ceres has no dedicated support in brahe:
```
duncaneddy: "no dedicated support" -> "no built-in constants" 

(We do support it through this mechanism, which is the point of the example to show, how we support arbitrary defined bodies.

brahe/plots/bodies.py
```diff
@@ -0,0 +1,66 @@
+"""Body visual registry for 3D plotting: radius, default texture, NAIF ID."""
+
+import brahe as bh
+
+# Radii: brahe constants where available, otherwise IAU/WGCCRE mean radii.
+BODY_VISUALS = {
+    "sun": {"radius": bh.R_SUN, "texture": "sun", "naif_id": 10},
+    "mercury": {"radius": 2439.7e3, "texture": "mercury", "naif_id": 199},
```
duncaneddy: Can we add these constants as proper planetary constants (aside from Ceres which should use the user-defined path/texture-provided, likely through radius/texture-path kwargs)

They should be added as constants.rs definitions grounded in a source using grounded references.

src/frames/transform.rs
```diff
@@ -84,8 +85,67 @@ use super::mars::rotation_mci_to_mcmf;
...
+/// [`SUN_EARTH_BARYCENTER_ID`] is this encoding evaluated at `(10, 399)`.
+///
+/// The full range at or below -1_000_000_000 is reserved for these synthetic
+/// barycenters — [`ReferenceFrame::BodyFixedCustom`] rejects self-assigned
+/// centers in that range so a user-registered custom frame can never collide
+/// with one.
+///
+/// # Arguments
+/// - `primary`: NAIF ID of the primary body. Must be in `0..=999`.
+/// - `secondary`: NAIF ID of the secondary body. Must be in `0..=999`.
```
duncaneddy: I don't think we want this to be restricted in ID, since that bars use of user-defined bodies and loaded kernels which is an unnecessary restriction.

src/propagators/force_model_config.rs
```diff
@@ -850,6 +850,10 @@ impl ForceModelConfig {
     /// 5. `GravityConfiguration::SphericalHarmonic` on a `CentralBody::Custom` body
     ///    requires `central_body.fixed_frame()` to be set (needed to rotate into the
     ///    body-fixed frame the harmonics are expressed in).
+    /// 6. `GravityConfiguration::SphericalHarmonic` as the central gravity field must
+    ///    model `central_body` itself — see
```
duncaneddy: Unclear wording. The spherical harmonic gravity field model must be a declared gravity field for the declared central body and match.

docs/examples/lro_lunar_orbit.md
```diff
@@ -0,0 +1,134 @@
+# LRO Lunar Orbit
+
+In this example we'll set up the Lunar Reconnaissance Orbiter's (LRO) low
+lunar science orbit and propagate it with brahe's lunar force model. LRO has
+flown a roughly 30 x 200 km class polar, near-frozen science orbit around
```
duncaneddy: I don't want to claim that this is the LRO orbit would a clear reference with stated epoch and elements. We can just say LRO-like orbit.

docs/examples/lro_lunar_orbit.md
```diff
@@ -0,0 +1,134 @@
...
+the Moon since 2009, mapping the surface at high resolution. We'll compare
+a full-fidelity propagation against a point-mass
+Moon to quantify how the Moon's lumpy gravity field ("mascons" - dense,
```
duncaneddy: Link to wikipedia page for mascons

docs/examples/lro_lunar_orbit.md
```diff
@@ -0,0 +1,134 @@
...
+concentrations rather than a smooth oblateness term. These mascons produce
+gravity anomalies strong enough that most low lunar orbits are unstable on
+timescales of weeks to months without station-keeping: eccentricity grows
+until the spacecraft impacts the surface. Orbit designers counter this by
+choosing a "frozen orbit" - an inclination and argument of perilune where the
+long-period perturbation from the gravity field's dominant harmonics
+averages to zero, so eccentricity and perilune altitude stay bounded over
+many orbits instead of drifting monotonically. LRO's polar, ~85 degree
+inclination orbit with perilune fixed over the southern hemisphere
+(argument of perilune 270 degrees) is one such configuration.
```
duncaneddy: This is too strong a statement regarding general design practices. I'd just end the paragraph with "orbits are unstable on
timescales of weeks to months without station-keeping. Proper modeling of the high-order gravity dynamics enable design of station-keeping maneuvers that minimize the growth of these instabilities, conserving fuel and extending mission-lifetime."

docs/examples/lro_lunar_orbit.md
```diff
@@ -0,0 +1,134 @@
...
+against the z-axis of whatever basis its output is read in, so interpreting
+its output directly as an LCI state would reference the 85.2 degree
+inclination and the south-pole perilune to the ICRF pole, not the lunar
+equator. To place the orbit correctly, we build a lunar-equatorial inertial
+basis - z-axis on the lunar mean pole (the third row of the LCI-to-LFME
+rotation), x-axis on the ascending node of the lunar equator on the ICRF
+equator - construct the state in that basis, and rotate its position and
+velocity into LCI. `state_koe_to_eci_for_body` converts the Keplerian elements
+to a Cartesian state using an arbitrary body's gravitational parameter, here
+`GM_MOON`, since `state_koe_to_eci` assumes Earth:
```
duncaneddy: So "state_koe_to_eci_for_body" is an escape that should have been previously caught but we need to fix it here. It doesn't make sense to have eci and for_body in the same function. these should be state_koe_to_inertial_for_body and state_interial_to_koe_for_body.

I don't get the argument that the function builds it's output against the z-axis, that should align with equatorial frame of the associated body (perpendicular to it). Regardless, if there is a discrepancy that should be fixed in the function so that the x-y axes properly align with the body equator on in the ICRF frame and the x-axis aligns with the vernal equinox as is convention for orbital elements.

docs/examples/lro_lunar_orbit.md
```diff
@@ -0,0 +1,134 @@
...
+(50x50 GRGM660PRIM gravity, SRP occulted by the Moon and Earth, and Earth/Sun
+third-body perturbations) and a point-mass Moon via `ForceModelConfig.for_body`
+with `GravityConfiguration.point_mass()`. Both propagators integrate in the
+Moon-Centered Inertial (LCI) frame:
+
+``` python
+--8<-- "./examples/examples/lro_lunar_orbit.py:propagation"
+```
+
+!!! note "state_bci vs. state_eci"
```
duncaneddy, line 69: This note is great! Using the user-guide to reference library functions that users might not be familiar with and explaining them is great. We should link out to the relevant learn or library_api functions when cited so users can also learn more.

docs/examples/lro_lunar_orbit.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
+``` python
+--8<-- "./examples/examples/lro_lunar_orbit.py:altitude_plot"
+```
+
+<div class="plotly-embed">
+  <iframe class="only-light" src="../figures/lro_lunar_orbit_altitude_light.html" loading="lazy"></iframe>
+  <iframe class="only-dark"  src="../figures/lro_lunar_orbit_altitude_dark.html"  loading="lazy"></iframe>
+</div>
```
duncaneddy, line 98: This plot is good but it's unclear whether it's the difference between point_mass and full-force, or full-force from true. That should be made clear through better axis labeling and plot title.

docs/examples/earth_moon_free_return.md
```diff
@@ -0,0 +1,131 @@
+# Earth-Moon Free-Return Trajectory
+
+In this example we'll design and fly an Earth-Moon **free-return trajectory**: a path that departs a low Earth parking orbit, coasts out to the Moon, and lets lunar gravity bend it back onto an Earth-return leg without any dedicated return burn. This is the trajectory class that gave the early Apollo missions their abort safety margin - if the service propulsion system had failed on the way out, the spacecraft would still have looped around the Moon and returned to a survivable re-entry. Apollo 13 depended on exactly this property after its oxygen tank ruptured, and Artemis I flew an extended "hybrid" free return uncrewed in 2022.
```
duncaneddy: Link out to the wikipedia page on it (https://en.wikipedia.org/wiki/Free-return_trajectory). Artemis I did not use a strict free-return trajectory. Artemis 2 did.

docs/examples/earth_moon_free_return.md
```diff
@@ -0,0 +1,131 @@
...
+
+The design has two competing requirements. The lunar flyby must pass close enough to the Moon to bend the path steeply back toward Earth, but not so close that the perilune drops below the surface. A pass that is too distant barely deflects the trajectory and the spacecraft escapes onto a distant loop or a heliocentric orbit instead of returning. The window between "escapes" and "impacts the Moon" is what makes free-return design a targeting problem rather than a closed-form calculation.
+
+## Geometry
+
+The departure geometry fixes everything about the transfer except its energy. We depart a 185 km circular parking orbit from a point near the **antipode** of the Moon's position at the expected arrival time, burning prograde in the Moon's instantaneous orbital plane so the transfer apogee reaches toward the Moon roughly half an orbit later.
```
duncaneddy: We can raise the departure altitude to 400km (ISS-like) if that helps reduce initial delta-v

docs/examples/earth_moon_free_return.md
```diff
@@ -0,0 +1,131 @@
...
+--8<-- "./examples/examples/earth_moon_free_return.py:preamble"
+```
+
+``` python
+--8<-- "./examples/examples/earth_moon_free_return.py:geometry"
+```
+
+## Force Model
+
+A free return is shaped by three bodies. We configure a point-mass Earth as the central body and add the Moon and Sun as third-body perturbers from the DE440s ephemeris. The lunar term is what bends the trajectory home; the Sun is a smaller but non-negligible perturbation accumulated over the eight-day flight. We deliberately omit Earth's oblateness, drag, and radiation pressure - none of them materially change the free-return geometry at these altitudes and distances.
```
duncaneddy: Would it not be better to do a 5x5 earth gravity model and do the computation in the EMBI frame?

docs/examples/earth_moon_free_return.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+``` python
+--8<-- "./examples/examples/earth_moon_free_return.py:distance_plot"
+```
+
+<div class="plotly-embed">
+  <iframe class="only-light" src="../figures/earth_moon_free_return_distance_light.html" loading="lazy"></iframe>
+  <iframe class="only-dark"  src="../figures/earth_moon_free_return_distance_dark.html"  loading="lazy"></iframe>
+</div>
```
duncaneddy, line 75: Plot embeddings should be centered.

docs/examples/earth_moon_free_return.md
```diff
@@ -0,0 +1,131 @@
...
+    A naive bisection on delta-v cannot converge here: the miss distance is not monotonic, so a bracket chosen blindly may straddle the bottom of the V, and the descending branch reaches the same perilune values without ever returning to Earth. The coarse scan finds the V's minimum so the bisection can be restricted to the ascending branch, where perilune increases monotonically with delta-v and the free-return solution for this geometry lives. The final propagation, terminated by the entry event, is what actually confirms the Earth return.
+
+## Flying the Mission
+
+We fly the tuned design as it would be flown. The propagator starts in the parking orbit, coasts one short arc, and applies the TLI impulsively through a `TimeEvent` callback that adds the delta-v along the velocity vector. A terminal `AltitudeEvent` at 120 km, triggered on decreasing altitude, stops the propagation at the atmospheric entry interface on the return leg - the point where a real capsule would begin re-entry. The flight time to that point falls out of the propagation rather than being prescribed.
+
+``` python
+--8<-- "./examples/examples/earth_moon_free_return.py:final_run"
+```
+
```
duncaneddy, line 59: Generally because we're not doing the right thing here (B-plane targeting w/ lambert solver) there's a lot of hackyness. Instead of discussing and justifying it, we should just say "right now this example is to show the EMB rotating frame, so we take some shortcuts to generate the trajectory, which in practice should not be done" and leave it at that.

docs/examples/earth_moon_free_return.md
```diff
@@ -0,0 +1,131 @@
...
+  <iframe class="only-light" src="../figures/earth_moon_free_return_emr_light.html" loading="lazy"></iframe>
+  <iframe class="only-dark"  src="../figures/earth_moon_free_return_emr_dark.html"  loading="lazy"></iframe>
+</div>
+
+*Body textures: [Solar System Scope](https://www.solarsystemscope.com/textures/), CC BY 4.0.*
+
+## Validation
+
+We confirm the trajectory is a genuine free return: the perilune passes above the lunar surface but within 20,000 km of the Moon's center, and the return leg descends below 1,000 km altitude, closing the loop back at Earth. Because the propagation stops at the 120 km entry interface, the minimum Earth distance measured is that entry crossing; the true geometric perigee lies below it, inside the atmosphere.
+
```
duncaneddy: The trajectory as generated is not a figure-8, it passes in front of the moon and is not a proper circum-lunar fly-by. This should be checked and fixed.

Additionally, we should have just a top-down (X/Y plane view of the trajectory in the EMR frame in 2D as well as the 3D version. Both plots would be improved by having arrows along the trajectory line showing the direction of travel.

docs/examples/earth_moon_free_return.md
```diff
@@ -0,0 +1,131 @@
...
+| Parameter | Value |
+|-----------|-------|
+| Parking orbit altitude | 185 km |
+| Transfer time (design) | 3.1 days |
+| Aim offset from antipode | 16 deg |
+| Target perilune radius | $R_{Moon} + 12{,}000$ km |
+| TLI delta-v | ~3.14 km/s |
+| Perilune altitude | ~12,000 km |
+| Return entry-interface altitude | ~120 km |
+| Flight time to re-entry | ~8.1 days |
```
duncaneddy: Never have hard-coded values. Generate as CSV (but I think better to just remove here and avoid the complexity.

docs/examples/mro_mars_orbit.md
```diff
@@ -0,0 +1,149 @@
+# MRO Mars Orbit
+
+In this example we'll set up the Mars Reconnaissance Orbiter's (MRO)
+sun-synchronous science orbit and propagate it with brahe's Mars force model.
+MRO has flown a ~255 x 320 km, 92.6 degree inclination orbit since 2006,
+imaging the surface at a consistent local solar time on every pass. We'll
+track how the osculating orbital elements evolve under the full force model
```
duncaneddy: Cite reference for orbit elements.

docs/examples/mro_mars_orbit.md
```diff
@@ -0,0 +1,149 @@
...
+relative to the Sun, so the spacecraft crosses each latitude at the same
+local solar time on every pass - valuable for consistent lighting in surface
+imagery. This only works because a planet's oblateness (the J2 zonal
+harmonic) causes the orbital plane to precess: for a near-polar, slightly
+retrograde inclination, the nodal precession rate can be tuned to exactly
+match the planet's mean motion around the Sun. At Mars, that inclination is
+about 92.6 degrees - close to polar, with the small excess over 90 degrees
+giving just enough retrograde precession to track Mars's slower (687-day)
+year. MRO's 255 x 320 km, 92.6 degree orbit is designed around this
+resonance.
```
duncaneddy: These values should be grounded in a calculation and equation.

docs/examples/mro_mars_orbit.md
```diff
@@ -0,0 +1,149 @@
...
+on the ICRF equator - construct the state in that basis, and rotate its
+position and velocity into MCI.
+
+The same step derives the ascending node from the Sun's direction rather than
+hardcoding it: the right ascension of the ascending node is set to the Sun's
+right ascension in the Mars-equatorial basis plus 45 degrees, placing the node
+near a 15:00 (mid-afternoon) local solar time. `state_koe_to_eci_for_body`
+converts the Keplerian elements to a Cartesian state using an arbitrary body's
+gravitational parameter, here `GM_MARS`, since `state_koe_to_eci` assumes
+Earth:
```
duncaneddy: As with the LRO orbit the definitions of state_koe_to_eci_for_body needs to be fixed.

The definition of the orbital elements that it returns should be consistent with the given body and the ICRF frame (which should be consistent inertial frame definition across bodies).

I would expect the output of the KOE<>inertial transformations would be directly composable with planet-specific state_bci_to_bcf style transforms.

docs/examples/mro_mars_orbit.md
```diff
@@ -0,0 +1,149 @@
...
+## Element Evolution
+
+Sampling the trajectory over the 2-day propagation and converting each
+Cartesian state back to Keplerian elements with `state_eci_to_koe_for_body`,
+then plotting the full six-element set with `plot_keplerian_trajectory`,
+shows how the orbit evolves under the full force model:
+
+``` python
+--8<-- "./examples/examples/mro_mars_orbit.py:element_history"
+```
```
duncaneddy, line 89: This plot is good but the boddom of the first row x-axis overlatps with the second row titles. Should be fixed.

docs/examples/mro_mars_orbit.md
```diff
@@ -0,0 +1,149 @@
...
+
+!!! note "state_bci vs. state_eci"
+    `state_bci` returns the propagator's native state in the central body's
+    body-centered inertial frame (MCI for a Mars-centered propagator).
+    `state_eci` instead always returns an Earth-centered state - for a
+    Mars-centered propagator it adds Mars's Earth-relative position, which
+    would report a semi-major axis near the Earth-Mars distance rather than
+    Mars orbit elements. Use `state_bci` whenever you need elements or
+    altitude relative to the body being orbited.
```
duncaneddy: Tip is useful but redundant with the LRO example tip. I think can be dropped.

docs/examples/dawn_ceres_orbit.md
```diff
@@ -0,0 +1,143 @@
...
+built-in `CentralBody` variant, no named inertial/fixed frame pair, no spin
+model. Everything about Ceres here - its gravitational parameter, radius,
+spin pole, prime meridian, and body-fixed frame - is supplied by the user via
+`CentralBody.Custom` and `register_custom_frame`. The same recipe applies to
+any body brahe doesn't know about: another dwarf planet, an asteroid, or a
+comet nucleus.
+
+NASA's Dawn spacecraft orbited Ceres from 2015 to 2018, spending much of its
+final year in LAMO: a ~375 km, near-circular polar orbit used for its
+highest-resolution gravity and neutron/gamma-ray mapping.
```
duncaneddy: We should make it's clear that this example is "inspired by" Dawn's LAMO orbit/phase, but not reference any parameters as exact to give leeway in design.

docs/examples/dawn_ceres_orbit.md
```diff
@@ -0,0 +1,143 @@
...
+conversion that measures inclination against the z-axis of whatever basis its
+output is read in, so interpreting its output directly as the propagation
+state would reference the 90 degree inclination to the ICRF pole, not Ceres's
+equator. To place the orbit correctly, we build a Ceres-equatorial inertial
+basis - z-axis on the Ceres spin pole (straight from the IAU pole constants),
+x-axis on the ascending node of the Ceres equator on the ICRF equator -
+construct the state in that basis, and rotate its position and velocity into
+the propagation frame. `state_koe_to_eci_for_body` converts the Keplerian
+elements using an arbitrary body's gravitational parameter, here `GM_CERES`,
+since `state_koe_to_eci` assumes Earth:
```
duncaneddy: Same issue as others for the koe_to_eci definitions.

docs/examples/dawn_ceres_orbit.md
```diff
@@ -0,0 +1,143 @@
...
+``` python
+--8<-- "./examples/examples/dawn_ceres_orbit.py:propagation"
+```
+
+!!! note "state_bci vs. state_eci"
+    `state_bci` returns the propagator's native state in the central body's
+    body-centered inertial frame - here, Ceres-centered. `state_eci` would
+    instead try to re-center the state onto Earth via SPK ephemeris data,
+    which raises for Ceres: NAIF ID 2000001 has no SPK coverage in the DE
+    kernels brahe loads. Use `state_bci` for any non-Earth propagator.
```
duncaneddy: This is now the third tip on state_bci, should be dropped given previous references.

examples/examples/dawn_ceres_orbit.py
```diff
@@ -0,0 +1,239 @@
...
+    fixed_frame=ceres_fixed,
+)
+
+# Gravity-only, point-mass force model: no third bodies or SRP, since the DE
+# kernels carry no Ceres ephemeris and the example stays entirely
+# Ceres-centered. ICGEM's only cataloged Ceres model (sphericalRFM_CERES_2519)
+# is a degree-2519 crustal forward-modeling research product, not a
+# conventional field normalized to the body's true GM (its C(0,0) ~ 0.126,
+# not the usual 1.0), so it isn't a drop-in gravity field here; point mass is
+# the standard starting point when defining your own body from scratch.
```
duncaneddy: Can drop the extensive comments in the body of the example and just leave short, informative ones explaining the code.

This one for example is duplicative with the example .md documentation.

examples/examples/dawn_ceres_orbit.py
```diff
@@ -0,0 +1,239 @@
...
+duration = 2 * 86400.0
+print(f"\nPropagating {duration / 86400.0:.0f} days...")
+prop.propagate_to(epoch + duration)
+print("  Complete!")
+
+# state_bci returns the propagator's native state in the central body's
+# body-centered inertial frame (here, Ceres-centered). state_eci would
+# instead try to re-center the state onto Earth via SPK ephemeris data, which
+# raises for Ceres since no SPK kernel covers NAIF ID 2000001.
+dt = 120.0
```
duncaneddy, line 160: This statement is not correct. state_eci is not relevant here so can be dropped from consideration and the discussion. It is also not true that no SPK kernel covers ceres. It just needs to be constructed/downloaded through the horizons's system.

docs/examples/dawn_ceres_orbit.md
duncaneddy: Removed in e1521073 — the tip duplicated the LRO page's, which now carries the linked explanation.

examples/examples/dawn_ceres_orbit.py
duncaneddy: Compressed in e1521073 — the force-model comment now states the choice in four lines and points to the docs page for the Horizons/ICGEM discussion.

src/frames/synodic.rs
```diff
@@ -147,6 +151,98 @@ pub(crate) fn state_synodic_to_inertial(
...
+/// Returns [`BraheError::Error`] for IDs without a packaged GM constant.
+pub(crate) fn body_gm(naif_id: i32) -> Result<f64, BraheError> {
+    match naif_id {
+        10 => Ok(GM_SUN),
+        1 | 199 => Ok(GM_MERCURY),
+        2 | 299 => Ok(GM_VENUS),
+        399 => Ok(GM_EARTH),
+        301 => Ok(GM_MOON),
+        3 => Ok(GM_EARTH + GM_MOON),
+        4 | 499 => Ok(GM_MARS),
```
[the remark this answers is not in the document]
duncaneddy: Fixed in 6ba1addc: split the combined `4 | 499 => GM_MARS` arm in `body_gm` — NAIF ID 4 (Mars-system barycenter) now returns `GM_MARS_SYSTEM`, 499 (Mars itself) keeps `GM_MARS`. Added direct test coverage for all `body_gm` match arms, including this one.

docs/learn/frames/synodic_frames.md
duncaneddy: Replaced with a tested code block in 216f6b4e: added `examples/frames/generic_synodic_frame.py` and `.rs`, included via the tabbed `--8<--` pattern and covered by `just test-example`.

docs/learn/plots/synodic_plots.md
duncaneddy: Added in 216f6b4e: `plots/learn/plots/synodic_plot_plotly.py` generates the figure via `just make-plot`, with the script embedded in the page as a Plot Source block.

examples/examples/dawn_ceres_orbit.py
duncaneddy: Reworded in 1ce7056f: the force-model comment and the page's Force Model section now note that a Ceres SPK can be generated via JPL Horizons (https://ssd-api.jpl.nasa.gov/doc/horizons.html) to enable third-body perturbations, and that a Horizons client is planned (issue #402). The example keeps gravity-only for self-containment in the meantime.

examples/examples/dawn_ceres_orbit.py
duncaneddy: Added standalone `rotation_x`/`rotation_y`/`rotation_z` functions in 4ffa3d08 (Rust + Python, returning matrices; delegate to the existing `RotationMatrix::Rx/Ry/Rz` constructors, which are kept). The inline equatorial-basis construction in this example (which built Rx/Rz by hand) was then deleted in 20dc3f24 in favor of calling `state_koe_to_inertial_for_body` directly.

examples/examples/dawn_ceres_orbit.py
duncaneddy: Cited in 1ce7056f: Archinal et al., 2018 (IAU WGCCRE report) for the ascending-node orientation convention, added to the Dawn/Ceres page, the `equatorial_basis_from_pole` rustdoc in src/coordinates/cartesian.rs, and the Python docstrings for `state_koe_to_inertial_for_body`/`state_inertial_to_koe_for_body`.

examples/examples/dawn_ceres_orbit.py
duncaneddy: Explained in 1ce7056f, added to the Dawn/Ceres page alongside the Archinal et al. 2018 citation: z_ICRF x pole is perpendicular to both poles, hence lies in both equatorial planes — the line of nodes.

examples/examples/dawn_ceres_orbit.py
duncaneddy: Reworded in 1ce7056f: "no dedicated support" -> "no built-in constants", in both the example docstring and the page, matching the existing-mechanism framing.

brahe/plots/bodies.py
duncaneddy: Added in ec2cc847: `R_MERCURY`, `R_VENUS`, `R_JUPITER`, `R_SATURN`, `R_URANUS`, `R_NEPTUNE` as packaged constants in src/constants/physical.rs (WGCCRE 2015-cited, same pattern as the existing `R_MARS`), exposed through the crate root and Python module. `brahe/plots/bodies.py` now uses them; Ceres keeps its literal radius deliberately since it has no packaged constant.

src/frames/transform.rs
duncaneddy: Removed in 28255275: `BodyFixedCustom` no longer rejects self-assigned centers at or below -1_000_000_000. Colliding with an actual synodic barycenter ID requires two packaged-GM NAIF IDs in 0..=999, which is astronomically unlikely; the doc comment now explains why the range isn't enforced rather than claiming it's rejected.

src/propagators/force_model_config.rs
duncaneddy: Reworded in 28255275: "...must model central_body itself" -> "must be a gravity field for the declared central_body; the model's source body must match", still pointing at `validate_gravity_model_body` for how that match is determined.

docs/examples/lro_lunar_orbit.md
duncaneddy: Reworded in 1ce7056f to "LRO-like" framing throughout, with an explicit note that the example uses an LRO-like ~30x180 km orbit rather than reproducing LRO's exact orbital history. Also fixed a pre-existing mismatch where the docstring said "~200 km class" while the actual orbit code builds a 30x180 km orbit — corrected to 180 km.

docs/examples/lro_lunar_orbit.md
duncaneddy: Linked in 1ce7056f: mascons now links to https://en.wikipedia.org/wiki/Mass_concentration_%28astronomy%29.

docs/examples/lro_lunar_orbit.md
duncaneddy: Reworded in 1ce7056f: dropped the blanket design-practice statement; the paragraph now ends "these mascon-driven gravity anomalies are strong enough that most low lunar orbits are unstable on timescales of weeks to months without station-keeping. Proper modeling of the high-order gravity dynamics enable design of station-keeping maneuvers that minimize the growth of these instabilities, conserving fuel and extending mission-lifetime."

docs/examples/lro_lunar_orbit.md
duncaneddy: Fixed in 20dc3f24: `state_koe_to_eci_for_body` -> `state_koe_to_inertial_for_body` (and `state_eci_to_koe_for_body` -> `state_inertial_to_koe_for_body`), taking a `CentralBody` instead of a bare `gm`. Elements are now referenced to the body's mean equator at J2000 (IAU pole, with the ascending node on the ICRF equator as x-axis) instead of the ICRF pole, so the functions compose with the body-fixed transforms. The LRO/MRO/Dawn examples were updated to call the renamed function directly instead of building the equatorial basis by hand.

docs/examples/lro_lunar_orbit.md
duncaneddy: Cross-linked in 1ce7056f: the state_bci/state_eci tip on this page now links `state_bci` to the cislunar propagation learn page and `state_eci` to the numerical propagator learn page's state-at-arbitrary-epochs section. The MRO page's duplicate of this tip was dropped in favor of this one (see the MRO thread).

docs/examples/lro_lunar_orbit.md
duncaneddy: Clarified in 1ce7056f: the plot title/labels now make explicit this is full-gravity-minus-point-mass, not full-force-minus-truth; figure regenerated.

docs/examples/earth_moon_free_return.md
duncaneddy: Fixed in fc6881be: linked https://en.wikipedia.org/wiki/Free-return_trajectory and corrected the Artemis claim — "Artemis I did not fly a strict free return; Artemis II, the first crewed Artemis flight, does."

docs/examples/earth_moon_free_return.md
duncaneddy: Raised in db63549b: parking orbit is now 400 km (ISS-like), with the aim offset retuned to +10 deg; TLI dv = 3.0943 km/s.

docs/examples/earth_moon_free_return.md
duncaneddy: Reworked in db63549b: propagation is now centered on the Earth-Moon barycenter (EMBI) with a 5x5 Earth spherical-harmonic field as an attributed third body plus Moon/Sun point-mass third bodies, replacing the earlier approach. The Force Model section was rewritten in fc6881be to match.

docs/examples/earth_moon_free_return.md
duncaneddy: Checked in fc6881be: the existing global `.plotly-embed iframe` rule already renders embeds full-width/centered, so no CSS change was needed. What was actually off-center was the 3D camera framing, which was reframed (elev 50, az -120, dist 2.4) so the figure-8 sits centered in view.

docs/examples/earth_moon_free_return.md
duncaneddy: Reworded in fc6881be: dropped the extended justification. The section now states plainly that this example takes shortcuts (fixed departure geometry, 1D bisection on TLI delta-v) "which in practice should not be done. Real mission design uses Lambert solvers and B-plane targeting."

docs/examples/earth_moon_free_return.md
duncaneddy: Checked and fixed in db63549b. The original trajectory (185 km parking, +16 deg aim) was actually already a retrograde far-side pass (h_selenocentric . h_moon_orbit < 0, ~200 deg selenocentric wrap, perilune beyond the Moon) — it only looked like it passed in front of the Moon because its ~12,000 km perilune was too distant to read as a circumlunar loop at plot scale, rendering as a leading-side teardrop. The fix was not to flip the aim offset (that produces a prograde near-side pass — the wrong geometry) but to tighten the perilune to ~2,000 km altitude on the same retrograde branch (400 km departure, +10 deg aim). At that perilune the top-down EMR plot added in fc6881be shows an unambiguous figure-8 with the outbound and return legs crossing near Earth. This is now verified with an explicit `h_sel . h_moon < 0` assert in the validation block, plus the new 2D top-down plot with direction arrows.

docs/examples/earth_moon_free_return.md
duncaneddy: Removed in fc6881be — dropped the hardcoded table rather than generating it as a CSV.

docs/examples/mro_mars_orbit.md
duncaneddy: Cited in 1ce7056f: Zurek & Smrekar (2007), https://doi.org/10.1029/2006JE002701, for MRO's sun-synchronous science-orbit characterization.

docs/examples/mro_mars_orbit.md
duncaneddy: Derived in 1ce7056f: added the J2 nodal-precession equation and solved it for inclination at this example's a/e against Mars's heliocentric mean motion (687-day year) with J2 = 1.96045e-3, giving i ~ 92.6 deg. Verified independently with the codebase's own GM_MARS/R_MARS constants, matching the example's design value to 1 decimal.

docs/examples/mro_mars_orbit.md
duncaneddy: Same fix as the LRO page: `state_koe_to_eci_for_body` -> `state_koe_to_inertial_for_body` in 20dc3f24, elements now referenced to Mars's mean equator at J2000 via the IAU pole rather than the ICRF pole.

docs/examples/mro_mars_orbit.md
duncaneddy: Fixed in 796494df: added `vertical_spacing` and dropped the top-row tick labels in `_keplerian_elements_plotly` (brahe/plots/trajectories.py); figure regenerated.

docs/examples/mro_mars_orbit.md
duncaneddy: Dropped in 1ce7056f: removed the MRO page's state_bci/state_eci tip box in favor of the LRO page's (now cross-linked to the relevant learn pages).

docs/examples/dawn_ceres_orbit.md
duncaneddy: Reworded in 1ce7056f: "fly the Dawn spacecraft's...LAMO" -> "fly an orbit inspired by the Dawn spacecraft's...LAMO", with an added closing line: "This example is inspired by that LAMO phase rather than reproducing its exact mission parameters."

docs/examples/dawn_ceres_orbit.md
duncaneddy: Same fix as the LRO/MRO pages: `state_koe_to_eci_for_body` -> `state_koe_to_inertial_for_body` in 20dc3f24. The page's Propagation section now describes the renamed function and its J2000 mean-equator reference instead of the manual basis construction.

examples/examples/dawn_ceres_orbit.py
duncaneddy: Fixed in 20dc3f24: the comment now correctly says `state_bci` returns the propagator's native Ceres-centered state and `state_eci` would try to re-center onto Earth via SPK, which raises since NAIF ID 2000001 (Ceres) has no SPK coverage in the DE kernels brahe loads — dropped the earlier incorrect claim that no SPK kernel covers Ceres at all (one can be generated via Horizons, per the Force Model section discussion).

src/coordinates/cartesian.rs
```diff
@@ -46,13 +52,15 @@ use crate::orbits;
 /// 1. O. Montenbruck, and E. Gill, *Satellite Orbits: Models, Methods and Applications*, pp. 24, eq. 2.43 & 2.44, 2012.
 #[allow(non_snake_case)]
 pub fn state_koe_to_eci(x_oe: SVector6, angle_format: AngleFormat) -> SVector6 {
-    state_koe_to_eci_for_body(x_oe, GM_EARTH, angle_format)
+    state_koe_to_inertial_gm(x_oe, GM_EARTH, angle_format)
```
duncaneddy: I don't love delegating the Earth-implementation to state_koe_to_inertial_gm. It isn't pedagogically illustrative. I'd prefer to have duplicative koe_to_eci and eci_to_koe implementations and use tests to confirm the correctness of the general-body implementation (using Earth as the example).

We should restore this back to the original implementation at the last tagged release.

src/frames/transform.rs
```diff
@@ -84,8 +85,67 @@ use super::mars::rotation_mci_to_mcmf;
...
+/// - `id`: Synthetic negative center ID
+///
+/// # Examples
+/// ```
+/// use brahe::frames::{synodic_barycenter_id, SUN_EARTH_BARYCENTER_ID};
+///
+/// assert_eq!(synodic_barycenter_id(10, 399), SUN_EARTH_BARYCENTER_ID);
+/// ```
+pub fn synodic_barycenter_id(primary: i32, secondary: i32) -> i32 {
+    debug_assert!((0..=999).contains(&primary) && (0..=999).contains(&secondary));
```
duncaneddy: We can drop this debug assert.

src/frames/transform.rs
```diff
@@ -84,8 +85,67 @@ use super::mars::rotation_mci_to_mcmf;
...
+/// [`SUN_EARTH_BARYCENTER_ID`] is this encoding evaluated at `(10, 399)`.
+///
+/// The full range at or below -1_000_000_000 is reserved for these synthetic
+/// barycenters; see [`ReferenceFrame::BodyFixedCustom`] for why a
+/// user-registered custom frame is nonetheless allowed to self-assign a
+/// center in that range.
+///
+/// # Arguments
+/// - `primary`: NAIF ID of the primary body. Must be in `0..=999`.
+/// - `secondary`: NAIF ID of the secondary body. Must be in `0..=999`.
```
duncaneddy: Comment should be updated to remove the "Must be in `0..=999`." since that is no longer a restriction (we should be able to support synodic computations for any provided NAIF ID and just explain the collision condition

src/propagators/force_model_config.rs
```diff
@@ -1064,6 +1068,165 @@ impl ForceModelConfig {
...
+    ///   still carries the real body GM (as with the ICGEM Ceres model
+    ///   `sphericalRFM_CERES_2519`, whose C̄00 ≈ 0.126 rather than 1.0) passes.
+    ///   `central_body.gm() <= 0.0` (possible for a `CentralBody::Custom` with an
+    ///   unset/zero `gm` — built-in barycenters are already rejected earlier in
+    ///   `validate()`) always errors here rather than silently skipping the GM
+    ///   comparison, since no real gravity model has a non-positive GM.
+    ///
+    /// # Returns
+    /// `Ok(())` if the gravity model's body matches `central_body` (or the check
+    /// does not apply), `Err(BraheError)` naming both bodies otherwise.
```
duncaneddy: Should be more brief and up front. Reduce the unnecessary information and make less wordy with unrelevant information. Just describe the checks it performs, configurations that will fail, and any other pitfalls relevant to the user. Bottom-line-up-front

src/constants/physical.rs
```diff
@@ -441,6 +441,84 @@ pub const GM_PHOBOS: f64 = 7.087546066894452e5;
...
+/// # References:
+///
+///  1. Archinal et al., *Report of the IAU Working Group on Cartographic
+///     Coordinates and Rotational Elements: 2015*, Celest Mech Dyn Astr, 2018.
+///
+/// # Examples
+/// ```
+/// assert_eq!(brahe::constants::R_MERCURY, 2439.7e3);
+/// ```
+pub const R_MERCURY: f64 = 2439.7e3;
```
duncaneddy: Should be 2439.4e3

Image

docs/examples/lro_lunar_orbit.md
```diff
@@ -0,0 +1,138 @@
...
+basis construction. We also evaluate the lunar mean pole at J2000 (the third
+row of the ICRF-to-lunar-body-fixed rotation) to confirm the geometry below:
+
+``` python
+--8<-- "./examples/examples/lro_lunar_orbit.py:preamble"
+```
+
+``` python
+--8<-- "./examples/examples/lro_lunar_orbit.py:orbit_setup"
+```
```
duncaneddy, line 55: And one sentence to break up the preample and the actual problem setup. You can just say something like (we use the standard preample) then go into the setup.

examples/examples/lro_lunar_orbit.py
```diff
@@ -0,0 +1,210 @@
...
+moon_pole = np.asarray(bh.rotation_icrf_to_body_fixed_iau(301, j2000))[2, :]
+
+print(
+    f"Perilune radius: {r_p / 1e3:.1f} km (altitude: {(r_p - bh.R_MOON) / 1e3:.0f} km)"
+)
+print(
+    f"Apolune radius:  {r_a / 1e3:.1f} km (altitude: {(r_a - bh.R_MOON) / 1e3:.0f} km)"
+)
+print(f"Semi-major axis: {a / 1e3:.1f} km, eccentricity: {e:.4f}")
+print(f"Moon spin pole (ICRF): {np.array2string(moon_pole, precision=4)}")
```
duncaneddy: Remove this section it isn't helpful or illustrative. just makes it longer and muddies the core implementation points.

docs/examples/earth_moon_free_return.md
```diff
@@ -0,0 +1,125 @@
...
+
+*Body textures: [Solar System Scope](https://www.solarsystemscope.com/textures/), CC BY 4.0.*
+
+## Validation
+
+We confirm the trajectory is a genuine circumlunar free return on three counts: the perilune passes above the lunar surface but within 20,000 km of the Moon's center; the return leg descends below 1,000 km altitude, closing the loop back at Earth; and the flyby is a retrograde selenocentric pass. That last check pins the geometry - the spacecraft's angular momentum about the Moon points opposite the Moon's orbital angular momentum about Earth, the signature of a far-side circumlunar swing-by rather than a near-side pass in front of the Moon. Because the propagation stops at the 120 km entry interface, the minimum Earth distance measured is that entry crossing; the true geometric perigee lies below it, inside the atmosphere.
+
+``` python
+--8<-- "./examples/examples/earth_moon_free_return.py:validation"
+```
```
duncaneddy: Can drop this section from both docs and code.

docs/examples/earth_moon_free_return.md
```diff
@@ -0,0 +1,125 @@
...
+
+---
+
+## What "Free Return" Means
+
+A free-return trajectory is a solution of the Earth-Moon-spacecraft three-body problem whose outbound leg is aimed so that the lunar flyby rotates the velocity vector back toward Earth. No propulsion is needed after the initial injection: the Moon's gravity does the work of turning the spacecraft around. The defining property is passive safety - once on the trajectory, the spacecraft returns to Earth's vicinity on its own.
+
+The flyby must be a genuine **circumlunar** pass: the spacecraft crosses to the far side of the Moon and swings around it, so its motion about the Moon is retrograde with respect to the Moon's orbital motion about Earth. That far-side geometry is what bends the path steeply back toward home. A pass on the near side, in front of the Moon, deflects the trajectory the wrong way. The pass must also be close enough to bend the path back to Earth but not so close that the perilune drops below the surface. The window between "escapes" and "impacts the Moon" is what makes free-return design a targeting problem rather than a closed-form calculation.
```
duncaneddy: Remove paragraph.

docs/examples/earth_moon_free_return.md
```diff
@@ -0,0 +1,125 @@
...
+
+A real mission targets a two-dimensional B-plane at the Moon (a miss distance and an approach angle). Here `AIM_OFFSET_DEG` is a simplified stand-in for that second dimension: rotating the departure point ahead of the pure antipode about the orbit normal sets up a flyby that swings around the far side of the Moon rather than passing in front of it. We use `spk_state` to query the Moon's Earth-relative state from the DE440s ephemeris and build an orthonormal departure frame from it. Because the mission is integrated about the Earth-Moon barycenter, the departure state is translated from ECI into the EMBI frame with `state_eci_to_emb`.
+
+``` python
+--8<-- "./examples/examples/earth_moon_free_return.py:preamble"
+```
+
+``` python
+--8<-- "./examples/examples/earth_moon_free_return.py:geometry"
+```
```
duncaneddy, line 27: Separate preample and setup.

docs/examples/mro_mars_orbit.md
```diff
@@ -0,0 +1,147 @@
...
+equatorial basis vectors (the ascending node and its quadrature axis) evaluated
+at J2000, the same reference plane the elements are referenced to:
+
+``` python
+--8<-- "./examples/examples/mro_mars_orbit.py:preamble"
+```
+
+``` python
+--8<-- "./examples/examples/mro_mars_orbit.py:orbit_setup"
+```
```
duncaneddy, line 67: Separate preample and setup

examples/examples/mro_mars_orbit.py
```diff
@@ -0,0 +1,218 @@
...
+    raan_deg.append(koe[3])
+    argp_deg.append(koe[4])
+    anom_deg.append(koe[5])
+sma_km = np.array(sma_km)
+ecc = np.array(ecc)
+inc_deg = np.array(inc_deg)
+raan_deg = np.array(raan_deg)
+argp_deg = np.array(argp_deg)
+anom_deg = np.array(anom_deg)
+times_hours = np.arange(0.0, duration, dt) / 3600.0
```
duncaneddy: I believe this should be passed in directly as seconds to properly set the x-axis. Right now it shows as "seconds"

Image

examples/examples/dawn_ceres_orbit.py
```diff
@@ -0,0 +1,225 @@
...
+        @ bh.Rx(np.pi / 2 - delta, bh.AngleFormat.RADIANS)
+        @ bh.Rz(np.pi / 2 + alpha, bh.AngleFormat.RADIANS)
+    )
+
+
+def ceres_omega(epc=None):
+    return np.array([0.0, 0.0, np.radians(CERES_W_RATE) / 86400.0])
+
+
+bh.register_custom_frame(1, ceres_rotation, ceres_omega)
```
duncaneddy, line 99: Explain what "1" is here in a comment (arbitrarily picked custom frame key just needs to be unique)

docs/examples/dawn_ceres_orbit.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Propagation
+
+Dawn's LAMO is a ~375 km circular polar orbit. The propagator integrates in a
+Ceres-centered inertial frame whose axes are ICRF-aligned, so its z-axis is
+the ICRF pole - about 23 degrees from Ceres's spin pole. Passing 90 degrees
+straight to `state_koe_to_eci` would measure the inclination against the ICRF
+pole, not Ceres's equator. `state_koe_to_inertial_for_body` instead references
+the elements to the body's mean equator at J2000 and returns the state in the
+Ceres-centered inertial frame. Because `ceres` is a `CentralBody.Custom` that
```
duncaneddy: Can be simplified the "Passing 90 degrees straight to state_koe_eci" is a bit confusing. Can just say, we need to use `state_koe_to_inertial_for_body` to reference the elements to the body's mean equator at J2000 and then use them to compute the state in the Ceres-centered, ICRF-aligned inertial frame.

examples/examples/dawn_ceres_orbit.py
```diff
@@ -0,0 +1,225 @@
...
+
+oe = np.array(
+    [R_CERES + 375e3, 0.001, 90.0, 0.0, 0.0, 0.0]
+)  # [m, -, deg, deg, deg, deg]
+
+# state_koe_to_inertial_for_body references the elements to Ceres's mean equator
+# at J2000. Because `ceres` is a Custom body, it reads the pole from the
+# registered body-fixed frame (`ceres_fixed`) and returns the state in the
+# Ceres-centered ICRF-aligned frame the propagator integrates in, so the 90 deg
+# inclination is measured against Ceres's equator rather than the ICRF pole.
```
duncaneddy: Can drop extraneous comment. Already covered in example .md

examples/examples/dawn_ceres_orbit.py
```diff
@@ -0,0 +1,225 @@
...
+force_config = bh.ForceModelConfig.for_body(ceres, bh.GravityConfiguration.point_mass())
+# --8<-- [end:force_model]
+
+# --8<-- [start:orbit_setup]
+# Dawn LAMO (Low Altitude Mapping Orbit): ~375 km circular polar orbit.
+epoch = bh.Epoch.from_datetime(2016, 1, 1, 0, 0, 0.0, 0.0, bh.TimeSystem.UTC)
+
+oe = np.array(
+    [R_CERES + 375e3, 0.001, 90.0, 0.0, 0.0, 0.0]
+)  # [m, -, deg, deg, deg, deg]
```
duncaneddy: Change comment to cover element order (a [m], e [-],...)

src/coordinates/cartesian.rs
duncaneddy: Restored in 0642e41d — state_koe_to_eci/state_eci_to_koe carry their standalone v1.6.2 implementations again; the general-body functions are implemented separately, with tests confirming CentralBody.Earth through the generic path reproduces the Earth implementation.

src/frames/transform.rs
duncaneddy: Dropped in 0642e41d.

src/frames/transform.rs
duncaneddy: Done in 0642e41d — the 0..=999 wording is gone from the Rust docs and learn page, and the Python constructor now accepts any NAIF ID for any origin. Docs explain the actual behavior: the barycenter encoding is injective within 0..=999; larger IDs decode as ordinary centers and fail loudly at transform time via the SPK/GM lookup, so there is no silent collision.

src/propagators/force_model_config.rs
duncaneddy: Rewritten BLUF in 0642e41d — states the checks performed, the configurations that fail, and the body-identity-not-model-quality scope in one line; narrative history cut.

src/constants/physical.rs
duncaneddy: Fixed in 0642e41d — R_MERCURY = 2439.4e3 (mean radius per the fact sheet); doc example and tests updated.

docs/examples/lro_lunar_orbit.md
duncaneddy: Added in f675d390.

examples/examples/lro_lunar_orbit.py
duncaneddy: Removed in 1c15a6e1; snippet markers and the page include updated.

docs/examples/earth_moon_free_return.md
duncaneddy: Dropped from both the page and the script in 1c15a6e1.

docs/examples/earth_moon_free_return.md
duncaneddy: Removed in f675d390.

docs/examples/earth_moon_free_return.md
duncaneddy: Added in f675d390.

docs/examples/mro_mars_orbit.md
duncaneddy: Added in f675d390.

examples/examples/mro_mars_orbit.py
duncaneddy: Fixed in 1c15a6e1 — the times column is now passed as elapsed seconds and the array path reuses the same auto-unit axis formatting as trajectory input (renders as Time (hours) for the 2-day span). Figure regenerated and inspected.

examples/examples/dawn_ceres_orbit.py
duncaneddy: Added in 1c15a6e1 — the comment now notes the key is arbitrary and only needs to be unique among registered custom frames.

docs/examples/dawn_ceres_orbit.md
duncaneddy: Reworded in f675d390 using your sentence.

examples/examples/dawn_ceres_orbit.py
duncaneddy: Dropped in 1c15a6e1.

examples/examples/dawn_ceres_orbit.py
duncaneddy: Updated in 1c15a6e1 to the [a [m], e [-], i [deg], RAAN [deg], argp [deg], M [deg]] form.

src/coordinates/cartesian.rs
```diff
@@ -242,13 +319,230 @@ pub fn state_eci_to_koe_for_body(x_cart: SVector6, gm: f64, angle_format: AngleF
...
+///   body without a `fixed_frame`: `Err` — a positive GM and a pole are
+///   both required.
+fn body_equatorial_basis(central_body: &CentralBody) -> Result<SMatrix3, BraheError> {
+    // Earth: GCRF already is the Earth mean-equator inertial frame; an
+    // exact-identity basis guarantees zero behavior change for Earth callers.
+    if let CentralBody::Earth = central_body {
+        return Ok(SMatrix3::identity());
+    }
+
+    // Barycenters and any other massless body have neither a mean equator
```
duncaneddy, line 382: We should add learn/* documentation for these functions to docs/learn/coordinates/cartesian.rs

PR #406, opened 2026-07-18
Add SSA sensor simulation: SSN dataset, az/el/range measurements, and tracking example

duncaneddy:
# Pull Request

## Description

Adds end-to-end SSA (space surveillance) sensor simulation: a dataset of representative US Space Surveillance Network sensors from Vallado 4th Ed. (Tables 4-2/4-3/4-4 — site locations, field-of-view limits, and bias/noise calibration values), an azimuth/elevation/range measurement model, a `SimpleSSNSensor` type that simulates tracking measurements (step-wise for sequential-planning loops, or batched over a trajectory), and a measurement-free `propagate_to` on the sequential filters for advancing estimators across tracking gaps.

Angle handling is wrap-aware end to end: azimuth residuals wrap across 0°/360° via a new `MeasurementModel::residual()` hook routed through all three estimators, the model's finite-difference Jacobian differences through the wrapped residual, and the UKF computes its measurement statistics with a reference-point mean so sigma points straddling north remain well-behaved (algebraically identical to the standard weighted mean for non-angular models).

Two runnable examples demonstrate the workflow (a compact standalone Rust/Python pair, and a 6-hour walkthrough that visualizes the sensor network, simulates pass-windowed measurements at a configurable cadence, and compares BLS/EKF/UKF estimation with covariance/error plots annotated with pass windows), with documentation pages for the dataset, measurement model, and walkthrough. Rust patch coverage is 100% under `cargo llvm-cov`; Python bindings mirror the Rust API 1:1 with parity tests.

### Added
- `datasets::ssn_sensors::load_ssn_sensors` / `bh.datasets.ssn_sensors.load()`: embedded dataset of 21 representative SSN sensor sites from Vallado 4th Ed. Tables 4-2/4-3/4-4 with locations, az/el/range field-of-view limits, and bias/noise calibration values
- `AzElRangeMeasurementModel`: topocentric azimuth/elevation/range measurement model with constant-bias support, azimuth residual wrapping, and a wrap-aware finite-difference Jacobian
- `SimpleSSNSensor`: ground-sensor measurement simulation with wrap-aware field-of-view checks, seeded Gaussian noise, step-wise (`measure`) and batched (`simulate_observations`) generation, and a matching filter-side `measurement_model()`
- `MeasurementModel::residual()`: overridable, dimension-validated residual hook used by all EKF/UKF/BLS residual computations
- `ExtendedKalmanFilter::propagate_to` / `UnscentedKalmanFilter::propagate_to`: measurement-free prediction step that records covariance growth across tracking gaps
- SSN tracking examples: standalone Rust/Python pair (`examples/estimation/ssn_tracking`) and a six-hour BLS/EKF/UKF walkthrough with sensor-network, measurement, and filter-comparison figures (`docs/examples/ssn_tracking.md`)

### Changed
- UKF measurement statistics use a reference-point mean and residual-based deviations, making the measurement update robust for angular measurements near a wrap (identical results for non-angular models)
- EKF/UKF/BLS pre-fit and post-fit residuals now route through `MeasurementModel::residual()` (default behavior unchanged)
- `rand`/`rand_distr` promoted from dev-dependencies to runtime dependencies for measurement noise sampling


crates/brahe-py/src/estimation.rs
```diff
@@ -1015,6 +1052,517 @@ impl_measurement_model_binding!(
     doc: "ECEF state measurement model (6D ECEF state from GNSS).\n\nInternally converts ECI state to ECEF state."
 );
 
+/// Validate station geodetic coordinates before constructing an
+/// `AzElRangeMeasurementModel`, so an out-of-range latitude/longitude raises a
+/// clear `ValueError` instead of tripping the infallible Rust constructor's
+/// `.expect("Invalid geodetic coordinates")` panic. Bounds scale with the
+/// angle format (degrees vs radians). NaN inputs fail the `<=` comparison and
```
duncaneddy: This seems like the Rust constructor should return a Result and this should convert that BraheError into a python exception instead of implementing custom error-handling

docs/examples/ssn_tracking.md
```diff
@@ -0,0 +1,93 @@
...
+  <iframe class="only-dark"  src="../figures/ssn_tracking_filters_dark.html"  loading="lazy"></iframe>
+</div>
+
+The EKF and UKF converge to nearly identical results here (6.6 m final position error for both) since the two-body dynamics and az/el/range geometry are only mildly nonlinear over this arc; the UKF's sigma-point propagation does not offer a material advantage at this level of nonlinearity. Azimuth wrapping is handled consistently: `AzElRangeMeasurementModel.residual()` wraps the residual, the model's Jacobian differences through it, and the UKF forms its predicted measurement mean and innovation deviations through `residual()` as well, so a pass whose sigma-point azimuths straddle the 0/360&deg; wrap stays well-behaved. BLS, solving the full arc in one batch rather than propagating through gaps, reaches a 9.5 m final position error -- higher than the sequential filters here because it minimizes the weighted residual cost over the whole arc rather than tracking the most recent state, but with a comparable formal 3-sigma uncertainty (39.5 m vs. 42.1 m for the EKF/UKF).
```
duncaneddy: Shouldn't quote numbers directly in text since those might change between simulated runs

docs/learn/datasets/ssn_sensors.md
```diff
@@ -0,0 +1,127 @@
...
+```python
+import brahe as bh
+
+sites = bh.datasets.ssn_sensors.load()
+radars = [s for s in sites if s.properties["sensor_type"] == "azel_range"]
+sensors = bh.SimpleSSNSensor.from_locations(sites, seed=42)
+
+sensor = sensors[0]
+model = sensor.measurement_model()  # AzElRangeMeasurementModel matching this sensor
+```
```
duncaneddy: Code example should be implement in code (with python and trust examples)

examples/estimation/ssn_tracking.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+# Hermite-cubic interpolation is required so the trajectory (stored at the
+# propagator's ~60 s adaptive-step cadence) can be sampled accurately at the
+# much finer measurement cadence below; linear interpolation of a curved
+# orbit between 60 s-spaced points is off by kilometers at intermediate
+# epochs.
+truth_config = bh.NumericalPropagationConfig.default().with_interpolation_method(
+    bh.InterpolationMethod.HERMITE_CUBIC
+)
```
duncaneddy: I don't think we need to set HERMITE_CUBIC explicitly here. That should be the default interpolation method for all propagators (if that was not fixed and I'm hallucinating it we should fix that default).

Also we can shorten up the comment to:

```python
# Hermite-cubic interpolation is required so the trajectory (stored at the
# propagator's ~60 s adaptive-step cadence) can be sampled accurately at the
# much finer measurement cadence used for measurement simualtion.
```

src/estimation/ekf.rs
```diff
@@ -463,6 +465,127 @@ impl ExtendedKalmanFilter {
...
+    pub fn propagate_to(&mut self, epoch: Epoch) -> Result<FilterRecord, BraheError> {
+        let current_epoch = self.dynamics.current_epoch();
+        let dt: f64 = epoch - current_epoch;
+        if dt < -1e-9 {
+            return Err(BraheError::Error(format!(
+                "Target epoch is before current filter epoch. \
+                 Filter epoch: {}, target epoch: {}.",
+                current_epoch, epoch
+            )));
+        }
```
duncaneddy: This check needs to be direction aware since it would prevent backwards smoothing (estimation). Instead it should implement direction aware step-checks like done in DNumericalOrbitPropagation `step`

src/estimation/ekf.rs
```diff
@@ -463,6 +465,127 @@ impl ExtendedKalmanFilter {
...
+
+    /// Fallible body of [`propagate_to`](Self::propagate_to). Mutates the
+    /// propagator; the caller rolls it back if this returns an error.
+    ///
+    /// Keep in sync with the predict half of `predict_and_update`.
+    fn predict_only(
+        &mut self,
+        epoch: Epoch,
+        current_epoch: Epoch,
+        dt: f64,
```
duncaneddy: Call signature should be (current_epoch, target_epoch, dt)

Also dt doesn't need to be an argument. It can be computed from epoch and current_epoch

src/estimation/ekf.rs
```diff
@@ -463,6 +465,127 @@ impl ExtendedKalmanFilter {
...
+        if dt > 0.0
+            && let Some(ref pn) = self.config.process_noise
+        {
+            if pn.scale_with_dt {
+                p_predicted += &pn.q_matrix * dt;
+            } else {
+                p_predicted += &pn.q_matrix;
+            }
+        }
+        let p_predicted = 0.5 * (&p_predicted + p_predicted.transpose());
```
duncaneddy: This seems like it would not be adding process noise correctly. It's only doing it in one significant step instead of adding it throughout the integration process to all the internal steps that `propagate_to` takes, and therefore accumulating the noise in the subsequent propagation.

src/estimation/mod.rs
```diff
@@ -51,10 +55,11 @@ pub use config::{
 pub use dynamics_source::DynamicsSource;
 pub use ekf::ExtendedKalmanFilter;
 pub use measurement::{
-    EcefPositionMeasurementModel, EcefStateMeasurementModel, EcefVelocityMeasurementModel,
-    InertialPositionMeasurementModel, InertialStateMeasurementModel,
+    AzElRangeMeasurementModel, EcefPositionMeasurementModel, EcefStateMeasurementModel,
+    EcefVelocityMeasurementModel, InertialPositionMeasurementModel, InertialStateMeasurementModel,
```
duncaneddy: All `Ecef*MeasurementModel` should be `ECEF*MeasurementModel`

(Properly capitalized to be consistent with use of the acronym and properly allowed by rust/clippy)

src/estimation/sensors.rs
```diff
@@ -0,0 +1,796 @@
...
+    /// length and raise `ValueError` instead.
+    pub fn azelrange(&self, epoch: &Epoch, state_eci: &DVector<f64>) -> Vector3<f64> {
+        assert!(
+            state_eci.len() >= 3,
+            "SimpleSSNSensor requires a state vector with at least 3 elements, got {}",
+            state_eci.len()
+        );
+        let pos_eci = Vector3::new(state_eci[0], state_eci[1], state_eci[2]);
+        let pos_ecef = position_eci_to_ecef(*epoch, pos_eci);
+        let sez = relative_position_ecef_to_sez(
```
duncaneddy: Can we use ENU instead of SEZ?

src/estimation/sensors.rs
```diff
@@ -0,0 +1,796 @@
...
+        let azel = self.azelrange(epoch, state_eci);
+        self.in_fov(&azel)
+    }
+
+    fn in_fov(&self, azel: &Vector3<f64>) -> bool {
+        azel[1] >= self.el_min
+            && azel[1] <= self.el_max
+            && self.az_window_contains(azel[0])
+            && self.range_max.is_none_or(|rmax| azel[2] <= rmax)
+    }
```
duncaneddy: Instead of implementing these as custom, simple constraints, can we leverage the existing constraint checking machinery of AccessComputation to define and compute these constraints? 

That both reduces duplication and improves reliance on common machinery. Potentially also enabling extension by users adding other access constraints (such as right looking passes or otherwise)

src/estimation/ukf.rs
```diff
@@ -600,6 +627,128 @@ impl UnscentedKalmanFilter {
...
+    /// Fallible body of [`propagate_to`](Self::propagate_to). Mutates the
+    /// propagator; the caller rolls it back if this returns an error.
+    ///
+    /// Keep in sync with the predict half of `predict_and_update`.
+    fn predict_only(
+        &mut self,
+        epoch: Epoch,
+        current_epoch: Epoch,
+        dt: f64,
+    ) -> Result<FilterRecord, BraheError> {
```
duncaneddy, line 772: Same comments as for EKF apply here

src/estimation/ukf.rs
```diff
@@ -600,6 +627,128 @@ impl UnscentedKalmanFilter {
...
+        if dt > 0.0
+            && let Some(ref pn) = self.config.process_noise
+        {
+            if pn.scale_with_dt {
+                p_predicted += &pn.q_matrix * dt;
+            } else {
+                p_predicted += &pn.q_matrix;
+            }
+        }
+        let p_predicted = 0.5 * (&p_predicted + p_predicted.transpose());
```
duncaneddy: Same question here on the proper application of process noise to long-term propagation

docs/examples/ssn_tracking.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+<div class="plotly-embed">
+  <iframe class="only-light" src="../figures/ssn_tracking_filters_light.html" loading="lazy"></iframe>
+  <iframe class="only-dark"  src="../figures/ssn_tracking_filters_dark.html"  loading="lazy"></iframe>
+</div>
+
+The EKF and UKF converge to nearly identical results here (6.6 m final position error for both) since the two-body dynamics and az/el/range geometry are only mildly nonlinear over this arc; the UKF's sigma-point propagation does not offer a material advantage at this level of nonlinearity. Azimuth wrapping is handled consistently: `AzElRangeMeasurementModel.residual()` wraps the residual, the model's Jacobian differences through it, and the UKF forms its predicted measurement mean and innovation deviations through `residual()` as well, so a pass whose sigma-point azimuths straddle the 0/360&deg; wrap stays well-behaved.
+
+## Batch Least Squares (CI-Gated Companion)
```
duncaneddy: Remove "(CI-Gated Companion)" from the BLS title. That is an implementation detail the users don't need to know.

docs/examples/ssn_tracking.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Visualize the Sensor Network
+
+Next we propagate a truth trajectory -- a 700 km, 72&deg; inclination LEO orbit -- and plot its ground track against the sensor network's coverage cones, using each sensor's minimum elevation angle:
+
+``` python
+--8<-- "./examples/examples/ssn_tracking.py:visualize_network"
+```
+
+The truth trajectory is generated with Hermite-cubic interpolation. The propagator stores states at its adaptive-step cadence (roughly 60 s); linear interpolation between those points would be off by kilometers when sampled at the much finer measurement cadence used below, so Hermite-cubic interpolation is required to keep the simulated measurements consistent with the true orbit.
```
duncaneddy: This comment isn't helpful as is since brahe uses Hermite-cubic interpolation by default. Maybe has it as a tip. Like:

Normally sampling measurements at a finer time step (15 seconds) than the propagation step (60 seconds) would introduce errors due to the interpolated position diverging from the ground-truth trajectory. However brahe uses Hermite-Cubic interpolation by default which helps negate these artifacts. See [link to interpolation options and function to set it] for more information.

docs/examples/ssn_tracking.md
```diff
@@ -0,0 +1,116 @@
...
+--8<-- "./examples/examples/ssn_tracking.py:load_sensors"
+```
+
+## Visualize the Sensor Network
+
+Next we propagate a truth trajectory -- a 700 km, 72&deg; inclination LEO orbit -- and plot its ground track against the sensor network's coverage cones, using each sensor's minimum elevation angle:
+
+``` python
+--8<-- "./examples/examples/ssn_tracking.py:visualize_network"
+```
```
duncaneddy, line 29: It would be useful if the colored sensor comm cones matched the subsequent measurements in "Simulate Measurements"

docs/examples/ssn_tracking.md
```diff
@@ -0,0 +1,116 @@
...
+
+First, we'll import the necessary libraries, initialize Earth orientation parameters, and configure the tracking arc:
+
+``` python
+--8<-- "./examples/examples/ssn_tracking.py:preamble"
+```
+
+## Load the Sensor Network
+
+We load the Vallado SSN sensor sites and build az/el/range sensors from them. Of the 21 sites in the dataset, 13 construct as `SimpleSSNSensor` instances -- the remaining 8 are optical (`radec`) trackers or sites (HAX, Haystack) that Table 4-2 lists without the az/el/range calibration values `SimpleSSNSensor` requires:
```
duncaneddy: This is not correct, Haystack is a radar site, clearly stated in the documentation (`radec`) isn't a reason they don't work with SimpleSSNSensor. It's just they don't have calibration data.

We should have the option to initialize SimpleSSNSensor with the entire site list or just those that have all data defined. This example should just use the well-defined sensors that have all data. But by default the sensors that don't have data defined should load with `None` and we should be able to override the values to use them as proper sensors (e.g. default to zero noise/bias but add a .with_range_bias() or .with_sensor()` constructor parameters.

docs/examples/ssn_tracking.md
```diff
@@ -0,0 +1,116 @@
...
+
+The script reproduces the same scenario as `ssn_tracking.py` -- the sensor dataset, the truth orbit with its Hermite-cubic interpolation, and the simulated measurements -- before solving with BLS. Its default state-correction convergence threshold (`1e-8`, in mixed position/velocity units) is far tighter than the centimeter-level precision this measurement set actually supports, so the script loosens `state_correction_threshold` to `1e-3` and adds a `cost_convergence_threshold`; this stops the solver once corrections are physically insignificant instead of exhausting `max_iterations` on progressively smaller, noise-driven corrections:
```
duncaneddy: All explanation of the constraints of Ci and that the example isn't actually run should be removed. That isn't helpful to the reader here. You can just say it's slower and that's sufficient. This is user docs.

Just talk about the trade with BLS and the fact that it minimizes the error over the entire arc.

> Its default state-correction convergence threshold (`1e-8`, in mixed position/velocity units) is far tighter than the centimeter-level precision this measurement set actually supports, so the script loosens `state_correction_threshold` to `1e-3` and adds a `cost_convergence_threshold`; this stops the solver once corrections are physically insignificant instead of exhausting `max_iterations` on progressively smaller, noise-driven corrections

is helpful discussion about actually _using_ a BLS estimator

docs/examples/ssn_tracking.md
```diff
@@ -0,0 +1,116 @@
...
+Finally, we compare true position error against each filter's formal 3-sigma uncertainty, and check that the uncertainty envelope actually bounds the error (filter consistency):
+
+``` python
+--8<-- "./examples/examples/ssn_tracking.py:analyze_results"
+```
+
+<div class="plotly-embed">
+  <iframe class="only-light" src="../figures/ssn_tracking_filters_light.html" loading="lazy"></iframe>
+  <iframe class="only-dark"  src="../figures/ssn_tracking_filters_dark.html"  loading="lazy"></iframe>
+</div>
```
duncaneddy, line 75: Ok we can drop the "Formal 3-Sigma position undertainty"

Just keep the plot of the true position error. It would be better to first show plots of the uncertainty and error in radial/along-track/cross-track directions, then the overall true position error.

For the uncertainty plots just shade in the covariance region of +/- and plot the line of the error to make it more clear like this:

Image

Image


Image

src/datasets/ssn_sensors.rs
```diff
@@ -0,0 +1,129 @@
...
+}
+
+#[cfg(test)]
+#[cfg_attr(coverage_nightly, coverage(off))]
+mod tests {
+    use super::*;
+    use crate::access::location::AccessibleLocation;
+    use crate::utils::identifiable::Identifiable;
+
+    #[test]
```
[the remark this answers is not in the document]
duncaneddy: Fixed in e9dcf3c8 — every test added on this branch now carries an explicit `#[serial]` or `#[parallel]` label (EOP/global-state tests serial, pure tests parallel).

crates/brahe-py/src/estimation.rs
duncaneddy: Fixed in e9dcf3c8 — `AzElRangeMeasurementModel::new` now returns `Result<Self, BraheError>` (propagating the geodetic conversion error, with explicit non-finite checks), and the binding's custom `validate_station_coords` helper is deleted; the Python constructors are now pure conversions that map the `BraheError` to `ValueError`.

docs/examples/ssn_tracking.md
duncaneddy: Fixed in 7e2c71a9 — run-specific numbers are removed from the prose on both pages; quantitative results now come only from the captured-output includes.

docs/learn/datasets/ssn_sensors.md
duncaneddy: Fixed in 4afad098 — the learn page's inline snippets are now real standalone example pairs (`examples/datasets/ssn_sensors_load.{py,rs}` and `ssn_sensors_sensor.{py,rs}`), included via snippets with captured outputs, mirroring the groundstations examples.

examples/estimation/ssn_tracking.py
duncaneddy: You weren't hallucinating — the default was Linear. Fixed in 13f92ff0 + f3346398: `NumericalPropagationConfig.interpolation_method` is now `Option<InterpolationMethod>` where `None` (the default) resolves per-propagator — orbit propagators pick Hermite-cubic, the generic propagator picks Linear (Hermite is invalid for arbitrary-dimension states, which is why a single concrete default couldn't work on the shared config). The explicit `.with_interpolation_method(...)` calls and the long comment are removed from all four examples, replaced with your shortened comment.

src/estimation/ekf.rs
duncaneddy: Fixed in e9dcf3c8 — `propagate_to` on both filters now accepts targets in either direction (the dynamics layer already handled backward propagation), so backward smoothing passes work. The backwards-rejection tests were updated to assert backward propagation succeeds with covariance growing by process noise scaled with |dt|.

src/estimation/ekf.rs
duncaneddy: Fixed in e9dcf3c8 — `predict_only(current_epoch, target_epoch)` with dt computed internally, on both filters.

src/estimation/ekf.rs
duncaneddy: You're right that this is the discrete-time approximation: Q is applied once per predict interval (identically in `process_observation` — this was deliberate for consistency), rather than the continuous-discrete form ∫Φ(t,τ)·Q_c·Φᵀ(t,τ)dτ accumulated through the propagation, where noise injected early in the interval also gets mapped through the dynamics. For the short intervals the filters see in practice (observation cadences and the example's 60 s gap-stepping) the difference is second-order, but for a long single-shot `propagate_to` it grows. Two ways to close it: (a) sub-step the filter-level prediction at a configurable max-dt, adding Q per sub-step so earlier noise is propagated by later sub-steps' STMs — cheap, no propagator changes, converges to the integral; (b) augment the variational equations to integrate the noise term alongside the STM — exact, but a bigger propagator change. I'd suggest (a) as a follow-up (e.g. a `max_noise_dt` field on `ProcessNoiseConfig`) rather than in this PR, since it changes covariance numerics for both filters and deserves its own validation — but happy to do it here if you prefer.

src/estimation/mod.rs
duncaneddy: Fixed in e9dcf3c8 — full rename to `ECEFPositionMeasurementModel`/`ECEFVelocityMeasurementModel`/`ECEFStateMeasurementModel` (no deprecated aliases), matching the Python names, with the targeted clippy allow.

src/estimation/sensors.rs
duncaneddy: Fixed in e9dcf3c8 — both the sensor and the measurement model now compute topocentric geometry via ENZ (`relative_position_ecef_to_enz` + `position_enz_to_azel`); verified the azimuth/elevation/range conventions are identical to the SEZ path (existing geometry tests pass unchanged).

src/estimation/sensors.rs
duncaneddy: Done in e9dcf3c8 — added `AzimuthConstraint` (wrap-aware window) and `RangeConstraint` to the access module (with bindings, standalone examples 20547d13, and docs 2d3bd7e9), and `SimpleSSNSensor` now builds its field-of-view from those `AccessConstraint`s and evaluates them through the shared machinery. `with_constraint()` is the extension point — e.g. adding a `LookDirectionConstraint` for right-looking passes works out of the box.

src/estimation/ukf.rs
duncaneddy: Fixed in e9dcf3c8 alongside the EKF — same direction-aware handling and `predict_only(current_epoch, target_epoch)` signature.

src/estimation/ukf.rs
duncaneddy: Same answer as the EKF thread — it's the standard discrete-time approximation, consistent between propagate_to and process_observation; see the proposal there for the sub-stepped accumulation follow-up.

docs/examples/ssn_tracking.md
duncaneddy: Fixed in 7e2c71a9 — the section is now just titled 'Batch Least Squares'.

docs/examples/ssn_tracking.md
duncaneddy: Fixed in 7e2c71a9 — replaced with a tip admonition in your wording, linking to the interpolation documentation. (And with 13f92ff0/f3346398 Hermite-cubic really is the default now, so the tip describes actual behavior.)

docs/examples/ssn_tracking.md
duncaneddy: Done in 3ca24c04 — each sensor gets a stable palette color used for both its coverage cone on the network map and its measurement traces, with a legend carrying the mapping.

docs/examples/ssn_tracking.md
duncaneddy: Fixed — docs prose corrected in 7e2c71a9 (optical/radec vs uncalibrated-radar are now distinct categories, Haystack described as a radar lacking calibration values), and the API reworked in e9dcf3c8 per your suggestion: `from_locations` loads every radar site (uncalibrated ones default to zero noise/bias, overridable via `with_noise()`/`with_bias()`), and `from_locations_calibrated` restricts to fully-calibrated sites — the example uses the calibrated set.

docs/examples/ssn_tracking.md
duncaneddy: Fixed in 7e2c71a9 — the CI explanation is gone; the section now leads with the BLS trade (batch minimization of the residual cost over the whole arc vs recursive updates with covariance growth between passes) and notes only that it is substantially slower.

docs/examples/ssn_tracking.md
duncaneddy: Done in 03292252 — the log-scale 3σ row is dropped; the figure now shows radial/along-track/cross-track error lines over shaded ±3σ covariance bands (rotated into the RTN frame of the truth state via `rotation_eci_to_rtn`), followed by the total position error, for both filters.

src/estimation/ekf.rs
duncaneddy: Implemented in 558574aa per your call: `ProcessNoiseConfig` gains `max_noise_dt` — predict intervals longer than the cap are split into chunks, with the covariance propagated through each chunk and `Q_c·δt` added per chunk, so earlier noise is propagated by later chunks' dynamics (the left-endpoint approximation of ∫ΦQ_cΦᵀdτ). Applies only to `scale_with_dt=true`; default `None` preserves single-shot behavior exactly. The cornerstone test asserts one sub-stepped `propagate_to` equals a chained sequence of single-shot calls at the same cadence, plus a shear-physics test showing the along-track growth the single-shot form misses. Exact variational Q_d integration filed as #408.

src/estimation/ukf.rs
duncaneddy: Implemented for the UKF in 558574aa as well — each chunk runs a full unscented transform (sigma-point regeneration per chunk, documented cost of 2n+1 propagations per chunk) with `Q_c·δt` added per chunk. Same definitional-equivalence and physics tests as the EKF. Exact formulation tracked in #408.

examples/examples/ssn_tracking.py
```diff
@@ -0,0 +1,534 @@
...
+from plotly.subplots import make_subplots
+
+bh.initialize_eop()
+
+# Configuration
+MEAS_INTERVAL = 15.0  # seconds between measurements during a pass
+DURATION = 6 * 3600.0  # tracking duration (seconds)
+# The run numbers quoted in the docs depend on this seed AND on the rand
+# crate's StdRng stream; a rand version bump can shift the exact figures while
+# the example's tolerance-based asserts still pass.
```
duncaneddy: Remove this comment not relevant to the user.

docs/examples/ssn_tracking.md
```diff
@@ -0,0 +1,119 @@
...
+
+First, we'll import the necessary libraries, initialize Earth orientation parameters, and configure the tracking arc:
+
+``` python
+--8<-- "./examples/examples/ssn_tracking.py:preamble"
+```
+
+## Load the Sensor Network
+
+We load the Vallado SSN sensor sites and build sensors from the radar sites. The dataset's sites split into two sensor types: optical (`radec`) trackers (Socorro, Maui, Diego Garcia, MSSS, MOTIF, Moron), which `SimpleSSNSensor` doesn't support, and radar/phased-array/mechanical (`azel_range`) sites. `SimpleSSNSensor.from_locations()` builds a sensor for every radar site, defaulting the two that lack Table 4-3/4-4 calibration (Haystack, HAX) to zero noise and bias -- overridable with `with_noise()`/`with_bias()`. `from_locations_calibrated()` restricts to the fully-calibrated sites; this example uses the calibrated set so simulated measurement noise always matches Table 4-4:
```
duncaneddy: `radec` is a terrible short-hand for optical trackers. There also isn't a good reason we can't support them. We just need to use an "AzElMeasurementModel` in addition to the azel/range variant, and use that for the optical sensors.

docs/library_api/estimation/azelrange_measurement_model.md
```diff
@@ -0,0 +1,16 @@
+# AzElRangeMeasurementModel
+
+Topocentric azimuth/elevation/range measurement model for ground sensors.
+
+---
+
+::: brahe.AzElRangeMeasurementModel
+    options:
+      show_root_heading: true
+      show_root_full_path: false
```
duncaneddy: This should be moved over to the existing  Measurement Models API reference page and this page deleted since it's duplicative.

docs/library_api/estimation/simple_ssn_sensor.md
```diff
@@ -0,0 +1,17 @@
+# SimpleSSNSensor
```
duncaneddy: Name this Page "Sensor Models"

examples/examples/ssn_tracking.py
duncaneddy: Fixed in 1f5d3e0f — the seed/rand implementation comment is removed from the walkthrough.

docs/examples/ssn_tracking.md
duncaneddy: Done in 34190a03 — `SensorType::RaDec` is renamed `Optical` (dataset value "radec" → "optical") and optical trackers are now fully supported: a new angles-only `AzElMeasurementModel` sits beside `AzElRangeMeasurementModel` (shared ENZ geometry, wrap-aware azimuth residual, and residual-based finite-difference Jacobian), and `SimpleSSNSensor` builds optical sensors measuring [az, el] with `measurement_model()` returning the model matching the sensor type. `from_locations` now returns all 21 sites and `from_locations_calibrated` 16 (13 radar + 3 calibrated optical). Tests (c39614fc), examples (1f5d3e0f), and docs (1f42b80e) updated. Two follow-ons surfaced while enabling this: a `reinitialize` step-size bug that stalled BLS re-propagation once optical observations were included (fixed in de511ab8 — the mixed 16-sensor network now converges in 4 iterations), and stale radar-only docstrings corrected in 94d72cc4.

docs/library_api/estimation/azelrange_measurement_model.md
duncaneddy: Fixed in 1f42b80e — the standalone page is deleted; `AzElRangeMeasurementModel` (and the new `AzElMeasurementModel`) are documented on the Measurement Models reference page, with nav and inbound links updated.

docs/library_api/estimation/simple_ssn_sensor.md
duncaneddy: Fixed in 1f42b80e — the page is renamed "Sensor Models" (`sensor_models.md`), with nav and links updated.

docs/examples/ssn_tracking.md
duncaneddy: Follow-up in a7141e93 — the type is now named parallel to `AzElRange`: `SensorType::AzEl` with dataset value "azel" (matching "azel_range"), and `SensorType` is now bound to Python 1:1 (`bh.SensorType.AZEL_RANGE` / `bh.SensorType.AZEL`, AngleFormat pattern) with a `sensor_type` getter on `SimpleSSNSensor`, documented on the Sensor Models API page.

PR #410, opened 2026-07-20
Convert recoverable panics to Result-based error handling

### LINKED ISSUE #279: [Bug]: Panics on recoverable errors instead of returning Result
### Pre-submission Checklist

- [x] I have searched existing issues to ensure this bug hasn't already been reported
- [x] I have checked the [documentation](https://duncaneddy.github.io/brahe/) for relevant guidance
- [x] I can reproduce this bug with the latest version of Brahe

### Description

Several error conditions that can be recoverable currently cause panics instead of returning idiomatic `Err` in Rust:

  - Rust API: `expect("Kepler's equation failed to converge")` and others should imho be treated as Result and not panic.
  - Python binding: dynamics callback failures in pymodule/ hit panic!(), crashing the interpreter instead of raising a Python exception.

Fixing the Rust API will requires changing some function signatures (e.g. dynamics functions from Fn(...) -> DVector<f64> to Fn(...) -> Result<...>), which would be a semver-breaking change, so I'm raising this as an issue first before fixing it

### Language

Rust

### Python Reproduction Code

```python

```

### Rust Reproduction Code

```rust

```

### Expected Behavior

Recoverable errors should be surfaced as Err-type via Result and handled gracefully

See Rust docs: https://doc.rust-lang.org/std/result/

### Actual Behavior

Recoverable errors crash process

### Brahe Version

1.3.4

### Python Version (if applicable)

_No response_

### Rust Version (if applicable)

_No response_

### Operating System

macOS

### Additional Context

Happy to open a MR that fixes this, but wanted to discuss the potential breaking change to the Rust API first

duncaneddy:
# Pull Request

## Description

Implements the comprehensive error-handling reevaluation from #279. An audit of every `unwrap`/`expect`/`panic!` in production code (994 sites in the Rust core, 178 in the Python bindings, after excluding tests and doc examples) found the actionable sites concentrated in six API families; this PR converts all of them to `Result<_, BraheError>` in a single breaking release while leaving idiomatic panics (lock poisoning, construction-guaranteed invariants, provably infallible operations — about 60% of sites) in place.

The three structural changes:

1. **Dynamics callbacks and integrators.** `DStateDynamics`, `SStateDynamics`, control inputs, and the Jacobian/sensitivity provider traits return `Result`, giving runtime failures inside an integration step an error channel for the first time. `DIntegrator`/`SIntegrator` `step*` methods return `Result` (adaptive-step non-convergence is `NumericalError` instead of a panic; missing-`dt`/missing-provider contract violations are `PropagatorError`). The force-model surface (`accel_gravity_spherical_harmonics*`, `accel_third_body*`) propagates degree/order and SPK ephemeris errors, and `SStatePropagator`/`DStatePropagator` `step`/`step_by`/`step_past`/`propagate_steps`/`propagate_to`/`propagate_trajectory` return `Result<(), BraheError>` so integration failures reach the caller. SGP4's graceful decay-termination semantics are preserved.

2. **Trajectory mutation and conversion.** `Trajectory::add`, the `add_full`/covariance/STM/sensitivity mutators, constructors, eviction builders, and the `OrbitalTrajectory` conversion family (`to_eci`/`to_ecef`/`to_gcrf`/`to_eme2000`/`to_itrf`/`to_keplerian`/`from_orbital_data`) return `Result`, propagating the frame-conversion and NAIF-resolution errors previously swallowed into panics ("panics ... matching this method's infallible signature"). `add_full` is transactional: a rejected write leaves the trajectory unchanged.

3. **Constructor validation.** `KeplerianPropagator`, `SGPPropagator::with_output_format`, `PointLocation`/`PolygonLocation`, `SamplingConfig::generate_sample_epochs`, the interpolation utilities, `WalkerConstellationGenerator`, `Epoch::from_day_of_year`, threading configuration, and `ParameterSource::get_value` validate inputs with `Result` instead of panicking.

**Deliberate non-conversions.** Three trunk functions keep infallible signatures with documented `# Panics` preconditions and actionable messages, because their failure modes are environment misconfiguration rather than caller-handleable conditions: the EOP-dependent frame rotations (`bias_precession_nutation`, `polar_motion`), the UT1 branches of `time_system_offset` (141 internal call sites — converting it would make every `Epoch` accessor fallible), and the Kepler solve inside `state_koe_to_eci` (the leaf `anomaly_mean_to_eccentric` already returns `Result` for callers who want it). The new `ensure_global_eop_coverage(mjd_start, mjd_end)` API provides the recoverable path at the right layer: validate EOP coverage once at setup instead of per call.

**Python bindings.** All 1:1 mirrors updated. Python callback trampolines (dynamics, control, Jacobian, sensitivity, measurement models) no longer panic through the interpreter: the original Python exception is stashed and re-raised with its traceback at the binding boundary instead of surfacing as `pyo3_runtime.PanicException` (which derives from `BaseException` and evades `except Exception`). Non-C-contiguous numpy inputs raise `ValueError` instead of panicking. `par_propagate_to` now rejects propagators carrying Python callbacks — construction-time dynamics/control or Python-backed event detectors (`ValueEvent`/`BinaryEvent`, or any event with a Python callback) — because those callbacks would deadlock on the GIL from rayon worker threads; Rust-native events remain allowed in parallel.

The audit also surfaced independent defects fixed here: a dead convergence guard in the ECEF→geodetic iteration, silent wrong-value parsing of malformed ICGEM gravity headers, a panic on sub-second epoch strings longer than nine digits, an NTP-clock-step panic in the EOP cache age check, and stale docs claiming the global EOP accessors panic when they return `Err`.

External code review ran three adversarial rounds (add_full partial-mutation atomicity including the zero-column-sensitivity trigger, stale Python-exception slots in parallel propagation, and the event-detector GIL bypass); all findings are fixed with regression tests and the final round approved with zero findings.

### Added

- `ensure_global_eop_coverage(mjd_start, mjd_end)` (Rust and Python) validates at setup time that the global EOP provider covers an epoch range, honoring the provider's extrapolation setting.
- `KeplerianPropagator::validate_format` rejects unsupported frame/representation/angle-format combinations at construction, including body-centered inertial frames (Earth-only propagator).

### Changed

- **Breaking:** Dynamics, control, Jacobian, and sensitivity callback types and provider traits return `Result<_, BraheError>`; `DIntegrator`/`SIntegrator` `step*` methods return `Result`; adaptive integrators return `NumericalError` on non-convergence instead of panicking.
- **Breaking:** `SStatePropagator`/`DStatePropagator` `step`, `step_by`, `step_past`, `propagate_steps`, `propagate_to`, and `propagate_trajectory` return `Result<(), BraheError>`; `SOrbitPropagator`/`DOrbitPropagator` `set_initial_conditions` returns `Result`; `par_propagate_to_s`/`par_propagate_to_d` return `Result`.
- **Breaking:** `accel_gravity_spherical_harmonics*` and `accel_third_body*` return `Result<Vector3<f64>, BraheError>`.
- **Breaking:** `Trajectory::add`, trajectory constructors, eviction builders, `add_full` and covariance/STM/sensitivity/acceleration mutators, and the `OrbitalTrajectory` conversion family return `Result`; trajectory `covariance_at` getters return `Result<Option<_>>`.
- **Breaking:** `KeplerianPropagator` constructors, `SGPPropagator::with_output_format`, `PointLocation::new`, `SamplingConfig::generate_sample_epochs`, math interpolation utilities, `WalkerConstellationGenerator::new`, `Epoch::from_day_of_year`, `set_num_threads`/`set_max_threads`/`set_ludicrous_speed`, and `ParameterSource::get_value` return `Result` on invalid input.
- **Breaking:** `BraheError` is `#[non_exhaustive]`; custom-frame rotation/omega callbacks return `Result`.
- Python: raising callbacks re-raise the original exception (with traceback) instead of `pyo3_runtime.PanicException`; non-contiguous numpy arrays raise `ValueError`; `par_propagate_to` rejects propagators carrying Python callbacks (constructor callbacks or Python-backed event detectors) with a GIL explanation.
- Panic messages on the intentionally infallible EOP/time/Kepler trunk functions name the failing epoch or eccentricity and the remedy, with `# Panics` documentation.

### Fixed

- `GravityModel::from_bufreader` returns `ParseError` for empty, truncated, or value-less ICGEM files instead of panicking, and no longer silently uses a header key as its value when the value is missing.
- `Epoch::from_string` no longer panics on sub-second strings longer than nine digits (u32 underflow); they now parse correctly.
- The ECEF→geodetic convergence guard was dead code (`while iter < 10` vs `iter == 20`); the iteration is bounded and documented instead.
- `CachingEOPProvider::file_age` no longer panics when the wall clock steps backwards (NTP adjustment).
- Global EOP accessor docs no longer claim panics for conditions that return `Err`.
- `add_full` is atomic across all trajectory types: a rejected write (including zero-column sensitivity) no longer partially enables storage.
- The groundstation GeoJSON loader skips invalid-coordinate features with a warning instead of panicking.

## Related Issue

Closes #279

## Note to Reviewers

The keep-infallible line for the three trunk functions (EOP-dependent rotations, `time_system_offset`, `state_koe_to_eci`) is the main design decision and is documented in the Description; it follows the slice-indexing precedent (`v[i]` panics, `v.get(i)` exists) with `ensure_global_eop_coverage` as the checked alternative. Verification: full Rust suite (4,939), Python suite (3,298), doctests (769), all standalone examples (Rust 327, Python 336), plot spot-checks, and `just check` all pass. Tests that previously asserted `#[should_panic]`/`PanicException` on converted sites now assert `Err`/`BraheError`; no tests were deleted.


tests/trajectories/test_orbit_trajectory.py
```diff
@@ -26,7 +26,7 @@
     state_gcrf_to_eme2000,
     state_eme2000_to_gcrf,
 )
-from brahe._brahe import PanicException
+from brahe._brahe import BraheError
```
duncaneddy: This should just be `from brahe import BraheError`

src/integrators/dp54.rs
```diff
@@ -72,15 +73,15 @@ impl<const S: usize, const P: usize> DormandPrince54SIntegrator<S, P> {
...
             };
             if let Some(ref ctrl) = self.control {
-                k0 += ctrl(t, &state, params);
+                k0 += ctrl(t, &state, params)?;
             }
             k.set_column(0, &k0);
 
             if compute_phi || compute_sens {
-                let a0 = self.varmat.as_ref().unwrap().compute(t, &state, params);
+                let a0 = self.varmat.as_ref().unwrap().compute(t, &state, params)?;
```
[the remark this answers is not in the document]
duncaneddy: Fixed in 2e8d46cc. Both `step_internal` implementations now return a `PropagatorError` up front when `step_with_varmat`/`step_with_sensmat` is called without the corresponding provider configured, instead of silently skipping the variational computation. Added tests covering the missing-varmat and missing-sensmat paths for both the S and D integrators.

src/propagators/dnumerical_propagator.rs
```diff
@@ -1430,7 +1444,7 @@ impl DNumericalPropagator {
                     params_ref,
                     phi,
                     Some(dt_requested),
-                );
+                )?;
```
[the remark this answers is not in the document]
duncaneddy: Fixed in 2e8d46cc. The dense propagators now clone the STM/sensitivity matrices instead of `take()`-ing them, so a failed step leaves them intact for a retry or inspection. Added `test_step_errors_when_mass_unconfigured_for_drag`/`_srp` regression tests that verify the matrices survive a failed step.

src/eop/global.rs
```diff
@@ -536,6 +536,76 @@ pub fn initialize_eop() -> Result<(), BraheError> {
...
+    if !provider.is_initialized() {
+        return Err(BraheError::EOPError(
+            "Global EOP provider is not initialized. Initialize one with \
+             set_global_eop_provider or initialize_eop before use."
+                .to_string(),
+        ));
+    }
+
+    if provider.extrapolation() == EOPExtrapolation::Error
+        && (mjd_start < provider.mjd_min() || mjd_end > provider.mjd_max())
```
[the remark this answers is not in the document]
duncaneddy: Fixed in 2e8d46cc. Under `Error` extrapolation, `ensure_global_eop_coverage` now bounds the usable range by `min(mjd_max, mjd_last_dxdy)`, so a window that extends past the end of the shorter dX/dY series is rejected. Added `test_ensure_global_eop_coverage_dxdy_gap_error_extrapolation` and documented the dX/dY bound in the Rust and Python docstrings.

tests/trajectories/test_orbit_trajectory.py
duncaneddy: Fixed in 2e8d46cc — now `from brahe import BraheError`.

duncaneddy:
Pushed updates addressing the review feedback and the Codecov patch-coverage report:

**Review fixes** (all four inline threads answered above): dp54 missing-provider guards, STM/sensitivity preservation on failed steps, dX/dY bound in `ensure_global_eop_coverage`, and the `BraheError` import.

**Coverage**: added error-path tests across the integrators (rk4/rkf45/rkf78/dp54/rkn1210 dt/provider/control guards), propagators (Keplerian/SGP format validation, force-model config, trait defaults, failing-dynamics paths), trajectories (dimension/frame/covariance error arms), interpolation range guards, gravity-model parsing and out-of-range harmonics, and EOP coverage checks. Remaining uncovered lines are intentional: SPK-kernel-gated third-body arms, documented panic paths kept infallible by design (`conversions.rs`, `cartesian.rs`, `gcrf_itrf.rs`, `location.rs`), `unreachable!` arms in SGP element parsing, and a defensively-unreachable branch in `properties.rs` (the offset>duration check guarantees at least one sample).

**Latent defect found by the new tests**: `KeplerianPropagator` derived `current_epoch()`/`current_state()` from the last entry of its epoch-sorted trajectory, so backward propagation (`set_step_size(-x)` + `step_past`/`propagate_to`) never advanced the reported epoch and looped forever. The propagator now tracks its current epoch and state explicitly (matching `SGPPropagator`).

**Merged `main`** (#406, #407, #409, #411): resolved conflicts in the EKF/UKF predict paths (kept the new process-noise sub-stepping structure, applied the fallible `propagate_to` conversion to it), adapted the new SSA sensor code and tests to the `Result` APIs (`PointLocation::new`, `propagate_to`, `step`), and updated the new estimation tests to expect `brahe.BraheError` in line with the rest of this PR.

Gates on the merged branch: 5,127 Rust lib tests, 776 doctests, 3,377 Python tests, clippy — all green.

src/frames/custom.rs
```diff
@@ -171,11 +171,14 @@ mod tests {
     use crate::time::TimeSystem;
 
     /// Uniform rotation about z at `rate` rad/s from `t0`.
-    fn spin_z(t0: Epoch, rate: f64) -> impl Fn(Epoch) -> SMatrix3 + Send + Sync + Clone {
+    fn spin_z(
+        t0: Epoch,
+        rate: f64,
+    ) -> impl Fn(Epoch) -> Result<SMatrix3, BraheError> + Send + Sync + Clone {
```
duncaneddy, line 177: I don't get why this needs to be a Result since there is no possible error potential for it.

duncaneddy: The test helper just inherits the callback type's signature — the `Result` lives on `CustomFrameRotation`/`CustomFrameOmega` themselves, and there are two real failure paths that need it:

1. **Python callbacks.** `register_custom_frame` is exposed to Python, and a Python callback can raise or return a non-3x3 value. On main the trampoline `.expect()`s in both cases, so the failure surfaces as `pyo3_runtime.PanicException` (which derives from `BaseException` and evades `except Exception`) — the exact failure class this PR removes for the dynamics/control callbacks. The `Result` signature lets the trampoline convert the exception into a `BraheError` that reaches the caller at the transform call site.

2. **Rust orientation models backed by data.** The motivating use case in the module docs is spin states from tabulated data (e.g. DAMIT); a lookup or interpolation outside the table's coverage has no way to report failure through an infallible signature except panicking mid-transform.

Since the router functions (`rotation_frame_to_frame` etc.) already return `Result`, propagating the callback error costs nothing at the API surface. For pure analytic models the cost is wrapping the return in `Ok(...)`. This also keeps the custom-frame callbacks consistent with the dynamics, control, and Jacobian callbacks converted in this PR.

PR #416, opened 2026-07-21
Add JPL SBDB and Horizons SPK clients

### LINKED ISSUE #402: Add Horizons client for targeted SPK ephemeris generation
Add a small-scope client for the JPL Horizons system (https://ssd-api.jpl.nasa.gov/doc/horizons.html) for generating, downloading, and loading targeted SPK files for bodies without packaged DE-kernel coverage (e.g. Ceres and other small bodies).

Scope is deliberately narrow: request an SPK for a given body/time span, cache it under the brahe cache directory, and load it into the SPICE registry so frame transforms and third-body perturbations work for that body. This unblocks proper perturbation modeling in small-body scenarios like the Dawn-at-Ceres example (docs/examples/dawn_ceres_orbit.md), which currently propagates gravity-only because the DE kernels carry no Ceres ephemeris and no other major kernels (that I can find) do either. They have to be custom-requested.

Arose during review of #399.

duncaneddy:
# Pull Request

## Description

Adds JPL Small-Body Database (SBDB) and Horizons SPK clients under `brahe.datasets`, and the SPK segment type-21 reader they require, to support a natural workflow for bodies without packaged DE-kernel coverage: resolve a name to its NAIF/SPK ID via SBDB, generate and cache a targeted SPK from Horizons, load it into the SPICE registry, and use it with `CentralBody.Custom` for third-body and solar radiation pressure forces. The Dawn-at-Ceres example is reworked to demonstrate the end-to-end path with real Sun and Jupiter third-body and SRP perturbations.

### Added
- JPL Small-Body Database (SBDB) Lookup client (`brahe.datasets.sbdb.SBDBClient` returning `SBDBObject`) that resolves a name or designation to its NAIF/SPK ID and SI physical parameters (GM in m³/s², radius in m), with on-disk caching.
- JPL Horizons SPK client (`brahe.datasets.horizons`: `HorizonsClient`, `HorizonsSPKRequest`, `HorizonsSPKResponse`) that generates, caches, and loads targeted small-body SPK kernels, with a response handle exposing the cached `.bsp` path, raw bytes, and a direct load into the SPICE registry.
- SPK segment type 21 (Extended Modified Difference Arrays) reader, enabling Horizons-generated small-body SPKs (e.g. Ceres) to load and answer position, velocity, and state queries; the interpolation is a port of CSPICE `spke21` validated bit-exact against CSPICE `spkgeo` on a real Ceres kernel.
- Learn and Library-API documentation pages for the SBDB and Horizons clients.

### Changed
- Reworked the Dawn-at-Ceres example to resolve Ceres via SBDB, load a Ceres SPK from Horizons, and model Sun and Jupiter third-body plus solar radiation pressure perturbations around the custom-defined body.
- Refactored the SPK reader behind a segment abstraction so a resolved chain can mix segment types, combining a type-21 small-body segment with the existing Chebyshev (types 2 and 3) segments from the DE kernels.

## Related Issue
Closes #402.

## Note to Reviewers
Type-21 SPK support was added mid-branch because Horizons emits small-body SPKs only as type 21 (Extended MDA) and the reader was previously Chebyshev-only, so it is a prerequisite for the example's perturbations rather than optional scope. The Horizons Ceres SPK is Sun-centered (NAIF 10) and chains through `de440s`; SBDB returns Ceres SPK-ID `20000001` (8-digit scheme). The reworked example is not network-gated: its SBDB lookup and Ceres SPK are warmed and cached in the example CI job. Type-21 correctness is established three ways — a citation-backed cross-check against the NAIF `spke21`/`spkw21` spec, a line-by-line port review, and bit-exact agreement with CSPICE on the real kernel.


docs/examples/dawn_ceres_orbit.md
```diff
@@ -34,6 +37,27 @@ and rotation of the body-fixed frame at any epoch.
...
+## Resolving Ceres and Loading an Ephemeris
+
+Ceres's NAIF/SPK ID and SI physical parameters - gravitational parameter and
+mean radius - come from the JPL Small-Body Database (SBDB) rather than being
+hardcoded. A targeted SPK covering the propagation span is then generated and
+loaded from JPL Horizons, so the Sun/Jupiter third-body accelerations and the
+solar radiation pressure model below have an ephemeris to resolve Ceres's
+position against. Both the SBDB lookup and the Horizons SPK are cached under
+the brahe cache directory, so the network is used only on the first run per
+machine.
```
duncaneddy: Remove

> Both the SBDB lookup and the Horizons SPK are cached under
the brahe cache directory, so the network is used only on the first run per
machine.

This is an implementation detail that isn't needed for the example.

docs/examples/dawn_ceres_orbit.md
```diff
@@ -122,12 +139,34 @@ construction:
...
+
+``` python
+--8<-- "./examples/examples/dawn_ceres_orbit.py:baseline_comparison"
+```
+
+At a ~375 km, ~5.4-hour LAMO orbit, the solar and Jovian third-body
+accelerations and SRP are small enough that they perturb the trajectory by
+tens of meters over 2 days (~9 revolutions) rather than reshaping it, but the
+divergence from the two-body baseline is measurable and grows with time - a
+signature a pure two-body run cannot reproduce.
```
duncaneddy, line 166: It would be interesting to generate a plot of either position/velocity state differences or orbital elements for each propagator over the horizon.

docs/learn/datasets/horizons.md
```diff
@@ -0,0 +1,51 @@
+# JPL Horizons SPK Generation
+
+The `brahe.datasets.horizons` module generates, caches, and loads targeted SPK
+(`.bsp`) kernels for small bodies without packaged DE-kernel coverage via the
+[JPL Horizons API](https://ssd-api.jpl.nasa.gov/doc/horizons.html). This
+unblocks third-body and solar radiation pressure perturbations for bodies such
+as Ceres, which carry no ephemeris in the DE kernels brahe ships.
```
duncaneddy: "which carry no ephemeris in the DE kernels brahe ships.
" -> which are not contained in the DE kernels and there for must be loaded separately.

docs/learn/datasets/horizons.md
```diff
@@ -0,0 +1,51 @@
...
+
+t0 = bh.Epoch.from_datetime(2015, 12, 1, 0, 0, 0.0, 0.0, bh.TimeSystem.TDB)
+t1 = bh.Epoch.from_datetime(2016, 3, 1, 0, 0, 0.0, 0.0, bh.TimeSystem.TDB)
+
+request = bh.datasets.horizons.HorizonsSPKRequest.for_spkid(20000001, t0, t1)
+response = bh.datasets.horizons.HorizonsClient().get_spk(request)
+
+response.load()          # load into the SPICE registry
+print(response.path)     # cached .bsp path
+```
```
duncaneddy: Should be implemented as a tested code example in both python and rust as is the standard for examples.

docs/learn/datasets/sbdb.md
```diff
@@ -0,0 +1,41 @@
...
+import brahe as bh
+
+client = bh.datasets.sbdb.SBDBClient()
+ceres = client.lookup("Ceres")
+
+print(ceres.naif_id())   # 20000001
+print(ceres.full_name)   # "1 Ceres (A801 AA)"
+print(ceres.gm)          # gravitational parameter [m^3/s^2]
+print(ceres.radius)      # mean radius [m]
+```
```
duncaneddy: Should be implemented as a tested code example in both python and rust as is the standard for examples.

docs/learn/datasets/sbdb.md
```diff
@@ -0,0 +1,41 @@
...
+[JPL Small-Body Database Lookup API](https://ssd-api.jpl.nasa.gov/doc/sbdb.html).
+Its primary use is obtaining the NAIF ID needed to request a targeted SPK from
+[Horizons](horizons.md) and to define a body via `CentralBody.Custom`.
+
+`SBDBClient.lookup(sstr)` returns an `SBDBObject`. Responses are cached under
+`$BRAHE_CACHE/sbdb` (default `~/.cache/brahe/sbdb`) and reused until the cache
+age exceeds the client's `cache_max_age` (default 30 days); pass
+`cache_max_age=0` to always refetch. Ambiguous search strings and unknown
+objects raise an error.
+
```
duncaneddy, line 14: Add a Tip admonition that the SBDB can be used in conjunction with the Horizons API to get SPK files for other bodies not included in major DE SPK or Planetary system SPK distributions.

examples/examples/dawn_ceres_orbit.py
```diff
@@ -47,24 +52,46 @@
...
+ceres_obj = bh.datasets.sbdb.SBDBClient().lookup("Ceres")
+CERES_NAIF_ID = ceres_obj.naif_id()  # 20000001 (SBDB SPK-ID scheme)
+GM_CERES = ceres_obj.gm  # m^3/s^2
+R_CERES = ceres_obj.radius  # m
+print(f"Ceres: {ceres_obj.full_name}, NAIF ID {CERES_NAIF_ID}")
+print(f"  GM = {GM_CERES:.6e} m^3/s^2, radius = {R_CERES / 1e3:.1f} km")
+
+# Load a de440s kernel so Sun/Jupiter positions are available, then fetch and
+# load a Ceres SPK covering the propagation span.
+bh.load_spice_kernel("de440s")
```
duncaneddy: I believe this should just be `bh.load_common_spice_kernels()` which includes the DE440s data.

examples/examples/dawn_ceres_orbit.py
```diff
@@ -109,11 +136,34 @@ def ceres_omega(epc=None):
...
+    ],
+)
+
+force_config = bh.ForceModelConfig.for_body(
+    ceres,
+    bh.GravityConfiguration.point_mass(),
+    srp=srp,
+    third_body=[
+        bh.ThirdBody.SUN,
+        bh.ThirdBody.Custom(name="Jupiter Barycenter", naif_id=5, gm=1.26686534e17),
```
duncaneddy: Instead of a custom definition I think we can just rely on the ThirdBody.JUPITER_BARYCENTER instead of a custom definition here.

examples/examples/dawn_ceres_orbit.py
```diff
@@ -175,24 +226,59 @@ def ceres_omega(epc=None):
...
-assert np.all(np.abs(incs_deg - 90.0) < 0.1), (
-    f"Inclination departed from the 90 deg polar design rel. Ceres pole: "
-    f"{incs_deg.min():.4f} to {incs_deg.max():.4f} deg"
+# The plane stays polar to well under a hundredth of a degree.
+assert max_excursion < 1e-3, (
+    f"Inclination departed further than expected from the 90 deg polar design "
+    f"rel. Ceres pole: {incs_deg.min():.6f} to {incs_deg.max():.6f} deg"
 )
 assert radii_km.min() * 1e3 > R_CERES, "Orbit descended below the Ceres surface"
 assert radii_km.max() * 1e3 < R_CERES + 1000e3, (
```
duncaneddy: Don't need hard asserts since that's brittle in a running example. Just prints are fine.

src/utils/cache.rs
```diff
@@ -215,6 +215,63 @@ pub fn get_star_catalogs_cache_dir() -> Result<String, BraheError> {
...
+///
+/// # Returns
+///
+/// * `String` - A 16-character lowercase hexadecimal digest.
+pub fn short_hash(input: &str) -> String {
+    use std::hash::{Hash, Hasher};
+    let mut hasher = std::collections::hash_map::DefaultHasher::new();
+    input.hash(&mut hasher);
+    format!("{:016x}", hasher.finish())
+}
```
duncaneddy, line 278: Instead of a short hash can we just use a proper MD5 or SHA-256?

src/datasets/sbdb/client.rs
```diff
@@ -0,0 +1,275 @@
...
+    ///
+    /// # Returns
+    ///
+    /// * `Ok(SBDBObject)` - The resolved object.
+    /// * `Err(BraheError)` - On ambiguous/no match, network, or parse errors.
+    pub fn lookup(&self, sstr: &str) -> Result<SBDBObject, BraheError> {
+        let url = format!(
+            "{}/sbdb.api?sstr={}&phys-par=1",
+            self.base_url,
+            urlencode(sstr)
```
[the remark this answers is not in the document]
duncaneddy: Added `&full-prec=1` to the SBDB lookup URL (dbe78e26) so GM/radius are returned at full precision before they are parsed, converted to SI, and cached.

examples/examples/dawn_ceres_orbit.py
```diff
@@ -47,24 +52,46 @@
...
 os.makedirs(OUTDIR, exist_ok=True)
 
+# --8<-- [start:body_resolution]
+# Resolve Ceres in the JPL Small-Body Database to get its NAIF/SPK ID and SI
+# physical parameters, then generate and load a targeted SPK from Horizons so
+# third-body and SRP perturbations resolve around Ceres. Both the SBDB query
+# and the SPK are cached under the brahe cache directory and reused on repeat
+# runs, so this hits the network only once per machine.
+ceres_obj = bh.datasets.sbdb.SBDBClient().lookup("Ceres")
+CERES_NAIF_ID = ceres_obj.naif_id()  # 20000001 (SBDB SPK-ID scheme)
```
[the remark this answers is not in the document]
duncaneddy: The example derives the frame center, the Horizons request, and the loaded SPK target all from the same `ceres_obj.naif_id()` value, so they cannot diverge: the custom body `naif_id` and `HorizonsSPKRequest.for_spkid(...)` are the one value returned by SBDB. Live SBDB returns `20000001` for Ceres (and the plotting registry uses the same 8-digit ID), so there is no `2000001`/`20000001` split at runtime and the perturbed run resolves Sun/Jupiter against the loaded kernel without a normalization step. Not changing the code here; happy to add a defensive assert if you prefer belt-and-braces.

docs/examples/dawn_ceres_orbit.md
duncaneddy: Removed (5b1f3a73).

docs/examples/dawn_ceres_orbit.md
duncaneddy: Added (63b7cf8c + 5b1f3a73): the example now propagates a two-body baseline and plots the perturbed-vs-two-body position (m) and velocity (mm/s) divergence over the span, embedded on the page below the baseline-comparison section.

docs/learn/datasets/horizons.md
duncaneddy: Reworded (5b1f3a73) to "which are not contained in the DE kernels and therefore must be loaded separately."

docs/learn/datasets/horizons.md
duncaneddy: Replaced with a tested standalone example in Python and Rust (`examples/datasets/horizons_spk.{py,rs}`, `FLAGS=["NETWORK"]`), embedded via `--8<--` Python/Rust tabs (c3343f13). Both run under `just test-examples`.

docs/learn/datasets/sbdb.md
duncaneddy: Replaced with a tested standalone example in Python and Rust (`examples/datasets/sbdb_lookup.{py,rs}`, `FLAGS=["NETWORK"]`), embedded via `--8<--` Python/Rust tabs (c3343f13). Both run under `just test-examples`.

docs/learn/datasets/sbdb.md
duncaneddy: Added a Tip admonition (5b1f3a73) on pairing SBDB with Horizons to obtain SPK files for bodies outside the major DE / planetary-system SPK distributions.

examples/examples/dawn_ceres_orbit.py
duncaneddy: Switched to `bh.load_common_spice_kernels()` (63b7cf8c).

examples/examples/dawn_ceres_orbit.py
duncaneddy: Switched to the built-in `bh.ThirdBody.JUPITER_BARYCENTER` (63b7cf8c), dropping the Custom definition.

examples/examples/dawn_ceres_orbit.py
duncaneddy: Replaced the hard asserts with a summary print (63b7cf8c); the example still computes and reports the divergence, inclination excursion, and radius band.

src/utils/cache.rs
duncaneddy: Switched `short_hash` to a truncated SHA-256 digest via the `sha2` crate (dbe78e26) — stable across platforms and Rust versions, unlike the previous `DefaultHasher`. (`sha2` is the RustCrypto/hashes SHA-256 crate and was already in the dependency tree.)

PR #436, opened 2026-08-16
Add batch evaluation primitives for vectorized transforms

duncaneddy:
# Pull Request

## Description

Adds the crate-private batch-evaluation primitives that the plural (`states_*`, `positions_*`, `rotations_*`) transformation functions in the rest of this stack are built on: `src/utils/batch.rs` with `PARALLEL_THRESHOLD`, `broadcast_len`, `pick`, `map_indices`, `batch_map`, `batch_zip`, and `batch_map_epochs`. Every slice argument follows one broadcast rule (length 1 or the common length N), a single shared epoch computes the transform context once, and evaluation runs on the global rayon pool above the threshold. No public API or Python changes in this PR; the fallible variants and Python-side helpers land with their first consumers in the following PRs.

### Added
- Crate-private `utils::batch` primitives (`batch_map`, `batch_zip`, `batch_map_epochs`, `PARALLEL_THRESHOLD`) with the length-1-or-N broadcast rule and shared-epoch context hoisting used by all vectorized transformation functions.

## Note to Reviewers
First layer of a six-PR stack (`vec/batch-infra` → `vec/frames-earth` → `vec/frames-planetary` → `vec/coordinates` → `vec/relative-motion` → `vec/orbits`). Tests cover ordering on the sequential and parallel paths, broadcast, empty input, and that the context closure runs exactly once for a shared epoch.

### DIFF
--- a/src/utils/batch.rs
+++ b/src/utils/batch.rs
@@ -11,6 +11,7 @@
 
 use rayon::prelude::*;
 
+use crate::time::Epoch;
 use crate::utils::errors::BraheError;
 use crate::utils::threading::get_thread_pool;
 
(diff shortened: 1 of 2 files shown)


src/utils/batch.rs
```diff
@@ -0,0 +1,402 @@
...
+///
+/// # Arguments
+///
+/// * `lens` - Lengths of each slice argument of a batch function
+///
+/// # Returns
+///
+/// The common length `N` (`1` if every length is `1`, `0` if any length is `0`
+/// and the rest are `1`), or an error naming the offending lengths.
+pub(crate) fn broadcast_len(lens: &[usize]) -> Result<usize, BraheError> {
```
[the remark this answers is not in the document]
duncaneddy: Addressed in c44ea482: every helper in `src/utils/batch.rs` now carries a `# Examples` section. The helpers are `pub(crate)`, so the examples use ```ignore``` fences (rustdoc doctests compile as an external crate and cannot reach crate-private items), following the existing precedent in `synodic_axes`.

src/utils/batch.rs
```diff
@@ -0,0 +1,462 @@
...
+use rayon::prelude::*;
+
+use crate::time::Epoch;
+use crate::utils::errors::BraheError;
+use crate::utils::threading::get_thread_pool;
+
+/// Minimum batch length at which batch primitives evaluate on the global
+/// rayon thread pool. Below this length evaluation is sequential, since the
+/// per-element kernels are sub-microsecond and thread hand-off would dominate.
+pub(crate) const PARALLEL_THRESHOLD: usize = 1024;
```
duncaneddy, line 21: This might be better set as an env variable / something settable similar to how the thread-counts is handled for the thread pool (e.g. `set_vectorization_length_threshold(n: i64)` / `get_vectorization_length_threshold()`. Can address this as a final PR stack.

src/utils/batch.rs
```diff
@@ -0,0 +1,462 @@
...
+///
+/// # Examples
+///
+/// ```ignore
+/// use crate::utils::batch::batch_map;
+///
+/// let doubled = batch_map(&[1.0, 2.0, 3.0], |x| 2.0 * x);
+/// assert_eq!(doubled, vec![2.0, 4.0, 6.0]);
+/// ```
+pub(crate) fn batch_map<T: Sync, U: Send>(inputs: &[T], f: impl Fn(&T) -> U + Sync) -> Vec<U> {
```
duncaneddy: nit: have function as consistent first argument and inputs as later arguments across `batch_*` functions

src/utils/batch.rs
duncaneddy: Added as the final PR of the stack (#451): `set_vectorization_length_threshold(n)` / `get_vectorization_length_threshold()` in Rust and Python, mirroring the `set_num_threads` pattern (process-global atomic, default 1024).

src/utils/batch.rs
duncaneddy: Done in d0f4a907 (and the follow-up commits on each layer): all `batch_*` primitives now take their kernel argument(s) first and the data slices after — `batch_map(f, inputs)`, `batch_zip(f, a, b)`, `batch_map_epochs(context, apply, epochs, inputs)`, and the `try_*` variants likewise — with every call site across the stack updated.

PR #437, opened 2026-08-16
Vectorize Earth frame transformations

duncaneddy:
# Pull Request

## Description

Vectorizes the Earth frame transformations. Rust gains plural functions for ECI/ECEF, GCRF/ITRF, and EME2000 (`rotations_*`, `positions_*`, `states_*`; 16 functions) built on the batch primitives; the GCRF/ITRF scalar state transforms are split into a private `GcrfItrfContext` + `apply_*` pair so a batch sharing one epoch computes the IAU 2006/2000A matrices once, with results bit-identical to the scalar functions. In Python the existing functions dispatch on input shape: a 1-D vector behaves exactly as before, an array with the components along `axis` (default `-1`) is transformed as a batch keeping its layout, and epoch arguments accept an `Epoch` or a sequence of epochs (one per vector, or one vector broadcast across all epochs). Batches call the core with the GIL released. Adds the shared Python dispatch helpers (`crates/brahe-py/src/batch.rs`), a Learn page, and a standalone example.

### Added
- Batch Earth frame transformations `rotations_/positions_/states_{eci_to_ecef,ecef_to_eci,gcrf_to_itrf,itrf_to_gcrf}` and `positions_/states_{gcrf_to_eme2000,eme2000_to_gcrf}` in Rust; a shared epoch hoists the IAU 2006/2000A rotation matrices across the batch.
- Python `rotation_/position_/state_*` Earth frame functions accept batches of vectors with an `axis` keyword (default `-1`, numpy component-axis convention) and `Epoch | Sequence[Epoch]` epoch arguments; single-vector calls are unchanged.
- Learn page "Vectorized Transformations" and `examples/frames/vectorized_transforms.{py,rs}`.

## Note to Reviewers
The scalar `state_gcrf_to_itrf`/`state_itrf_to_gcrf` bodies move verbatim into `apply_state_*`; the scalar functions become `apply(&context(epc), &x)`. Rust tests assert exact equality of every plural against a loop of the scalar on the hoisted, per-epoch, and broadcast paths; Python tests mirror them and cover `axis`, N-D input, list input, epoch sequence forms, and error cases.

### DIFF
--- a/crates/brahe-py/src/frames.rs
+++ b/crates/brahe-py/src/frames.rs
@@ -226,8 +226,10 @@ fn py_rotation_ecef_to_eci<'py>(py: Python<'py>, epc: &Bound<'py, PyAny>) -> PyR
 ///         one epoch per vector (or broadcasts a single vector across all epochs).
 ///     x (numpy.ndarray or list): Position vector in `GCRF` frame (m), shape `(3,)`, or a batch
 ///         of vectors with the 3 components along `axis` (for example shape `(n, 3)`).
-///     axis (int, optional): Axis of `x` holding the vector components for batched
-///         input. Defaults to `-1` (the last axis).
+///     axis (int, optional): The axis of `x` along which the 3 components of a
+///         single vector lie; the remaining axes enumerate the batch. For a batch of
+///         shape `(n, 3)` the components lie along the last axis, so the default `-1`
+///         applies; a `(3, n)` column layout uses `axis=0`.
 ///
 /// Returns:
 ///     numpy.ndarray: Position vector in `ITRF` frame (m), shape `(3,)` for a single
@@ -248,6 +250,14 @@ fn py_rotation_ecef_to_eci<'py>(py: Python<'py>, epc: &Bound<'py, PyAny>) -> PyR
 ///     # Transform to ITRF
 ///     r_itrf = bh.position_gcrf_to_itrf(epc, r_gcrf)
 ///     print(f"ITRF position: {r_itrf}")
+///
+///     # Batch: one row per position, one shared epoch
+///     positions = np.tile(r_gcrf, (10, 1))                     # shape (10, 3)
+///     positions_itrf = bh.position_gcrf_to_itrf(epc, positions)  # shape (10, 3)
+///
+///     # One position at a sequence of epochs
+///     epochs = [epc + 60.0 * i for i in range(5)]
+///     track = bh.position_gcrf_to_itrf(epochs, r_gcrf)         # shape (5, 3)
 ///     ```
 #[pyfunction]
 #[pyo3(signature = (epc, x, axis=-1))]
@@ -274,8 +284,10 @@ fn py_position_gcrf_to_itrf<'py>(
 ///         one epoch per vector (or broadcasts a single vector across all epochs).
 ///     x (numpy.ndarray or list): Position vector in `ECI` frame (m), shape `(3,)`, or a batch
 ///         of vectors with the 3 components along `axis` (for example shape `(n, 3)`).
-///     axis (int, optional): Axis of `x` holding the vector components for batched
-///         input. Defaults to `-1` (the last axis).
+///     axis (int, optional): The axis of `x` along which the 3 components of a
+///         single vector lie; the remaining axes enumerate the batch. For a batch of
+///         shape `(n, 3)` the components lie along the last axis, so the default `-1`
+///         applies; a `(3, n)` column layout uses `axis=0`.
 ///
 /// Returns:
 ///     numpy.ndarray: Position vector in `ECEF` frame (m), shape `(3,)` for a single
@@ -322,8 +334,10 @@ fn py_position_eci_to_ecef<'py>(
 ///         one epoch per vector (or broadcasts a single vector across all epochs).
 ///     x (numpy.ndarray or list): Position vector in `ITRF` frame (m), shape `(3,)`, or a batch
 ///         of vectors with the 3 components along `axis` (for example shape `(n, 3)`).
-///     axis (int, optional): Axis of `x` holding the vector components for batched
-///         input. Defaults to `-1` (the last axis).
+///     axis (int, optional): The axis of `x` along which the 3 components of a
+///         single vector lie; the remaining axes enumerate the batch. For a batch of
+///         shape `(n, 3)` the components lie along the last axis, so the default `-1`
+///         applies; a `(3, n)` column layout uses `axis=0`.
 ///
 /// Returns:
 ///     numpy.ndarray: Position vector in `GCRF` frame (m), shape `(3,)` for a single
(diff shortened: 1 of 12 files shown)


crates/brahe-py/src/frames.rs
```diff
@@ -409,16 +443,16 @@ fn py_position_ecef_to_eci<'py>(
...
-    x_gcrf: Bound<'py, PyAny>,
-) -> PyResult<Bound<'py, PyArray<f64, Ix1>>> {
-    let vec = frames::state_gcrf_to_itrf(epc.obj, pyany_to_svector::<6>(&x_gcrf)?);
-
-    Ok(vector_to_numpy!(py, vec, 6, f64))
+    epc: &Bound<'py, PyAny>,
+    x_gcrf: &Bound<'py, PyAny>,
+    axis: isize,
+) -> PyResult<Bound<'py, PyAny>> {
+    dispatch_epoch_vec::<6>(py, epc, x_gcrf, axis, frames::state_gcrf_to_itrf, frames::states_gcrf_to_itrf)
```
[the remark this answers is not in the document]
duncaneddy: Intentional, per the approved design for this stack: the Python surface keeps the existing function names and dispatches on input shape (`state_eci_to_ecef(epc, states, axis=-1)` accepts a batch and `Epoch | Sequence[Epoch]`), rather than doubling the namespace with `states_*` twins. The Rust plurals exist because Rust has no dimensional dispatch; every Rust plural is reachable from Python through its scalar-named function, and the Python tests exercise each one. The one exception is `states_koe_osc_to_mean`/`states_koe_mean_to_osc` (#441), which are series-level operations rather than a broadcast of the scalar.

crates/brahe-py/src/frames.rs
```diff
@@ -240,16 +250,16 @@ unsafe fn py_rotation_ecef_to_eci<'py>(py: Python<'py>, epc: &PyEpoch) -> Bound<
 ///     print(f"ITRF position: {r_itrf}")
 ///     ```
```
duncaneddy, line 261: Would be useful to also add a vectorization example to some of these (showing setting the axis)

crates/brahe-py/src/frames.rs
```diff
@@ -218,11 +222,17 @@ unsafe fn py_rotation_ecef_to_eci<'py>(py: Python<'py>, epc: &PyEpoch) -> Bound<
...
 ///
 /// Args:
-///     epc (Epoch): Epoch instant for the transformation
-///     x (numpy.ndarray or list): Position vector in `GCRF` frame (m), shape `(3,)`
+///     epc (Epoch or Sequence[Epoch]): Epoch instant for the transformation. A sequence evaluates
+///         one epoch per vector (or broadcasts a single vector across all epochs).
+///     x (numpy.ndarray or list): Position vector in `GCRF` frame (m), shape `(3,)`, or a batch
+///         of vectors with the 3 components along `axis` (for example shape `(n, 3)`).
+///     axis (int, optional): Axis of `x` holding the vector components for batched
+///         input. Defaults to `-1` (the last axis).
```
duncaneddy: Might want to be a little more clear about what "holding the vector components for batched input" means, does it mean the axis where different vectors we're batching over are contained or the axis which defines a single "element set"/vector which the function gets applied over. eg for an input of position vectors of shape `(100,3)`, would it be (100) or (3). (so -1 returns the 3/last axis)

docs/learn/frames/vectorized.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Two Regimes
+
+The batch speed-up comes from two different sources, and which one applies depends on the epochs.
+
+When all vectors share one epoch, the transformation context (for ECI/ECEF, the bias-precession-nutation, Earth-rotation, and polar-motion matrices) is computed once and applied to every vector. The per-vector work is a handful of matrix products, so the cost of the batch is dominated by the single context evaluation.
+
+When each vector has its own epoch, the context must be evaluated per vector and the batch runs those evaluations across the thread pool. The speed-up over a Python loop is then the sum of avoiding per-call overhead and using multiple cores.
+
+Results are identical to calling the scalar function in a loop in both regimes; the plural functions apply the same expressions in the same order.
```
duncaneddy: Remove section. Unnecessary.

docs/learn/frames/vectorized.md
```diff
@@ -0,0 +1,59 @@
...
+        ```
+        --8<-- "./docs/outputs/frames/vectorized_transforms.rs.txt"
+        ```
+
+## Rust API
+
+The Rust core exposes each batch operation as a plural function alongside its scalar counterpart: `rotation_eci_to_ecef` and `rotations_eci_to_ecef`, `position_eci_to_ecef` and `positions_eci_to_ecef`, `state_eci_to_ecef` and `states_eci_to_ecef`. Plural functions take slices and return a `Vec`. Epoch-dependent functions take `&[Epoch]` as their first argument and follow one broadcast rule: every slice argument has length `1` or the common batch length, and the output has the common length. A length mismatch returns `Err(BraheError::Error)`.
+
+Batches evaluate sequentially for small inputs and on the global thread pool for large ones. The pool size is controlled with `set_num_threads`; see [Threading](../utilities/threading.md).
+
```
duncaneddy: Remove section as well.

examples/frames/vectorized_transforms.py
```diff
@@ -0,0 +1,56 @@
...
+station_ecef = bh.position_geodetic_to_ecef(
+    np.array([-122.4, 37.8, 0.0]), bh.AngleFormat.DEGREES
+)
+station_eci = bh.position_ecef_to_eci(epochs, station_ecef)
+print(f"Station ECI positions shape: {station_eci.shape}")
+for e, r in zip(epochs, station_eci):
+    print(f"  {e}: [{r[0]:.1f}, {r[1]:.1f}, {r[2]:.1f}] m")
+
+# Many epochs, many states: one epoch per row
+states_ecef_series = bh.state_eci_to_ecef(epochs, states_eci[:6])
```
duncaneddy: Shouldn't this just be passed as `bh.state_eci_to_ecef(epochs, states_eci)`? since `states_eci` is `(N, 6)` already and the default should be `axis=-1` so shouldn't it "just work"? as expected with passing in the default so it's easy to use and matches default numpy conventions.

crates/brahe-py/src/frames.rs
duncaneddy: Reworded across all binding docstrings (frames, planetary, coordinates, relative motion, orbits): the axis is the one along which the components of a *single* vector lie — for your `(100, 3)` example that is the length-3 axis, so the default `-1` applies, and a `(3, 100)` column layout uses `axis=0`; the remaining axes enumerate the batch. The docstrings now state exactly this.

crates/brahe-py/src/frames.rs
duncaneddy: Added batch examples to the representative docstrings: `state_eci_to_ecef` shows an `(n, 6)` batch plus the `axis=0` column layout, and `position_gcrf_to_itrf` shows an `(n, 3)` batch plus one position over an epoch sequence.

docs/learn/frames/vectorized.md
duncaneddy: Removed (both the Rust API and Two Regimes sections).

docs/learn/frames/vectorized.md
duncaneddy: Removed.

examples/frames/vectorized_transforms.py
duncaneddy: Agreed — the example now uses matching epoch and state counts so the per-epoch case is exactly `bh.state_eci_to_ecef(epochs, states_eci)` with the default axis; the Rust example passes the full slice the same way.

PR #457, opened 2026-08-28
Resolve Celestrak single-object lookups from the active group

duncaneddy:
# Pull Request

## Description

Celestrak single-object GP lookups (`get_gp_by_catnr`, `get_gp_by_intdes`, `get_gp_by_name`, and any `query_gp` whose only server-side selector is one of `CATNR`, `INTDES`, `NAME`) now resolve from a cached copy of the `active` group instead of issuing one request per object. Resolution order is the exact per-object cache file, then the `active` group (fetched once through the ordinary cached path, so it obeys the client TTL and `BRAHE_NETWORK_MODE`), then a per-object request for objects not in `active` (debris, inactive payloads). `CelestrakQuery::local_selector` identifies eligible queries; `LocalSelector::matches` implements the matching (`catnr` exact, `intdes` case-insensitive equality, `name` case-insensitive substring). Client-side `filter`/`order_by`/`limit` apply after either path; nothing is written to the per-object cache key when `active` answers; group, special, file, sup-GP, SATCAT, `query_raw`, and `download` are unchanged. A `name` search that matches in `active` returns only active objects. A client with a zero cache age sends every query directly, and a failed `active` fetch is not retried for one cache age. Documented in the Rust and Python API docs and in `docs/learn/ephemeris/celestrak.md`.

Second PR of the stack that moves the Celestrak-backed examples to run from a warmed cache on every PR (after #454).

### Changed
- Celestrak single-object lookups (`catnr`, `intdes`, `name`) resolve from a cached copy of the `active` group, fetching it once instead of one request per object; objects not in `active` are still requested individually, and a `name` search that matches in `active` returns only active objects.

### Fixed
- Python `CelestrakClient` methods release the GIL while a request is in flight, so other Python threads can run during Celestrak downloads.

## Note to Reviewers
Examples that search by `name` (UMBRA, CAPELLA, ICEYE, NISAR) will report active-only counts once they run from the cache in PR 4. The 36 pre-existing tests in `src/celestrak/query.rs` were given `#[serial_test::parallel]` labels (attribute only). Three pre-existing single-object client tests now also see an unmatched `GROUP=active` request against their mock server (harmless 404 with a stderr warning); adding an `active` mock to them is a test-body change left for your call. `SpaceTrackClient`'s Python bindings hold the GIL across requests in the same way and are left for a follow-up.

### DIFF
--- a/docs/learn/ephemeris/celestrak.md
+++ b/docs/learn/ephemeris/celestrak.md
@@ -63,18 +63,19 @@ keyword arguments to `get_gp` or as builder methods on `CelestrakQuery.gp`.
 
 A lookup by catalog number, international designator, or name is answered from a cached copy of the
 `active` group whenever the object is in it. The first such lookup downloads `active` once (about
-11,000 records); every later lookup by any client in the same cache reads that file until it passes
-the cache TTL. The order is: the exact per-object cache file if it is present and servable, then the
-cached `active` group, then a request for the object itself. Objects that are not in `active` --
-debris, inactive payloads -- are requested individually, so `get_gp(catnr=...)` works for any catalog
-number.
+11,000 records, several megabytes); every later lookup by any client in the same cache reads that
+file until it passes the cache TTL. The order is: the exact per-object cache file if it is present and
+servable, then the cached `active` group, then a request for the object itself. Objects that are not
+in `active` -- debris, inactive payloads -- are requested individually, so `get_gp(catnr=...)` works
+for any catalog number. `catnr` and `intdes` results are unaffected by which step answers them --
+resolution only changes where the record comes from, not its content.
 
 A name search follows the same order, so a search that matches in `active` returns only active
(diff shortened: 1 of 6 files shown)


docs/learn/ephemeris/celestrak.md
```diff
@@ -57,6 +59,50 @@ Other common lookup patterns include querying by group name (`group="stations"`)
...
+=== "Rust"
+
+    ```rust
+    use brahe::celestrak::CelestrakClient;
+
+    let client = CelestrakClient::new();
+    let iss = client.get_gp_by_catnr(25544).unwrap();
+    let nisar = client.get_gp_by_name("NISAR").unwrap();
+    let debris = client.get_gp_by_catnr(34427).unwrap();
+    ```
```
duncaneddy, line 100: Ideally should be implemented as code examples.

PR #460, opened 2026-08-28
Run CI offline from per-family warmed caches

duncaneddy:
# Pull Request

## Description

CI now runs every test, example, plot, and docs step with `BRAHE_NETWORK_MODE=offline`, reading from caches restored per data family (`naif-kernels-v1-<manifest hash>`, `brahe-icgem-v1-<manifest hash>`, `brahe-sbdb-horizons-v1-<manifest hash>` with `-v1-` prefix fallback so a manifest edit rolls the key; `star-catalogs-v1`; `brahe-celestrak-<run id>`; plus textures/basemap/cartopy). The reusable example workflow takes a `network_mode` input (default `offline`) applied as job-level `env`; warm/download steps override to `online`; the weekly integration workflow passes `online` to the example workflow. The Rust and Python unit-test jobs (`test_rust.yml`, `test_python.yml`) now run offline the same way: loopback URLs (`localhost`, `127.0.0.1`, `::1`) are exempt from `BRAHE_NETWORK_MODE`, so the job-level offline mode no longer breaks the local mock servers their suites use (`tests/plots/test_download.py`'s `http.server` fixture, Rust's httpmock-based tests). Both jobs take a `network_mode` input (default `offline`) applied as job-level `BRAHE_NETWORK_MODE` env. Their caching is centralized in a new `.github/actions/data-cache` composite action (restore or save any subset of the families, keyed from the manifest) and a single `data-cache` job at the front of each workflow: it restores the one family the unit suites read (`naif`; the scratch-cache sweep shows nothing else is needed), and when the family is missing or its cache predates the manifest it builds the extension once, downloads the missing entries with `scripts/warm_data_cache.py --only kernel`, saves them under the current key for later runs (best effort — fork pull requests cannot write the cache), and hands the files to the matrix legs as a run artifact, so the legs never touch the cache or the network themselves. An evicted or stale cache therefore self-heals in one job instead of failing the run or racing a download per leg, and a test whose data is still missing fails with the `BRAHE_NETWORK_MODE` error naming the resource, which points at a manifest gap. `test_examples.yml` uses the same composite for all five families (restore, warm only the misses, save) ahead of its offline example run instead of re-downloading everything with `just download-resources`. Off `main`, saves go under a run-suffixed key that later runs of the same ref find by prefix, so a pull request heals its own cold cache without writing `main`'s keys. `warm_data_cache.yml` is the single weekly producer: it restores, warms, and saves each family independently (`scripts/warm_data_cache.py --only <family>`, `continue-on-error` on each warm step, `!cancelled()` guards on every step so one failing cache action never skips the other families, one failing step at the end if any family failed) and refreshes the Celestrak family (`active`, `starlink`, `gps-ops`, `cosmos-1408-debris`, `fengyun-1c-debris`, `iridium-33-debris`, `cosmos-2251-debris`) under a run-id key that consumers restore by prefix. The three plot-resource caches (textures, Natural Earth basemap, cartopy data) save independently of each other and of the other families, gated only on the warm step's own outcome, so one family's failure never blocks another's save. The monolithic `brahe-data-v1-<manifest hash>` cache and the `cache-warm` job in `integration_tests.yml` are removed. `scripts/warm_data_cache.py` reports an unknown manifest prefix as a failed entry instead of aborting, and the integration jobs restore the star catalogs read-only.

Third PR of the stack (after #454 and #457); PR 4 drops the `NETWORK` flag from the Celestrak examples so they run from this cache on every PR.

### Added
- `scripts/warm_data_cache.py --only <family>` / `--refresh` and the `celestrak:group:<name>` manifest form.

### Changed
- CI test, example, plot, and docs steps all run with `BRAHE_NETWORK_MODE=offline` from per-family restored caches after an online warm step; loopback URLs are exempt from `BRAHE_NETWORK_MODE`, which is what lets the unit-test jobs' local mock servers keep working offline. The weekly `warm_data_cache.yml` warms each family independently, refreshes a run-id-keyed Celestrak cache (`active`, `starlink`, `gps-ops`, and four debris groups) so no PR run contacts Celestrak, and saves the plot-resource caches independently of each other.

## Note to Reviewers
Workflows cannot be exercised locally; this PR's own `pr_tests.yml` run uses the modified reusable workflows and is the live check. The first run after merge may restore no Celestrak cache (no producer has run yet) — the example warm step then downloads the groups once, online; dispatch `warm_data_cache.yml` manually after merging to seed the shared cache.


.github/workflows/test_rust.yml
```diff
@@ -32,13 +38,54 @@ jobs:
...
+                    dir=~/.cache/brahe/naif
+                    mkdir -p "$dir"
+                    cp test_assets/de440s.bsp "$dir/"
+                    warm() {
+                        name=$1; url=$2
+                        if [ -s "$dir/$name" ]; then echo "  $name -> cached"; return; fi
+                        curl -fsSL --retry 3 --retry-delay 5 -o "$dir/$name.tmp" "$url" && mv "$dir/$name.tmp" "$dir/$name" && echo "  $name -> downloaded"
+                    }
+                    warm moon_pa_de440_200625.bpc https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/moon_pa_de440_200625.bpc
+                    warm mar099s.bsp https://naif.jpl.nasa.gov/pub/naif/generic_kernels/spk/satellites/mar099s.bsp
```
duncaneddy: This would appear to run in unit tests and download these files even if not present which seems to be to be inconsistent treatment with how ICGEM, SBDB, and star catalogs are handled which would jsut error if not presented and can be restored.

.github/workflows/test_python.yml
```diff
@@ -63,13 +103,24 @@ jobs:
...
                     cargo llvm-cov clean --workspace
                     eval "$(cargo llvm-cov show-env --export-prefix)"
                     uv pip install maturin --quiet
                     uv run maturin develop --uv --features pyo3/extension-module
+
+            -   name: Warm data cache
+                if: matrix.os == 'ubuntu-latest' && matrix.python-version == inputs.coverage-python-version
+                env:
+                    BRAHE_NETWORK_MODE: online
+                run: uv run --frozen python scripts/warm_data_cache.py --only kernel
```
duncaneddy: Shouldn't warming the cache only occur in integration tests and the in test_python/test_rust

.github/workflows/test_rust.yml
duncaneddy: Agreed — done in 90ad3fff. `test_rust.yml` and `test_python.yml` now only restore the per-family caches (plus the committed `de440s.bsp` test asset) and never download; a test whose data is missing fails with the `BRAHE_NETWORK_MODE` error naming the resource, which surfaces a manifest gap instead of a quiet live fetch. `moon_pa_de440` and `mar099s` are manifest entries so the NAIF cache carries them. Warming is confined to `warm_data_cache.yml` and the online steps of the example/plot/docs jobs. I dispatched the warm job on the stack branches so the caches exist for these PR runs.

.github/workflows/test_python.yml
duncaneddy: Agreed — done in 90ad3fff. `test_rust.yml` and `test_python.yml` now only restore the per-family caches (plus the committed `de440s.bsp` test asset) and never download; a test whose data is missing fails with the `BRAHE_NETWORK_MODE` error naming the resource, which surfaces a manifest gap instead of a quiet live fetch. `moon_pa_de440` and `mar099s` are manifest entries so the NAIF cache carries them. Warming is confined to `warm_data_cache.yml` and the online steps of the example/plot/docs jobs. I dispatched the warm job on the stack branches so the caches exist for these PR runs.

duncaneddy:
Reworked the unit-test caching per the review discussion (commit 17c6be23): a new `.github/actions/data-cache` composite action owns restore/save for every family, and `test_rust.yml`/`test_python.yml` now front their matrix with one `data-cache` job that restores the `naif` family (the only one the unit suites read — the scratch-cache sweep passes with just the three kernels), warms whatever is missing or predates the manifest once, and saves it. The matrix legs only restore, so an evicted or stale cache self-heals in a single job instead of failing the run or racing a download per leg. The example/plot/docs/integration workflows keep their inline blocks for now; moving them onto the composite is a follow-up once this pattern has run.

PR #462, opened 2026-08-28
Run the Celestrak examples from the warmed cache

duncaneddy:
# Pull Request

## Description

The 24 Celestrak-backed documentation examples (`first_script`, `clients_celestrak`, the `datasets/celestrak_*` set, the `common/` access and Starlink examples, and the seven `examples/examples/*.py` studies) no longer carry the `NETWORK` flag, so they run on every PR, main push, and docs build. With `BRAHE_NETWORK_MODE=offline` on the example steps and the `brahe-celestrak-` cache restored, their `catnr`/`name` lookups resolve from the cached `active` group and the `starlink` and `gps-ops` groups are served from disk (#460 warms every Celestrak group the examples and plots use, including `gps-ops` and the four debris groups behind the gabbard documentation plots); their stdout and figures then reach the docs through the ordinary `doc-outputs`/`doc-figures` artifacts instead of the cross-run overlay. The overlay stays for the remaining `NETWORK`-flagged examples (dataset, space-weather, ocean-tide, and canary examples) and its comments describe only those. `examples/test_network.*` keep the flag as the flag-system canaries. The line-numbered snippet includes of the edited examples are realigned to the shorter headers.

`just setup` now runs `download-resources`, so a fresh checkout warms the Celestrak and other network caches once; `test-examples`, `test-example`, `make-plots`, and `make-plot` run from that cache without re-checking each resource on every invocation.

Fourth PR of the stack (after #454, #457, and the CI cache warm PR).

### Changed
- The Celestrak-backed documentation examples run on every PR and docs build from the warmed Celestrak cache, so their output and figures are always published; the `NETWORK` flag now marks only examples whose data cannot be served from the CI caches.
- `just setup` warms the example and plot data caches via `download-resources`, so the example and plot recipes run from a warm cache instead of racing cold downloads.

## Note to Reviewers
A name search that finds no match in the cached `active` group falls through to a per-object request, which is refused under `BRAHE_NETWORK_MODE=offline`; if one of the named satellites (ICEYE, UMBRA, CAPELLA, NISAR) leaves the active catalog, its example fails on every PR until the example is updated, so the weekly warm job's success is worth watching. Examples that search by `name` (UMBRA, CAPELLA, ICEYE, NISAR) report active-only matches now that lookups resolve from the `active` group, so their captured counts may differ from a live server search. The live check is this PR's own CI run: `test-examples` must run (not skip) the 24 examples with no Celestrak requests in the example step, and `doc-outputs` must contain `getting_started/first_script.py.txt`.


justfile
```diff
@@ -98,11 +99,11 @@ download-resources: _setup
     @{{python}} -c "import brahe as bh; c = bh.celestrak.CelestrakClient(cache_max_age=60*86400); c.get_gp(group='active'); c.get_gp(group='starlink'); c.get_gp(group='gps-ops'); c.get_gp(group='cosmos-1408-debris'); c.get_gp(group='fengyun-1c-debris'); c.get_gp(group='iridium-33-debris'); c.get_gp(group='cosmos-2251-debris')"
 
 # Test all documentation examples (delegates to scripts/test_examples.py)
-test-examples *args: _setup
+test-examples *args: _setup download-resources
```
duncaneddy: This would appear to do a download on every single invocation of test-examples/test-plots (even if they are already present) which would make this even slower

duncaneddy: Fair — each downloader fast-paths on a warm cache, but the recipe still launched ~10 interpreters (SBDB, Horizons, star catalogs, ICGEM, Celestrak, textures, cartopy) on every invocation, and anything past its TTL refetched. Changed in 0998b837: the four recipes no longer depend on `download-resources`; `just setup` runs it once for a fresh checkout (it already fetched the kernels), and the CI example/plot/docs jobs keep their explicit warm step. Re-run `just download-resources` after a manifest change or a cleared cache.

PR #464, opened 2026-08-29
Add CCSDS APM message support

duncaneddy:
# Pull Request

## Description

Third PR of the CCSDS ADM/TDM support stack ([#450](https://github.com/duncaneddy/brahe/issues/450)), stacked on [#455](https://github.com/duncaneddy/brahe/pull/455). Adds full CCSDS APM (Attitude Parameter Message, 504.0-B-2 section 3) support: typed message structs with the six repeatable v2 logical blocks (quaternion state with optional derivative, Euler angle state, angular velocity, spin with the three nutation variants, inertia with the Annex F6 sign convention, and torque maneuvers), storing brahe-native values (radians, rad/s, scalar-first `Quaternion`) with wire conversion (degrees, scalar-last `Q1 Q2 Q3 QC`) at the codec boundaries. Message frame endpoints are typed with `ADMReferenceFrame`, the SANA wire vocabulary added by [#453](https://github.com/duncaneddy/brahe/pull/453); the bridges to the native `ReferenceFrame` type live in that PR, so nothing here depends on the frame vocabulary the stack originally carried. KVN uses a delimited-block state machine with unterminated-block detection; XML emits the required `xmlns:xsi` root attribute and escapes free text; JSON round-trips all comment vectors; every writer serializes epochs in the metadata `TIME_SYSTEM` through the shared `format_ccsds_datetime_in` helper, and `CREATION_DATE` in UTC as the ODM writers do. Version 1.0 files are rejected with a descriptive error, at least one logical block is enforced on parse and write, and `USER_DEFINED_*` is rejected as a non-APM construct per 3.2.4.2. Test fixtures are verbatim transcriptions of the standard's Annex G figures G-1, G-2, G-3, and G-10 plus negative fixtures; Python bindings expose `APM` and the six block classes with owned-copy semantics; documentation and paired runnable examples included. The shared CCSDS datetime parser now accepts the optional trailing `Z` terminator. Related follow-up issues filed for pre-existing defects found during this work: [#459](https://github.com/duncaneddy/brahe/issues/459), [#463](https://github.com/duncaneddy/brahe/issues/463).

### Added
- CCSDS APM (Attitude Parameter Message, 504.0-B-2) support: typed message structs with quaternion, Euler-angle, angular-velocity, spin, inertia, and maneuver blocks; KVN, XML, and JSON reading and writing; and Python bindings. The APM XML writer rejects characters XML 1.0 forbids, matching the other XML writers.

### Changed
[...]

Issue #452, opened 2026-08-28
[Feature]: Rotation implementations for remaining orbit-relative frames (LVLH, NTW, TNW, PQW, EQW, SEZ, VNC, NSW)


duncaneddy:
### Description

Brahe currently implements orbit-relative frame rotations only for RTN (`src/relative_motion/eci_rtn.rs`: `rotation_rtn_to_eci`, `rotation_eci_to_rtn`, plus state and batch variants). The CCSDS ADM/TDM support work (#450) introduces typed representations for the full SANA orbit-relative reference frame registry — LVLH, RTN/RSW, NTW, TNW, PQW, EQW, SEZ, VNC, NSW, each in `_ROTATING` and `_INERTIAL` variants — but only as frame metadata. No rotation math is needed for parsing, writing, or trajectory conversion of ADM messages, so the remaining rotation implementations were deferred out of that effort.

This issue tracks implementing the rotation (and state transformation) functions for the remaining orbit-relative frames, modeled on the existing `eci_rtn.rs` pattern:

- LVLH (rotating and inertial/snapshot variants)
- NTW and TNW
- PQW (perifocal; inertial only per SANA)
- EQW (equinoctial; inertial only per SANA)
- SEZ (topocentric south/east/zenith)
- VNC (velocity/normal/co-normal)
- NSW (nadir/sun/normal — note: requires a Sun ephemeris input in addition to the orbit state)

Frame definitions per the SANA orbit-relative reference frames registry (https://sanaregistry.org/r/orbit_relative_reference_frames). `_ROTATING` variants are the true local orbital frames; `_INERTIAL` variants are quasi-inertial snapshots evaluated at each time of interest.

### Use Case

Re-referencing attitude data (AEM/APM) whose `REF_FRAME_A` is an orbit-relative frame into inertial frames, given an orbit state provider; general relative-motion analysis in local orbital frames beyond RTN.

### Impact on Existing API

No impact (new functionality only). New rotation functions in `src/relative_motion/` following the existing naming conventions, with Python bindings, vectorized batch variants, tests, and docs.

Issue #450, opened 2026-08-26
[Feature]: CCSDS AEM support


guija:
### Pre-submission Checklist

- [x] I have searched existing issues to ensure this feature hasn't already been requested
- [x] I have checked the [documentation](https://docs.brahe.space/) to confirm this feature doesn't already exist

### Description

Hi, I'm trying out brahe for a project and it would be very convenient to have CCSDS AEM support:

- AEM parsing
- AEM interpolation.

### Use Case

Using this for a 3D simulation that I am working on.

### Proposed Solution

- Adding new module `brahe::ccsds::aem`
  - Parser
  - AEM struct
- Adding an `AttitudeProvider` trait or similar (similar to Ephemeris, this is how Orekit does it).
- Adding AEM Interpolators (Can existing interpolators be used like for Ephemerides?)

### Impact on Existing API

No impact (new functionality only)

### Contribution

- [x] I can provide test data or validation cases
- [x] I am willing to contribute to the implementation
- [x] I can provide reference material (papers, algorithms, existing implementations)

duncaneddy:
Glad to hear you're trying Brahe out!

Thank you for the feature request. It sounds like a great one. I have some deadlines this week, but should be able to start working on implementing early next week.

Some questions/comments.
1. If you have example AEM files you could share or you want to confirm works, that would be the most helpful thing. I can pull from the Orekit test examples, but always helpful to have specific use cases
1. Do you have any specific methods in mind for 
1. Some existing should be able to be use for attitude interpolation. Generally, you need/should treat attitude interpolation a bit differently since you know that orientations lie on the `SO(3)` manifold (e.g. unit norm, orthogonal in axes when expressed as a rotation matrix). We do already have quaternion spherical linear interpolation (slerp) implemented, so that forms a solid baseline, but might be worth investigating other methods as well.
1. For the `AttitudeProvider` trait do you have any sense of the standard access methods that would be helpful? My immediate thought would be standard interplation accessors for different types of state representation: 
    - `quaterion(epc: Epoch) -> Quaterion`
    - `euler_angle(epc: Epoch, order: EulerAngleOrder) -> EulerAngle`
    - `euler_axis(epc: Epoch) -> EulerAxis`
    - `rotation_matrix(epc: Epoch) -> RotationMatrix`
    - Something for body rates (in body frame)? `body_rates(epoch) -> SVector3`
[...]

PR #471, opened 2026-08-29
Add CCSDS AEM support and attitude trajectories

duncaneddy:
# Pull Request

## Description

Fourth PR of the CCSDS ADM/TDM support stack ([#450](https://github.com/duncaneddy/brahe/issues/450)), stacked on [#464](https://github.com/duncaneddy/brahe/pull/464). Delivers the issue's core ask: parse CCSDS AEM attitude ephemerides and interpolate them. Adds full AEM (Attitude Ephemeris Message, 504.0-B-2 section 4) support with all nine attitude data types, segment metadata with conditional validation enforced identically on parse and write (Euler sequence, angular-velocity frame including the literal `REF_FRAME_A`/`REF_FRAME_B` aliases from Annex G-13, interpolation degree, time coverage, inter-segment useable-time ordering), KVN/XML/JSON codecs each in their own per-message module file, and fixtures transcribed verbatim from Annex G. Adds the native `AttitudeTrajectory` storing canonical states (`AttitudeState` = scalar-first quaternion plus optional body-frame angular velocity with enforced rate uniformity), interpolated by slerp (default), linear, or Lagrange with sequential hemisphere alignment. Its frame endpoints are the unified `ReferenceFrame`, unbound at parse since a CCSDS message names its frames but not the object they belong to. The trajectory implements the merged `OrientationProvider`, which makes it registrable as a frame-graph link: `AEM::register_for(name)` binds the body endpoint to an object, inverts the quaternion series when the celestial endpoint is `REF_FRAME_B`, and registers the attitude as that object's orientation. Interop normalizes every AEM representation to canonical states via the PR 2 kinematics (Euler rates and quaternion derivatives to angular velocity; frame-A angular velocities re-expressed in frame B); spin-parameterized types read and write but error on conversion; the file's `INTERPOLATION_METHOD` is honored with `HERMITE` directing an explicit choice. Python bindings mirror the full surface.

### Added
- CCSDS AEM (Attitude Ephemeris Message, 504.0-B-2) support: segmented attitude ephemerides with all nine attitude data types; KVN, XML, and JSON reading and writing; conversion to and from attitude trajectories; and Python bindings. [@duncaneddy](https://github.com/duncaneddy)
- `AttitudeTrajectory`, relating two `ReferenceFrame` endpoints, with slerp, linear, and Lagrange attitude interpolation and attitude kinematics-based normalization of AEM data on ingest. It implements `OrientationProvider`, so an attitude history can serve as a frame-graph orientation source. [@duncaneddy](https://github.com/duncaneddy)
- `AEM::register_for(name)` registers a single-segment AEM's attitude as the orientation of an object's body frame, binding the message's unbound body endpoint to the object and inverting the series when the celestial endpoint is `REF_FRAME_B`. [@duncaneddy](https://github.com/duncaneddy)

### Changed
- `AttitudeTrajectory` accepts repeated epochs, which represent a discontinuity so an impulsive attitude maneuver can hold its pre- and post-maneuver state at the same instant. A query at a repeated epoch returns the most recently added state, and Lagrange windowing is confined to the run of strictly increasing epochs containing the bracket rather than fitting a polynomial across the jump. This matches the orbit and generic trajectories. [@duncaneddy](https://github.com/duncaneddy)
- `state_providers` is now a module directory with one file per provider family; every trait is re-exported, so existing paths resolve unchanged. [@duncaneddy](https://github.com/duncaneddy)

## Related Issue

#450

## Note to Reviewers

PR 4 of the 5-PR stack (base: #464). Review focus: the interop normalization table in `docs/learn/ccsds/aem.md` (per-type conversion rules and the ANGVEL frame-side re-expression), and `AEM::register_for`'s endpoint-resolution and series-inversion logic in `src/ccsds/interop.rs`, which is the one place the A/B ordering ambiguity turns into a directional decision. The trajectory implements `Trajectory` but not `InterpolatableTrajectory` (that trait's default-method bounds cannot be satisfied by a quaternion state); interpolation is an inherent method.

Two review rulings on this branch reversed earlier ones, so they are worth confirming. Repeated-epoch handling now follows #482's discontinuity semantics rather than the rejection that closed the earlier Codex comment on `from_data`; the NaN that rejection guarded against is instead prevented by returning the stored state at the discontinuity and by confining the Lagrange window. And the branch's own `AttitudeProvider` trait is gone, absorbed into the merged `OrientationProvider`: `angular_velocity` now returns `Ok(None)` for a trajectory carrying no rates instead of erroring, since the merged trait reserves `Err` for real evaluation failures and the graph layer raises the rateless-link error. That also makes `with_numerical_rates` available for opt-in finite-differenced rates. Wire formats are unchanged throughout — the Annex G fixtures are byte-identical to the pre-rebase branch.

One known limitation is filed as [#518](https://github.com/duncaneddy/brahe/issues/518): `register_for` requires one endpoint to resolve to a celestial frame, so an AEM expressed against an orbit-relative frame (e.g. `REF_FRAME_A = LVLH`) parses and converts but cannot be registered. That waits on the orbit-relative rotation derivations in #452. The issue also carries a note about the pre-existing Rust/Python test-count disparity on the attitude trajectory surface, which this PR mirrors for its own new tests but does not otherwise close.


src/trajectories/attitude_trajectory.rs
```diff
@@ -0,0 +1,1815 @@
...
+            ));
+        }
+
+        validate_rate_uniformity(&states)?;
+
+        // Ensure epochs are sorted
+        let mut indices: Vec<usize> = (0..epochs.len()).collect();
+        indices.sort_by(|&i, &j| epochs[i].partial_cmp(&epochs[j]).unwrap());
+
+        let sorted_epochs: Vec<Epoch> = indices.iter().map(|&i| epochs[i]).collect();
```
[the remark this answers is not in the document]
duncaneddy: Fixed in 04b4cffe: `from_data` now rejects duplicate epochs after sorting with the same error style as `add`, so the NaN interval can no longer be constructed.

src/ccsds/interop.rs
```diff
@@ -186,6 +195,344 @@ impl TryFrom<&OEM> for DOrbitTrajectory {
...
+        let mut metadata = AEMMetadata::new(
+            object_name,
+            object_id,
+            ref_frame_a,
+            ref_frame_b.clone(),
+            time_system,
+            start_time,
+            stop_time,
+            attitude_type,
+        );
```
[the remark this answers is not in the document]
duncaneddy: Fixed in 04b4cffe: `from_attitude_trajectory` now maps `Linear` and `Lagrange { degree }` into the segment's `INTERPOLATION_METHOD`/`INTERPOLATION_DEGREE`, with round-trip tests preserving both; `Slerp` correctly writes nothing (no wire token). This also closes #472.

crates/brahe-py/src/ccsds.rs
```diff
@@ -6056,3 +6060,1046 @@ fn pyany_to_optional_array3(
...
+
+            let states = pyo3::types::PyList::empty(py);
+            for state in &segment.states {
+                let st_dict = pyo3::types::PyDict::new(py);
+                st_dict.set_item(
+                    "epoch",
+                    brahe::ccsds::common::format_ccsds_datetime(&state.epoch),
+                )?;
+                st_dict.set_item("attitude_type", format!("{}", state.data.attitude_type()))?;
+                states.append(st_dict)?;
```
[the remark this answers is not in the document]
duncaneddy: Fixed in 04b4cffe: `to_dict()` now emits the variant-specific state payload (quaternion components, derivatives, Euler angles and rates, angular velocity, spin parameters) mirroring the JSON writer's keys and wire conventions.

crates/brahe-py/src/trajectories.rs
```diff
@@ -3965,3 +3965,325 @@ impl PyTrajectoryIterator {
...
+    /// AttitudeFrame: Frame endpoint A
+    #[getter]
+    fn frame_a(&self) -> PyAttitudeFrame {
+        PyAttitudeFrame { frame: self.trajectory.frame_a.clone() }
+    }
+
+    /// AttitudeFrame: Frame endpoint B (stored quaternions rotate from frame_a to frame_b)
+    #[getter]
+    fn frame_b(&self) -> PyAttitudeFrame {
+        PyAttitudeFrame { frame: self.trajectory.frame_b.clone() }
```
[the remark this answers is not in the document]
duncaneddy: Fixed in 04b4cffe: `AttitudeTrajectory` bindings now expose `name` (getter/setter) and a `metadata` dict getter, so the values preserved from AEM segments (object id, center name, useable times) are reachable from Python, with tests.

src/ccsds/xml/parser.rs
```diff
@@ -1992,6 +1996,625 @@ pub fn parse_apm_xml(content: &str) -> Result<APM, BraheError> {
...
+/// Parse an AEM message from XML format.
+pub fn parse_aem_xml(content: &str) -> Result<AEM, BraheError> {
+    let xml_aem: XMLAEM = quick_xml::de::from_str(content)
+        .map_err(|e| ccsds_parse_error("AEM", &format!("XML parse error: {}", e)))?;
+
+    let format_version = xml_aem
+        .version
+        .as_ref()
+        .and_then(|v| v.parse::<f64>().ok())
+        .unwrap_or(2.0);
```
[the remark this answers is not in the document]
duncaneddy: Same ruling as the APM comment: missing-version fallback on XML read is the codebase's established lenient-read behavior (OEM precedent); explicit `1.0` is rejected. Tracked as a repo-wide policy question rather than changed piecemeal in this stack.

src/trajectories/attitude_trajectory.rs
```diff
@@ -0,0 +1,1845 @@
...
+    /// use nalgebra::Vector3;
+    ///
+    /// let state = AttitudeState::new(Quaternion::new(1.0, 0.0, 0.0, 0.0))
+    ///     .with_angular_velocity(Vector3::new(0.0, 0.0, 0.01));
+    /// assert!(state.angular_velocity.is_some());
+    /// ```
+    pub fn with_angular_velocity(mut self, omega: Vector3<f64>) -> Self {
+        self.angular_velocity = Some(omega);
+        self
+    }
```
duncaneddy, line 90: Is constructing extra states with `with_*` instead of having `new` be the omnibus constructor with a specific `from_quarterion(q: Quaternion)` consistent with the Orbit Trajectory implementation?

src/trajectories/attitude_trajectory.rs
```diff
@@ -0,0 +1,1845 @@
...
+}
+
+/// Interpolation method for retrieving an [`AttitudeState`] at an arbitrary epoch.
+#[derive(Debug, Clone, Copy, PartialEq, Default)]
+pub enum AttitudeInterpolationMethod {
+    /// Spherical linear interpolation (slerp) of the bracketing quaternions.
+    /// Exact for constant-angular-rate motion and always produces a unit
+    /// quaternion. Angular velocity, if present, interpolates linearly.
+    #[default]
+    Slerp,
```
duncaneddy, line 110: Is Slerp the better attitude interpolation if we do have angular rates? Is there another better default?

src/trajectories/attitude_trajectory.rs
```diff
@@ -0,0 +1,1845 @@
...
+
+                let quaternion = state1.quaternion.slerp(state2.quaternion, t);
+                let angular_velocity =
+                    lerp_angular_velocity(state1.angular_velocity, state2.angular_velocity, t);
+
+                Ok(AttitudeState {
+                    quaternion,
+                    angular_velocity,
+                })
+            }
```
duncaneddy, line 595: We need to make sure that AttitudeTrajectories similarly support impulsive attitude maneuvers and interpolation works properly around impulsive discontinuities (https://github.com/duncaneddy/brahe/pull/482)

src/utils/state_providers.rs
```diff
@@ -921,6 +926,224 @@ pub trait DOrbitCovarianceProvider: DCovarianceProvider {
     fn covariance_rtn(&self, epoch: Epoch) -> Result<DMatrix<f64>, BraheError>;
 }
 
+// ============================================================================
```
duncaneddy: We might want to break `state_provider.rs` into a module with each different state-provider subtype being its own file

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ Each release groups entries under the Keep a Changelog section headings in the o
...
 - CCSDS APM (Attitude Parameter Message, 504.0-B-2) support: typed message structs with quaternion, Euler-angle, angular-velocity, spin, inertia, and maneuver blocks; KVN, XML, and JSON reading and writing; and Python bindings. [@duncaneddy](https://github.com/duncaneddy) ([#464](https://github.com/duncaneddy/brahe/pull/464))
+- CCSDS AEM (Attitude Ephemeris Message, 504.0-B-2) support: segmented attitude ephemerides with all nine attitude data types; KVN, XML, and JSON reading and writing; conversion to and from attitude trajectories; and Python bindings. [@duncaneddy](https://github.com/duncaneddy) ([#471](https://github.com/duncaneddy/brahe/pull/471))
+- `AttitudeTrajectory` with slerp, linear, and Lagrange attitude interpolation, the `AttitudeProvider` trait with epoch-based attitude accessors, and attitude kinematics-based normalization of AEM data on ingest. [@duncaneddy](https://github.com/duncaneddy) ([#471](https://github.com/duncaneddy/brahe/pull/471))
```
duncaneddy: Remove CHANGELOG commits

src/trajectories/attitude_trajectory.rs
duncaneddy: Checked against `SOrbitTrajectory`/`DOrbitTrajectory`: the convention there is `new(...)` taking the identity arguments, `with_*` builders layering optional configuration, and `from_*` named bulk constructors. `AttitudeTrajectory` already matched on all three — `new(frame_a, frame_b)`, `with_interpolation_method`, and `from_data` — and `AttitudeState::new(quaternion)` + `with_angular_velocity(omega)` is that same shape at state level: the required field in `new`, the optional one layered through a `with_*`. So an omnibus `new` plus a `from_quaternion` shorthand would move away from the orbit convention rather than toward it, and I have left the constructors as they are.

The one place the surfaces genuinely diverged was eviction policy: the orbit trajectories expose `with_eviction_policy_max_size` and `with_eviction_policy_max_age` builders, and `AttitudeTrajectory` only had the in-place setters from the `Trajectory` trait. Both builders are added in 3f9ce94d, so the builder surface now matches.

src/trajectories/attitude_trajectory.rs
duncaneddy: Slerp stands as the default, with the rationale now written into the `AttitudeInterpolationMethod` docstring (3f9ce94d).

The reason it wins as a *default* is that it is the only method here that depends on nothing beyond the two bracketing quaternions: it needs no angular-velocity data, stays on the unit sphere without renormalization, and resolves the double cover to the shorter arc. That means it behaves identically for a trajectory that carries rates and one that does not, which is what a default has to do given rate presence is a property of the source file.

When rates *are* present there is a better-fitting family — attitude analogues of Hermite interpolation, which consume the stored body rates for a C1 fit and are exact for more motions than constant-rate. But they are only available for the rate-carrying subset, so they are additive rather than a replacement default. Not implemented here; worth its own issue if you want it.

src/trajectories/attitude_trajectory.rs
duncaneddy: Ported in 3f9ce94d, following #482's semantics exactly.

Repeated epochs are now accepted rather than rejected: an impulsive attitude maneuver holds its pre- and post-maneuver state at the same instant. A query exactly at a repeated epoch returns the most recently added state — `index_before_epoch` lands on the last of the run and `index_after_epoch` on the first, so the higher index is the newest — which makes the query right-continuous with the states that follow instead of dividing by a zero-length interval and returning NaN.

Interpolation around the discontinuity is handled by dropping this file's private window helper and using the shared `compute_lagrange_window`, which confines the window to the run of strictly increasing epochs containing the bracket. No polynomial can represent a jump, so the window shrinks (lowering the effective degree) rather than fitting across one. That also picks up #482's balanced even-window split.

This supersedes the earlier duplicate-epoch rejection that closed the Codex comment on `from_data`, so I have replaced those two rejection tests with discontinuity tests in Rust and mirrored one in Python. The NaN that rejection was guarding against is now prevented by the two mechanisms above rather than by refusing the input.

src/utils/state_providers.rs
duncaneddy: Done in 3f9ce94d. `src/utils/state_providers.rs` is now `src/utils/state_providers/` with one file per provider family: `state.rs` (frame-agnostic `SStateProvider`/`DStateProvider`), `covariance.rs`, `orbit_state.rs`, `orbit_covariance.rs`, `combined.rs` (the identity-combined traits and `ToPropagatorRefs`), and `attitude.rs` (the frames-router composition on `AttitudeTrajectory`). `mod.rs` re-exports every trait, so `crate::utils::state_providers::X` and the `crate::traits` re-exports resolve exactly as before — no call site changed.

The tests moved with the code they exercise rather than staying in one block, and each file carries its own `mod tests`. Orientation stayed in `frames`, since `OrientationProvider` lives there.

CHANGELOG.md
duncaneddy: Done. Both `CHANGELOG.md` and `docs/changelog.md` are restored to the base branch's versions — `git diff de/ccsds-apm -- CHANGELOG.md docs/changelog.md` is empty at the tip, and neither file appears in this PR's diff. The changelog-only commit that had pointed the AEM entries at the PR link became empty during the rebase and was dropped.

The content now lives in this PR description's `## Changelog` section, updated to post-rework names (`OrientationProvider` rather than `AttitudeProvider`, `ReferenceFrame` endpoints, plus the new `AEM::register_for` and the repeated-epoch behavior change).

src/trajectories/attitude_trajectory.rs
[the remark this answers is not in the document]
duncaneddy: Superseded as of 3f9ce94d, recording it here so the thread is not misleading.

The duplicate-epoch rejection this thread asked for has been removed in favour of #482's discontinuity semantics, per the later review comment on `attitude_trajectory.rs`. Repeated epochs are now accepted and represent an impulsive attitude maneuver holding its pre- and post-maneuver state at the same instant, matching the orbit and generic trajectories.

The NaN this comment correctly identified is still prevented, just by different means: a query exactly at a repeated epoch returns the stored state (the most recently added one, making the query right-continuous) instead of interpolating across a zero-length interval, and Lagrange windowing is confined to the run of strictly increasing epochs containing the bracket so no polynomial is ever fitted across the jump. So the failure mode is closed, while the repeated epochs themselves are now meaningful input rather than an error.

docs/learn/ccsds/aem.md
```diff
@@ -0,0 +1,196 @@
...
+    }).unwrap();
+    segment.push_state(AEMAttitudeState {
+        epoch: t1,
+        data: AEMAttitudeData::Quaternion { quaternion: Quaternion::new(0.9998, 0.0, 0.0, 0.0196) },
+    }).unwrap();
+
+    let mut aem = AEM::new("BRAHE");
+    aem.push_segment(segment);
+    let kvn = aem.to_string(CCSDSFormat::KVN).unwrap();
+    ```
```
duncaneddy, line 100: Example needs to be implemented as code

docs/learn/ccsds/aem.md
```diff
@@ -0,0 +1,196 @@
...
+ATTITUDE_TYPE   = QUATERNION
+INTERPOLATION_METHOD = hermite
+INTERPOLATION_DEGREE = 7
+META_STOP
+
+DATA_START
+1996-11-28T21:29:07.2555 0.56748  0.03146  0.45689  0.68427
+1996-11-30T01:28:02.5555 0.74563  -0.45375  0.36875  0.31964
+DATA_STOP
+```
```
duncaneddy, line 122: Might be best implemented as a data file that's imported

docs/learn/trajectories/attitude_trajectory.md
```diff
@@ -0,0 +1,138 @@
...
+    use brahe::trajectories::{AttitudeState, AttitudeTrajectory};
+
+    let mut traj = AttitudeTrajectory::new(
+        ReferenceFrame::from(CelestialFrame::GCRF),
+        ReferenceFrame::from(BodyFrame::SCBody(Some("1".to_string()))),
+    );
+    let epoch = Epoch::from_datetime(2024, 1, 1, 0, 0, 0.0, 0.0, TimeSystem::UTC);
+    traj.add(epoch, AttitudeState::new(Quaternion::new(1.0, 0.0, 0.0, 0.0))).unwrap();
+    assert!(!traj.has_rates());
+    ```
```
duncaneddy, line 21: Should be implemented as code

docs/learn/trajectories/attitude_trajectory.md
```diff
@@ -0,0 +1,138 @@
...
+
+=== "Rust"
+    ``` rust
+    use brahe::attitude::EulerAngleOrder;
+    use brahe::frames::OrientationProvider;
+
+    let q = traj.quaternion(epoch).unwrap();
+    let angles = traj.euler_angle(epoch, EulerAngleOrder::ZXZ).unwrap();
+    ```
```
duncaneddy, line 90: Should be implemented as code

docs/learn/trajectories/attitude_trajectory.md
```diff
@@ -0,0 +1,138 @@
...
+
+## frame_a / frame_b Semantics
+
+`frame_a` and `frame_b` are [`ReferenceFrame`](../frames/index.md) values — each one either a celestial frame, an orbit-relative frame, or a body frame. AEM endpoints parse as unbound frames: the message names the frame but not the object it belongs to. Every stored quaternion rotates from `frame_a` to `frame_b`, mirroring the `REF_FRAME_A`/`REF_FRAME_B` pair on an AEM segment. Which frame is inertial and which is body-fixed is not fixed by the trajectory type itself — some sources put the spacecraft frame first, others put it second — so `frame_a`/`frame_b` are the only reliable way to tell which is which for a given trajectory.
```
duncaneddy: Overly detailed comment, just keep to essential explanation:

`frame_a` and `frame_b` are [`ReferenceFrame`](../frames/index.md) values — each one either a celestial frame, an orbit-relative frame, or a body frame. AEM endpoints parse as unbound frames: the message names the frame but not the object it belongs to. Every stored quaternion rotates from `frame_a` to `frame_b`.

docs/learn/trajectories/attitude_trajectory.md
```diff
@@ -0,0 +1,138 @@
...
+    ```
+
+=== "Rust"
+    ``` rust
+    use brahe::frames::ReferenceFrame;
+
+    let q = traj.quaternion_from_frame(epoch, ReferenceFrame::EME2000).unwrap();
+    ```
+
+Because brahe's Hamilton product `x * y` applies `x` first, composing "from $\to$ frame_a" then "frame_a $\to$ frame_b" is written `q_from_to_a * q_a_to_b`, not the reverse; `quaternion_from_frame` handles this composition internally. It errors if `frame_a` is not a reference frame, if the frame-router transformation fails, or if the attitude at `epoch` cannot be computed.
```
duncaneddy: Overly detailed explanation on implementation (e.g. the Hamilton product) Remove and just state the error conditions

docs/learn/trajectories/attitude_trajectory.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Composing with a Different Reference Frame
+
+`quaternion_from_frame(epoch, from)` re-expresses the trajectory's attitude relative to an arbitrary reference frame, when `frame_a` is itself a celestial frame (constructed via `ReferenceFrame.celestial`/`ReferenceFrame::from(CelestialFrame::…)`). Internally, this composes the frame-router rotation from `from` to `frame_a`'s reference frame with the trajectory's own stored rotation from `frame_a` to `frame_b`. `frame_a` must be a celestial frame for this to succeed, so the call below assumes a trajectory built as in the [rate uniformity example](#canonical-state-and-rate-uniformity) above, where `frame_a` is `GCRF`:
```
duncaneddy: This is a good explanation to have but overly detailed and circular (saying a lot of words while saying nothing) in discussing frames.

Can we condense and explain is `quaternion_from_frame` changing the destination frame (frame_b) or the source frame (frame_a)?

These attitude helps (`*_from_frame`) might also be overly complicated so better to remove entirely until we're able to think through the design YAGNI-principle (which is actually what I'd choose first)

docs/learn/trajectories/attitude_trajectory.md
```diff
@@ -0,0 +1,138 @@
...
+    let epoch = Epoch::from_datetime(2024, 1, 1, 0, 0, 0.0, 0.0, TimeSystem::UTC);
+    traj.add(epoch, AttitudeState::new(Quaternion::new(1.0, 0.0, 0.0, 0.0))).unwrap();
+    assert!(!traj.has_rates());
+    ```
+
+## Interpolation
+
+`interpolate` (Rust) or the [`OrientationProvider`](#orientationprovider) accessors (Rust and Python) retrieve the attitude at an arbitrary epoch within the trajectory's span, using the configured `AttitudeInterpolationMethod`:
+
+- **`Slerp`** (default): spherical linear interpolation of the bracketing quaternions. Exact for constant-angular-rate motion, and always produces a unit quaternion. This is the same interpolation family used for attitude in most spacecraft dynamics software, and is brahe's chosen default for an AEM segment whose `INTERPOLATION_METHOD` is not set, since the standard itself does not mandate a method in that case.
```
duncaneddy: Link to SLERP's wikipedia page somewhere in this (https://en.wikipedia.org/wiki/Spherical_linear_interpolation)

Issue #463, opened 2026-08-29
[Bug]: ODM writer conformance gaps (header order, epoch time system, XML escaping, JSON comments)


duncaneddy:
### Description

Four writer-side conformance/fidelity gaps exist in the pre-existing OEM/OMM/OPM (and partly CDM) codecs. Each was found during the APM implementation (#450 stack), where the corresponding APM code was fixed on branch `de/ccsds-apm`; the older writers were deliberately left unchanged for surgical scope:

1. **Header keyword order**: KVN writers emit `CLASSIFICATION` before header `COMMENT` lines (`src/ccsds/kvn/writer.rs:19-28` and siblings), but the standards fix the order as VERS then COMMENT then CLASSIFICATION. On reparse, header comments after CLASSIFICATION can be misrouted into metadata comments.
2. **Epoch time system**: writers serialize each `Epoch` in its own time system via `format_ccsds_datetime` without converting to the message metadata's `TIME_SYSTEM`, so a user-constructed message with mixed systems writes inconsistent epoch strings. The APM writers now convert; OEM/OMM/OPM/CDM should follow.
3. **XML text escaping**: XML writers interpolate free text (comments, originator, object names) without escaping `&`, `<`, `>`, producing malformed XML for legal values. An `escape_xml_text` helper now exists in `src/ccsds/xml/writer.rs` (used by APM) and can be applied to the other writers.
4. **JSON comment drop**: OPM-pattern JSON writers omit header/metadata/block comment arrays, silently losing comments on JSON round-trips. The APM JSON codec now round-trips all comment vectors and its field-match test helper compares them.

Also related: #459 (multi-COMMENT XML parse crash in the same legacy parsers).

### Language

Both

### Impact

Standards-conformance and data-fidelity defects on write paths for OEM/OMM/OPM/CDM; parse paths mostly unaffected.

duncaneddy:
One more writer-side gap in the same family, found during AEM hardening (#471): writers serialize epochs via a fallback (unconverted, as stored) when the metadata TIME_SYSTEM has no brahe mapping (MET/MRT/SCLK/GMST/TDR), but `parse_ccsds_datetime` rejects those systems — so a message can be written that its own parser cannot read. AEM now rejects such metadata in `validate_for_write`; the APM writers (and the older ODM writers) still carry the silent fallback and should adopt the same write-side rejection.

duncaneddy:
[...]

PR #478, opened 2026-08-30
Write ODM epochs in the message's declared time system

duncaneddy:
# Pull Request

## Description

CCSDS 502.0-B-3 subsection 7.5.11 fixes `CREATION_DATE` to UTC and ties every other time or epoch keyword to the message's `TIME_SYSTEM`; CCSDS 508.0-B-1 subsection 6.2.3.4 fixes every CDM time tag to UTC. A brahe `Epoch` carries its own time system, which for a user-constructed message need not be the one the metadata declares, and the writers formatted each epoch as stored. A message holding TAI epochs under `TIME_SYSTEM = UTC` was written with UTC in the metadata and TAI in the data — an error of the current leap-second count, silently produced and silently misread.

A new `format_ccsds_datetime_in` converts to the declared system before formatting, and the OEM, OMM, OPM, and CDM writers use it in all three encodings, along with the OMM-to-`GPRecord` conversion and the Python `to_dict` views so they report the same strings the writers emit. The five CCSDS time systems with no `TimeSystem` counterpart — `MET`, `MRT`, `SCLK`, `GMST`, `TDR` — are mission- or spacecraft-specific clocks with no fixed relationship to a physical time scale, so an epoch declared in one of those is written as stored.

### Fixed
- CCSDS OEM, OMM, OPM, and CDM writers now convert each epoch to the message's declared `TIME_SYSTEM` before formatting, per CCSDS 502.0-B-3 subsection 7.5.11 and CCSDS 508.0-B-1 subsection 6.2.3.4, instead of writing it in whichever time system the `Epoch` happened to carry. `CREATION_DATE` and all CDM time tags are written in UTC.

## Related Issue

Covers item 2 of #463.

## Note to Reviewers

`codecov/patch` initially failed here for the same reason as the PR below it: the sweep replaces the formatter at 81 call sites, many on epoch fields no stock fixture populates, so pre-existing gaps became patch misses. Rather than add a coverage-shaped test, the existing retag test was extended to populate every epoch field the writers can emit — `REF_FRAME_EPOCH` on the OMM and OPM, the covariance epoch, and the CDM message-linking epochs — and moved to an OPM fixture that actually carries maneuvers. Its claim is that retagging *any* epoch leaves output identical, so it should have covered those all along. Patch coverage went from 95.7% to 99.6%; the two lines still uncovered are arguments to `assert!` failure messages.

Tests exercise the helper directly, assert that retagging every epoch of an OEM, OMM, OPM, and CDM leaves KVN, XML, and JSON output byte identical, and assert that redeclaring `TIME_SYSTEM` moves the written epochs onto that scale. All three were confirmed to fail against the previous behaviour. Messages parsed from a file were already consistent, since the parsers build epochs in the declared system; the defect only bites user-constructed or programmatically retagged messages.

### DIFF
--- a/src/ccsds/common.rs
+++ b/src/ccsds/common.rs
@@ -534,6 +534,45 @@ pub fn format_ccsds_datetime(epoch: &Epoch) -> String {
     }
 }
 
+/// Format an `Epoch` as a CCSDS datetime string in a given CCSDS time system.
+///
+/// CCSDS 502.0-B-3 subsection 7.5.11 ties every time or epoch keyword other
+/// than `CREATION_DATE` to the message's `TIME_SYSTEM`, and CCSDS 508.0-B-1
+/// subsection 6.2.3.4 fixes every CDM time tag to UTC. An `Epoch` carries its
+/// own time system, which need not be the one the message declares, so writers
+/// convert before formatting.
+///
+/// Five CCSDS time systems — `MET`, `MRT`, `SCLK`, `GMST`, and `TDR` — are
+/// mission- or spacecraft-specific clocks with no fixed relationship to the
+/// physical time systems `Epoch` represents. For those the epoch is formatted
+/// as stored, unconverted.
+///
+/// # Arguments
+///
+/// * `epoch` - The epoch to format.
+/// * `time_system` - The CCSDS time system the message declares.
+///
+/// # Returns
+///
+/// * `String` - The epoch rendered as `YYYY-MM-DDThh:mm:ss.sss`.
+///
+/// # Examples
+///
+/// ```
+/// use brahe::ccsds::common::{CCSDSTimeSystem, format_ccsds_datetime_in};
+/// use brahe::time::{Epoch, TimeSystem};
+///
+/// let epoch = Epoch::from_datetime(2024, 3, 1, 12, 0, 0.0, 0.0, TimeSystem::TAI);
+/// let written = format_ccsds_datetime_in(&epoch, &CCSDSTimeSystem::UTC);
+/// assert!(written.starts_with("2024-03-01T11:59:"));
+/// ```
+pub fn format_ccsds_datetime_in(epoch: &Epoch, time_system: &CCSDSTimeSystem) -> String {
+    match time_system.to_time_system() {
+        Some(ts) => format_ccsds_datetime(&epoch.to_time_system(ts)),
+        None => format_ccsds_datetime(epoch),
+    }
+}
+
 /// Strip unit annotations from a CCSDS KVN value string.
 ///
 /// CCSDS KVN values may contain optional unit annotations in square brackets
@@ -713,6 +752,10 @@ pub fn covariance_to_lower_triangular(matrix: &SMatrix<f64, 6, 6>, scale: f64) -
 #[cfg_attr(coverage_nightly, coverage(off))]
 mod tests {
     use super::*;
+    use crate::ccsds::cdm::CDM;
+    use crate::ccsds::oem::OEM;
+    use crate::ccsds::omm::OMM;
+    use crate::ccsds::opm::OPM;
 
     #[test]
     fn test_ccsds_format_display() {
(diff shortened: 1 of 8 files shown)


src/ccsds/kvn/writer.rs
```diff
@@ -46,29 +48,29 @@ pub fn write_oem(oem: &OEM) -> Result<String, BraheError> {
         if let Some(ref epoch) = segment.metadata.ref_frame_epoch {
             out.push_str(&format!(
                 "REF_FRAME_EPOCH = {}\n",
-                format_ccsds_datetime(epoch)
+                format_ccsds_datetime_in(epoch, &segment.metadata.time_system)
```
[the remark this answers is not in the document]
duncaneddy: Confirmed and fixed. Measured a 30 s shift on OEMExample1 with `TIME_SYSTEM = TAI`: KVN and JSON both moved `REF_FRAME_EPOCH` by the full TAI-UTC offset. The raw value is now kept and parsed once the metadata is complete, in all three KVN parsers.

The same investigation turned up a second instance you did not flag: the JSON readers flatten a metadata object in whatever order its keys arrive, which puts `START_TIME` ahead of `TIME_SYSTEM` and shifted it identically. `TIME_SYSTEM` is now emitted first when flattening. `test_ref_frame_epoch_survives_a_non_utc_time_system` covers both fields across KVN, XML, and JSON.

tests/ccsds/test_time_system.py
```diff
@@ -0,0 +1,56 @@
...
+    assert opm.time_system == "UTC"
+    utc = opm.to_string("kvn")
+
+    opm.time_system = "TAI"
+    assert opm.to_string("kvn") != utc
+
+
+def test_writers_use_declared_time_system_not_epoch_time_system(eop):
+    """Mirror of test_writers_use_declared_time_system_not_epoch_time_system in Rust."""
+    oem = OEM.from_file("test_assets/ccsds/oem/OEMExample1.txt")
```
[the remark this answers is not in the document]
duncaneddy: Added. `test_cdm_writes_time_tags_in_utc` covers the CDM across all three encodings and asserts `PyCDM::__repr__` reports the same instant. It compares written time codes rather than `Epoch` values, because an `Epoch` carries sub-nanosecond representation noise that a CCSDS time code does not encode.

src/ccsds/kvn/parser.rs
```diff
@@ -111,7 +111,11 @@ pub fn parse_oem(content: &str) -> Result<OEM, BraheError> {
     let mut meta_object_id: Option<String> = None;
     let mut meta_center_name: Option<String> = None;
     let mut meta_ref_frame: Option<CCSDSRefFrame> = None;
-    let mut meta_ref_frame_epoch: Option<Epoch> = None;
+    // CCSDS 502.0-B-3 subsection 7.5.11 puts REF_FRAME_EPOCH in the message's
+    // TIME_SYSTEM, but the fixed keyword order of tables 3-2, 4-2, and 5-3
+    // places it before TIME_SYSTEM, so the scale is not known when the line is
+    // read. The raw value is kept and parsed once the metadata is complete.
```
duncaneddy: Remove comment

src/ccsds/kvn/parser.rs
```diff
@@ -475,7 +484,11 @@ pub fn parse_omm(content: &str) -> Result<OMM, BraheError> {
     let mut object_id: Option<String> = None;
     let mut center_name: Option<String> = None;
     let mut ref_frame: Option<CCSDSRefFrame> = None;
-    let mut ref_frame_epoch: Option<Epoch> = None;
+    // CCSDS 502.0-B-3 subsection 7.5.11 puts REF_FRAME_EPOCH in the message's
+    // TIME_SYSTEM, but the fixed keyword order of tables 3-2, 4-2, and 5-3
+    // places it before TIME_SYSTEM, so the scale is not known when the line is
+    // read. The raw value is kept and parsed once the metadata is complete.
```
duncaneddy: Remove comment, too verbose given other implementation doesn't have similar comments.

src/ccsds/kvn/parser.rs
```diff
@@ -900,7 +912,11 @@ pub fn parse_opm(content: &str) -> Result<OPM, BraheError> {
...
```
duncaneddy: Remove comment, too verbose

src/ccsds/common.rs
```diff
@@ -1265,4 +1304,245 @@ mod tests {
...
+        }
+    }
+
+    #[test]
+    #[serial_test::parallel]
+    fn test_writers_use_declared_time_system_not_epoch_time_system() {
+        use crate::ccsds::cdm::CDM;
+        use crate::ccsds::oem::OEM;
+        use crate::ccsds::omm::OMM;
+        use crate::ccsds::opm::OPM;
```
duncaneddy: All use statements should be at top of `mod tests{` block

src/ccsds/common.rs
```diff
@@ -1265,4 +1304,245 @@ mod tests {
...
+                .unwrap()
+                .contains("START_TIME = 1996-12-18T12:00:00.331")
+        );
+
+        // Redeclaring TIME_SYSTEM moves the written epochs onto that scale.
+        oem.segments[0].metadata.time_system = CCSDSTimeSystem::TAI;
+        assert!(
+            oem.to_string(CCSDSFormat::KVN)
+                .unwrap()
+                .contains("START_TIME = 1996-12-18T12:00:30.331")
```
duncaneddy, line 1545: This doesn't appear to be testing what the comment states it is. Shouldn't it be confirming the time system field is actually TAI, not that the start time remains the same.

src/ccsds/kvn/parser.rs
duncaneddy: Removed. It was the same five-line note repeated at all three parsers.

src/ccsds/kvn/parser.rs
duncaneddy: Removed. It was the same five-line note repeated at all three parsers.

src/ccsds/kvn/parser.rs
duncaneddy: Removed. It was the same five-line note repeated at all three parsers.

src/ccsds/common.rs
duncaneddy: Moved. `CDM`, `OEM`, `OMM`, and `OPM` are now imported once at the top of `mod tests` instead of inside individual test functions.

src/ccsds/common.rs
duncaneddy: You are right, it did not check that. The assertions only showed the instants were preserved, which a writer that dropped `TIME_SYSTEM` and wrote UTC epochs would also satisfy.

Added `assert_eq!(reparsed.segments[0].metadata.time_system, CCSDSTimeSystem::TAI)` inside the per-format loop, so the declaration itself has to round-trip before the epoch assertions mean anything. Mirrored in the Python test. The doc comment now states what is actually checked: the message comes back declaring the same system, with every epoch on it.

PR #480, opened 2026-08-30
Attribute KVN comments to the section they introduce

### LINKED ISSUE #475: [Bug]: CDM KVN comments are attributed to the wrong section and do not round-trip
### Description

The CDM KVN codec attributes `COMMENT` lines to the wrong buckets. There are two layers, and they compound.

**1. Parse-side misattribution.** `parse_cdm` in `src/ccsds/kvn/parser.rs` files each comment under whichever section is currently active, and only switches section when it reads the *following* keyword line. CCSDS CDMs conventionally place a section-introducing comment immediately before that section's first keyword, so every such comment is filed under the *preceding* section. This is visible on the very first parse, before any writing happens.

**2. Round-trip instability.** `write_cdm` in `src/ccsds/kvn/writer.rs` emits `relative_metadata.comments` after the header block but *before* the `TCA` line, while `parse_cdm` classifies every comment seen before `TCA` as a header comment (the `CurrentObject::None` arm). Relative-metadata comments therefore migrate into `header.comments` on re-parse, and `write(parse(write(x))) != write(x)`.

`CDMExample2.txt`, `CDMExample3.txt`, `CDMExample4.txt`, and `CDMExample_issue_940.txt` all carry a `COMMENT` ahead of `TCA` and exercise this.

Before fixing, the intended placement should be checked against CCSDS 508.0-P-1.1 table 3-2 and section 6.3.1.9. Note that the CDM's `COMMENT` keyword, unlike the ODM's, carries no "only immediately after the version line" restriction, so section-leading comments are legal and the parser is what needs to attribute them forward. The two layers may share a single fix.

Related to #463, which tracks the other CCSDS writer conformance and comment-fidelity gaps found in the same area, though this is not one of that issue's four listed items: those are writer-side conformance gaps, whereas the root cause here is parser-side comment attribution. Also distinct from #459, which is an XML-only `duplicate field COMMENT` deserialization failure in the OEM/OMM/OPM intermediate structs.

### Language

Both

### Rust Reproduction Code

```rust
use brahe::ccsds::{CCSDSFormat, CDM};

fn main() {
    let src = std::fs::read_to_string("test_assets/ccsds/cdm/CDMExample2.txt").unwrap();
    let first = CDM::from_str(&src).unwrap();

    // Layer 1: every section-introducing comment is filed one section early.
    println!("header:   {:?}", first.header.comments);
    println!("rel meta: {:?}", first.relative_metadata.comments);
    println!("obj1data: {:?}", first.object1.data.comments);

    // Layer 2: the round trip does not converge.
    let written = first.to_string(CCSDSFormat::KVN).unwrap();
    let rewritten = CDM::from_str(&written).unwrap().to_string(CCSDSFormat::KVN).unwrap();
    println!("stable: {}", rewritten == written);
}
```

### Python Reproduction Code

```python
from brahe.ccsds import CDM

cdm = CDM.from_file("test_assets/ccsds/cdm/CDMExample2.txt")
written = cdm.to_string("kvn")
again = CDM.from_str(written).to_string("kvn")
print("stable:", again == written)
```

### Expected Behavior

Each `COMMENT` is attributed to the section it introduces, and writing then re-parsing a CDM preserves the bucket every comment sits in, so `write(parse(write(x))) == write(x)`.

### Actual Behavior

Parsing `CDMExample2.txt` files each section-introducing comment under the preceding section:

```
header.comments            = ["Relative Metadata/Data"]
relative_metadata.comments = ["Object1 Metadata"]
object1.data.comments      = [..., "Object2 Metadata"]
```

After a write and re-parse, `relative_metadata.comments` empties and its content moves into `header.comments`:

```
header.comments            = ["Relative Metadata/Data", "Object1 Metadata"]
relative_metadata.comments = []
```

The Python and Rust reproductions both print `stable: false`.

### Brahe Version

1.7.0

### Operating System

macOS

### Additional Context

Found while fixing the header `COMMENT`/`CLASSIFICATION` ordering in #474; unrelated to that change and present independently of it.

duncaneddy:
# Pull Request

## Description

CCSDS 508.0-B-1 tables 3-1 through 3-4 give the header, the relative metadata/data section, the object metadata, and each logical block of the object data its own leading `COMMENT` row, and the standard's own example places every section-introducing comment immediately before that section's first keyword. The parser instead filed each comment under whichever section was already open, so every one landed a section early: `Relative Metadata/Data` became a header comment, `Object1 Metadata` a relative-metadata comment, and the whole of object 1's block comments piled into its data comments. The same defect applied to the OMM and OPM KVN parsers.

Comments are now buffered until a keyword names the block they introduce, with a keyword table mapping every keyword to its block. The shared covariance keywords take the same open-matrix context the value dispatch uses, and `OBJECT` resolves its own object from its value, so `COMMENT Object2 Metadata` reaches object 2 rather than object 1. A comment run before an unrecognized keyword stays buffered rather than being filed under a guess. On the write side, the CDM emitted data-section comments just before the state vector; table 3-4 lists the Data `COMMENT` row ahead of the OD Parameters sub-block, so they now lead the data section.

### Fixed
- CCSDS CDM, OMM, and OPM KVN parsers now attribute each `COMMENT` to the section it introduces rather than to the preceding one, so a comment ahead of `TCA` is a relative-metadata comment and one ahead of `OBJECT` is that object's metadata comment.
- CCSDS CDM KVN round trips are now stable: `write(parse(write(x))) == write(x)` holds for `CDMExample1` through `CDMExample4` and `CDMExample_issue_940`, which previously migrated relative-metadata comments into the header on every cycle.

## Related Issue

Closes #475.

## Note to Reviewers

KVN puts nothing between a Data comment and the OD Parameters comment that follows it, so a run spanning both is attributed to the innermost block the next keyword opens — `["Object1 Data", "Object1 OD Parameters"]` both land in `od_parameters.comments`. Content and order are preserved and the round trip converges; only the bucket boundary, which KVN does not encode, is lost. This was the agreed reading over splitting heuristically or hoisting to the outer block.

Worth checking independently: the keyword-to-block table is a second list alongside the value dispatch and could drift from it. `test_comment_target_classifies_every_block` covers one keyword per block plus the shared-covariance context, but a keyword added to the dispatch without a table entry would silently leave its comments unattributed rather than fail.


src/ccsds/kvn/parser.rs
```diff
@@ -879,13 +944,71 @@ fn parse_scientific_notation(s: &str) -> Result<f64, BraheError> {
...
+fn opm_section(key: &str) -> Option<OpmSection> {
+    Some(match key {
+        "CCSDS_OPM_VERS" | "CLASSIFICATION" | "CREATION_DATE" | "ORIGINATOR" | "MESSAGE_ID" => {
+            OpmSection::Header
+        }
+
+        "OBJECT_NAME" | "OBJECT_ID" | "CENTER_NAME" | "REF_FRAME" | "REF_FRAME_EPOCH"
+        | "TIME_SYSTEM" => OpmSection::Metadata,
+
+        "EPOCH" | "X" | "Y" | "Z" | "X_DOT" | "Y_DOT" | "Z_DOT" => OpmSection::StateVector,
```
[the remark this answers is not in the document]
duncaneddy: Fixed, though not reachable from our own writer any more — #484 stops the OPM and OMM writers emitting a covariance `EPOCH` at all, since CCSDS 502.0-B-3 tables 3-3 and 4-3 do not define one. It is still reachable from a third-party file, so the classification is worth getting right.

Rather than making `EPOCH` section-aware, both `omm_section` and `opm_section` now return `None` for it. `EPOCH` genuinely names more than one block, so leaving it unclassified holds any pending comments until a keyword that names exactly one arrives — `X` for the state vector, `COV_REF_FRAME` for the covariance. That needs no extra context threaded through.

src/ccsds/kvn/parser.rs
```diff
@@ -450,6 +450,73 @@ pub fn parse_oem(content: &str) -> Result<OEM, BraheError> {
...
+/// and 4-3 (data) list them.
+///
+/// # Arguments
+///
+/// * `key` - The KVN keyword, without surrounding whitespace.
+///
+/// # Returns
+///
+/// * `Option<OmmSection>` - The owning section, or `None` if unrecognized.
+fn omm_section(key: &str) -> Option<OmmSection> {
```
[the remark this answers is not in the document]
duncaneddy: Added `# Examples` to both `omm_section` and `opm_section`, including the `EPOCH` case now that it returns `None`.

src/ccsds/kvn/parser.rs
```diff
@@ -459,6 +459,84 @@ pub fn parse_oem(content: &str) -> Result<OEM, BraheError> {
     Ok(OEM { header, segments })
 }
 
+/// The OMM section a keyword belongs to.
+#[derive(Clone, Copy, PartialEq)]
+enum OmmSection {
```
duncaneddy: Should be `OMMSection`

src/ccsds/kvn/parser.rs
```diff
@@ -891,13 +967,87 @@ fn parse_scientific_notation(s: &str) -> Result<f64, BraheError> {
         .map_err(|_| ccsds_parse_error("OMM", &format!("invalid numeric value '{}'", s)))
 }
 
+/// The OPM section a keyword belongs to.
+#[derive(Clone, Copy, PartialEq)]
+enum OpmSection {
```
duncaneddy: Should be `OPMSection`

src/ccsds/kvn/parser.rs
```diff
@@ -1361,6 +1539,28 @@ pub fn parse_opm(content: &str) -> Result<OPM, BraheError> {
     })
 }
 
+/// The logical block a CDM keyword belongs to, per the CCSDS 508.0-B-1 tables.
+/// Table 3-4 gives the object Data section a COMMENT row of its own and then a
+/// further COMMENT row for each of its sub-blocks.
+#[derive(Clone, Copy, PartialEq)]
+enum CdmBlock {
```
duncaneddy: Should be `CDMBlock`

src/ccsds/kvn/parser.rs
```diff
@@ -1361,6 +1539,28 @@ pub fn parse_opm(content: &str) -> Result<OPM, BraheError> {
...
+    Metadata,
+    OdParameters,
```
duncaneddy: Should be `ODParameters`

src/ccsds/kvn/parser.rs
```diff
@@ -1361,6 +1539,28 @@ pub fn parse_opm(content: &str) -> Result<OPM, BraheError> {
...
+    AdditionalParameters,
+    StateVector,
+    RtnCovariance,
```
duncaneddy: Should be `RTNCovariance`

src/ccsds/kvn/parser.rs
```diff
@@ -1361,6 +1539,28 @@ pub fn parse_opm(content: &str) -> Result<OPM, BraheError> {
...
+    XyzCovariance,
```
duncaneddy: Should be XYZCovariance

src/ccsds/kvn/parser.rs
```diff
@@ -3058,4 +3470,205 @@ mod tests {
         assert_eq!(opm.metadata.time_system, CCSDSTimeSystem::GPS);
         assert_eq!(opm.maneuvers.len(), 3);
     }
+
+    /// An OPM whose every logical block carries a leading COMMENT, the shape
+    /// CCSDS 502.0-B-3 table 3-3 defines.
+    const OPM_KVN_SECTION_COMMENTS: &str = "\
```
duncaneddy: Should make test asset file

src/ccsds/kvn/parser.rs
```diff
@@ -3058,4 +3470,205 @@ mod tests {
...
+MAN_DELTA_MASS = -2.5
+MAN_REF_FRAME = RTN
+MAN_DV_1 = 0.0
+MAN_DV_2 = 0.002
+MAN_DV_3 = 0.0
+";
+
+    /// An OMM whose every logical block carries a leading COMMENT, the shape
+    /// CCSDS 502.0-B-3 table 4-3 defines.
+    const OMM_KVN_SECTION_COMMENTS: &str = "\
```
duncaneddy: Should make test asset file

tests/ccsds/test_kvn_comments.py
```diff
@@ -0,0 +1,147 @@
...
+FIXTURES = [
+    "CDMExample1.txt",
+    "CDMExample2.txt",
+    "CDMExample3.txt",
+    "CDMExample4.txt",
+    "CDMExample_issue_940.txt",
+]
+
+
+OPM_KVN_SECTION_COMMENTS = """CCSDS_OPM_VERS = 3.0
```
duncaneddy: Should make test asset file

src/ccsds/kvn/parser.rs
duncaneddy: Renamed to `OMMSection`.

src/ccsds/kvn/parser.rs
duncaneddy: Renamed to `OPMSection`.

src/ccsds/kvn/parser.rs
duncaneddy: Renamed to `CDMBlock`.

src/ccsds/kvn/parser.rs
duncaneddy: Renamed to `ODParameters`.

src/ccsds/kvn/parser.rs
duncaneddy: Renamed to `RTNCovariance`.

src/ccsds/kvn/parser.rs
duncaneddy: Renamed to `XYZCovariance`.

src/ccsds/kvn/parser.rs
duncaneddy: Moved to `test_assets/ccsds/opm/OPM-section-comments.txt` and `test_assets/ccsds/omm/OMM-section-comments.txt`, read by both the Rust and Python tests.

These two also replaced the near-duplicate literals in the JSON tests one PR up: the OMM fixture was byte-identical and the OPM one was a strict subset, so there is now a single fixture per message type shared by the KVN tests, the JSON tests, and both Python suites.

src/ccsds/kvn/parser.rs
duncaneddy: Moved to `test_assets/ccsds/opm/OPM-section-comments.txt` and `test_assets/ccsds/omm/OMM-section-comments.txt`, read by both the Rust and Python tests.

These two also replaced the near-duplicate literals in the JSON tests one PR up: the OMM fixture was byte-identical and the OPM one was a strict subset, so there is now a single fixture per message type shared by the KVN tests, the JSON tests, and both Python suites.

tests/ccsds/test_kvn_comments.py
duncaneddy: Moved to `test_assets/ccsds/opm/OPM-section-comments.txt` and `test_assets/ccsds/omm/OMM-section-comments.txt`, read by both the Rust and Python tests.

These two also replaced the near-duplicate literals in the JSON tests one PR up: the OMM fixture was byte-identical and the OPM one was a strict subset, so there is now a single fixture per message type shared by the KVN tests, the JSON tests, and both Python suites.

PR #481, opened 2026-08-30
Round-trip CCSDS comments through JSON

duncaneddy:
# Pull Request

## Description

The JSON codecs emitted no `COMMENT` content at all, so every header, metadata, and block comment was silently dropped the first time a message was written as JSON. CCSDS 502.0-B-3 and 508.0-B-1 give each block a `COMMENT` row, and the KVN and XML codecs already carry them, so JSON was the one encoding that lost them.

Each block now writes a `comments` array beside its keywords, and the readers emit those back as `COMMENT` lines immediately ahead of the block they introduce, which is where the KVN parsers the JSON readers delegate to expect them. The CDM covariance is written as an ordered array of name/value pairs with no room for comments, so its comments ride in a sibling object.

Covers the OEM header, metadata, data, and covariance blocks; the OMM header, metadata, mean elements, TLE parameters, and spacecraft parameters; the OPM header, metadata, state vector, Keplerian elements, spacecraft parameters, covariance, and each maneuver; and the CDM header, relative metadata, object metadata, state vector, OD parameters, additional parameters, and RTN covariance.

### Fixed
- CCSDS OEM, OMM, OPM, and CDM JSON writers now emit each block's comments, and the JSON readers restore them to the block they belong to; previously every comment was dropped on the first write to JSON.

## Related Issue

Covers item 4 of #463, completing that issue alongside #474 (item 1) and the three PRs below this one in the stack.

## Note to Reviewers

The `comments` key is skipped by every flattener rather than being emitted as a `COMMENTS = ...` keyword line, so it cannot be mistaken for a CCSDS keyword on the way back in.

Separately, and not fixed here: `write_cdm_json` emits no `od_parameters` or `additional_parameters` block at all, so a CDM round-tripped through JSON loses both entirely (`back.object1.data.od_parameters.is_some()` is `false`). That is a data-loss defect rather than a comment one and deserves its own issue and PR.

### DIFF
--- a/src/ccsds/json.rs
+++ b/src/ccsds/json.rs
@@ -50,7 +50,7 @@ pub fn parse_oem_json(content: &str) -> Result<crate::ccsds::oem::OEM, BraheErro
 
     // Header
     if let Some(header) = v.get("header") {
-        flatten_object(&mut kvn_lines, header);
+        flatten_block(&mut kvn_lines, header);
     }
 
     // Segments
@@ -59,9 +59,14 @@ pub fn parse_oem_json(content: &str) -> Result<crate::ccsds::oem::OEM, BraheErro
             // Metadata block
             kvn_lines.push("META_START".to_string());
             if let Some(meta) = seg.get("metadata") {
+                emit_json_comments(&mut kvn_lines, meta);
                 flatten_object_ordered(&mut kvn_lines, meta, &TIME_SYSTEM_FIRST);
             }
             kvn_lines.push("META_STOP".to_string());
+
+            // Data-section comments sit between META_STOP and the first
+            // ephemeris line.
+            emit_json_comments(&mut kvn_lines, seg);
 
             // State vectors (data lines)
             if let Some(Value::Array(states)) = seg.get("states") {
@@ -110,7 +115,10 @@ pub fn parse_oem_json(content: &str) -> Result<crate::ccsds::oem::OEM, BraheErro
                 kvn_lines.push("COVARIANCE_START".to_string());
                 for cov in covariances {
                     if let Some(obj) = cov.as_object() {
-                        // Epoch
+                        // EPOCH delimits one covariance from the next, so it
+                        // leads: comments emitted ahead of it would be flushed
+                        // into the preceding block. This is the order the KVN
+                        // writer uses.
                         if let Some(epoch_val) = obj
                             .get("EPOCH")
                             .or_else(|| obj.get("epoch"))
@@ -126,6 +134,7 @@ pub fn parse_oem_json(content: &str) -> Result<crate::ccsds::oem::OEM, BraheErro
                         {
                             kvn_lines.push(format!("COV_REF_FRAME = {}", frame));
                         }
+                        emit_json_comments(&mut kvn_lines, cov);
                         // Lower-triangular values
                         if let Some(Value::Array(values)) =
                             obj.get("VALUES").or_else(|| obj.get("values"))
@@ -180,6 +189,9 @@ pub fn write_oem_json(
     header.insert(key("ORIGINATOR", key_case), json!(&oem.header.originator));
     if let Some(ref msg_id) = oem.header.message_id {
         header.insert(key("MESSAGE_ID", key_case), json!(msg_id));
+    }
+    if !oem.header.comments.is_empty() {
+        header.insert("comments".into(), json!(oem.header.comments));
     }
     root.insert("header".into(), Value::Object(header));
 
@@ -245,6 +257,9 @@ pub fn write_oem_json(
         if let Some(deg) = seg.metadata.interpolation_degree {
             meta.insert(key("INTERPOLATION_DEGREE", key_case), json!(deg));
         }
+        if !seg.metadata.comments.is_empty() {
+            meta.insert("comments".into(), json!(seg.metadata.comments));
+        }
         seg_obj.insert("metadata".into(), Value::Object(meta));
 
         // States (convert m → km, m/s → km/s for CCSDS standard units)
@@ -271,6 +286,9 @@ pub fn write_oem_json(
             }
             states.push(Value::Object(state_obj));
         }
+        if !seg.comments.is_empty() {
+            seg_obj.insert("comments".into(), json!(seg.comments));
+        }
         seg_obj.insert("states".into(), Value::Array(states));
 
         // Covariances
@@ -286,6 +304,9 @@ pub fn write_oem_json(
                 }
                 if let Some(ref frame) = cov.cov_ref_frame {
                     cov_obj.insert(key("COV_REF_FRAME", key_case), json!(format!("{}", frame)));
+                }
+                if !cov.comments.is_empty() {
+                    cov_obj.insert("comments".into(), json!(cov.comments));
                 }
                 // Convert m² → km² (factor 1e-6)
                 let values = covariance_to_lower_triangular(&cov.matrix, 1e-6);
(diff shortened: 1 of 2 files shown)


src/ccsds/json.rs
```diff
@@ -262,6 +273,9 @@ pub fn write_oem_json(
             }
             states.push(Value::Object(state_obj));
         }
+        if !seg.comments.is_empty() {
+            seg_obj.insert("comments".into(), json!(seg.comments));
```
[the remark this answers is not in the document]
duncaneddy: Fixed — the covariance `comments` array is now written in `write_oem_json`.

src/ccsds/json.rs
```diff
@@ -100,6 +104,7 @@ pub fn parse_oem_json(content: &str) -> Result<crate::ccsds::oem::OEM, BraheErro
             if let Some(Value::Array(covariances)) = seg.get("covariances") {
                 kvn_lines.push("COVARIANCE_START".to_string());
                 for cov in covariances {
+                    emit_json_comments(&mut kvn_lines, cov);
```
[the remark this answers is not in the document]
duncaneddy: Confirmed and fixed. Reproduced exactly as described: with two commented covariances the reparse gave `[0]=["first", "second"]` and `[1]=[]`. `EPOCH` and `COV_REF_FRAME` are now emitted before the comments, matching the KVN writer's order. `test_oem_json_keeps_each_covariance_comment_with_its_own_block` covers it.

src/ccsds/json.rs
```diff
@@ -1357,11 +1448,45 @@ fn emit_kvn(lines: &mut Vec<String>, ukey: &str, val: &Value) {
...
+    }
+}
+
+/// JSON key holding a block's comment array, in either case.
+const COMMENTS_KEY: &str = "COMMENTS";
+
+/// Emit a block's `comments` array as CCSDS `COMMENT` lines.
+///
+/// The KVN parsers attribute a comment to the block whose keywords follow it,
+/// so these are emitted immediately ahead of the block being flattened.
```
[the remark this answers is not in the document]
duncaneddy: Added complete rustdoc for both `emit_json_comments` and `flatten_block`.

src/ccsds/json.rs
```diff
@@ -1212,6 +1292,9 @@ pub fn write_cdm_json(
             key("Z_DOT", key_case),
             json!(d.state_vector.velocity[2] / 1e3),
         );
+        if !d.state_vector.comments.is_empty() {
+            sv.insert("comments".into(), json!(d.state_vector.comments));
+        }
```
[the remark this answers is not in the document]
duncaneddy: The OD and additional parameter blocks are now written in full, in #486 further up the stack — they were missing entirely, not just their comments.

The data-level comments needed a different fix. KVN places nothing between the Data `COMMENT` row and the first sub-block's, so the parser attributes a run spanning both to the innermost block; that is the agreed reading and it means `data.comments` cannot be recovered from flattened KVN text at all. Since JSON does keep them apart, the CDM JSON reader now takes the object-level `comments` array from the document directly rather than routing it through the KVN form. Covered by `test_cdm_json_round_trip_preserves_data_section_comments`.

src/ccsds/json.rs
```diff
@@ -2409,4 +2576,178 @@ mod tests {
         assert!((acc[1] - 0.002e3).abs() < 1e-6);
         assert!((acc[2] - 0.003e3).abs() < 1e-6);
     }
+
+    /// OMM in KVN with a comment introducing every block that can carry one.
+    const OMM_KVN_WITH_COMMENTS: &str = "CCSDS_OMM_VERS = 3.0
```
duncaneddy: Should make test asset file

src/ccsds/json.rs
```diff
@@ -2409,4 +2576,178 @@ mod tests {
...
+COMMENT spacecraft comment
+MASS = 300.0
+SOLAR_RAD_AREA = 5.0
+SOLAR_RAD_COEFF = 1.2
+DRAG_AREA = 5.0
+DRAG_COEFF = 2.2
+";
+
+    /// OPM in KVN with a comment introducing every block that can carry one.
+    const OPM_KVN_WITH_COMMENTS: &str = "CCSDS_OPM_VERS = 3.0
```
duncaneddy: Should make test asset file

tests/ccsds/test_json_comments.py
```diff
@@ -0,0 +1,106 @@
+"""Tests for CCSDS JSON comment fidelity — parity with Rust tests in src/ccsds/json.rs."""
+
+import brahe  # noqa: F401
+from brahe.ccsds import CDM, OEM, OMM, OPM
+
+OMM_KVN_WITH_COMMENTS = """CCSDS_OMM_VERS = 3.0
```
duncaneddy: Should make test asset file

tests/ccsds/test_json_comments.py
```diff
@@ -0,0 +1,106 @@
...
+EPOCH = 2007-064T10:34:41.4264
+MEAN_MOTION = 1.00273272
+ECCENTRICITY = 0.0005013
+INCLINATION = 3.0539
+RA_OF_ASC_NODE = 81.7939
+ARG_OF_PERICENTER = 249.2363
+MEAN_ANOMALY = 150.1602
+"""
+
+OPM_KVN_WITH_COMMENTS = """CCSDS_OPM_VERS = 3.0
```
duncaneddy: Should make test asset file

src/ccsds/json.rs
duncaneddy: Removed rather than moved to new files: these were near-duplicates of the KVN comment fixtures added one PR down. The OMM literal was byte-identical to `test_assets/ccsds/omm/OMM-section-comments.txt` and the OPM one was a strict subset of `test_assets/ccsds/opm/OPM-section-comments.txt`, so both tests now read those.

One fixture per message type, shared by the KVN tests, the JSON tests, and both Python suites. The OPM fixture carries two maneuvers, so the assertion is now stronger — it checks both maneuvers keep their own comment instead of just the first.

src/ccsds/json.rs
duncaneddy: Removed rather than moved to new files: these were near-duplicates of the KVN comment fixtures added one PR down. The OMM literal was byte-identical to `test_assets/ccsds/omm/OMM-section-comments.txt` and the OPM one was a strict subset of `test_assets/ccsds/opm/OPM-section-comments.txt`, so both tests now read those.

One fixture per message type, shared by the KVN tests, the JSON tests, and both Python suites. The OPM fixture carries two maneuvers, so the assertion is now stronger — it checks both maneuvers keep their own comment instead of just the first.

tests/ccsds/test_json_comments.py
duncaneddy: Removed rather than moved to new files: these were near-duplicates of the KVN comment fixtures added one PR down. The OMM literal was byte-identical to `test_assets/ccsds/omm/OMM-section-comments.txt` and the OPM one was a strict subset of `test_assets/ccsds/opm/OPM-section-comments.txt`, so both tests now read those.

One fixture per message type, shared by the KVN tests, the JSON tests, and both Python suites. The OPM fixture carries two maneuvers, so the assertion is now stronger — it checks both maneuvers keep their own comment instead of just the first.

tests/ccsds/test_json_comments.py
duncaneddy: Removed rather than moved to new files: these were near-duplicates of the KVN comment fixtures added one PR down. The OMM literal was byte-identical to `test_assets/ccsds/omm/OMM-section-comments.txt` and the OPM one was a strict subset of `test_assets/ccsds/opm/OPM-section-comments.txt`, so both tests now read those.

One fixture per message type, shared by the KVN tests, the JSON tests, and both Python suites. The OPM fixture carries two maneuvers, so the assertion is now stronger — it checks both maneuvers keep their own comment instead of just the first.

PR #484, opened 2026-08-30
Emit covariance EPOCH only where the ODM defines it

duncaneddy:
# Pull Request

## Description

CCSDS 502.0-B-3 gives the OEM covariance block an `EPOCH` keyword (table 5-4, subsection 5.2.5.3) because each matrix belongs to a navigation solution of its own. Tables 3-3 and 4-3 give the OPM and OMM covariance blocks only `COMMENT`, `COV_REF_FRAME`, and the matrix entries; their covariance applies to the message's single state and no epoch is defined for it.

The OPM and OMM writers emitted the stored covariance epoch anyway, in both KVN and XML. Beyond the conformance error this corrupted data: the KVN parser matches `EPOCH` positionally, so the second assignment overwrote the state vector's own epoch on re-read while the covariance came back with no epoch at all. A message whose covariance epoch differed from its state epoch silently moved its state — for `OPMExample3` with the covariance epoch set an hour later, the state epoch came back an hour late.

Two smaller fidelity gaps found alongside it are folded in: the OEM JSON writer omitted covariance comments that its own reader already accepted, and the CDM JSON writer omitted the data-section comments; and the CDM XML reader ignored CDATA sections, discarding any text written that way.

### Fixed
- CCSDS OPM and OMM writers no longer emit an `EPOCH` in the covariance block, which CCSDS 502.0-B-3 tables 3-3 and 4-3 do not define. The stray keyword overwrote the state vector's epoch when the message was read back, silently moving the state. The OEM covariance block, where `EPOCH` is required, is unchanged.
- CCSDS OEM JSON output now includes covariance comments, and CDM JSON output includes the data-section comments; both were read but never written.
- CCSDS CDM XML parsing now reads CDATA sections, which were previously discarded.

## Related Issue

Not in the original issues; found by the Codex review of this stack.

## Note to Reviewers

The covariance-epoch defect predates this stack but the time-system PR below touches the same emission path, so it is fixed here rather than left in place. Tests assert the OPM and OMM state epochs survive a KVN and XML round trip with a differing covariance epoch set, and that the OEM keeps its covariance epoch in both encodings.

Epoch comparisons in the tests are made on the written time codes rather than on `Epoch` values, because an `Epoch` carries sub-nanosecond representation noise that a CCSDS time code does not encode.

### DIFF
--- a/src/ccsds/json.rs
+++ b/src/ccsds/json.rs
@@ -1080,7 +1080,25 @@ pub fn parse_cdm_json(content: &str) -> Result<crate::ccsds::cdm::CDM, BraheErro
     }
 
     let kvn_content = kvn_lines.join("\n");
-    crate::ccsds::kvn::parse_cdm(&kvn_content)
+    let mut cdm = crate::ccsds::kvn::parse_cdm(&kvn_content)?;
+
+    // The object Data block and its sub-blocks each carry a COMMENT row, but
+    // KVN places nothing between them, so the parser attributes a run spanning
+    // both to the innermost block. JSON keeps them apart, so the data-level
+    // comments are read from the document rather than from the flattened text.
+    for (key, object) in [("object1", &mut cdm.object1), ("object2", &mut cdm.object2)] {
+        if let Some(obj) = v.get(key).or_else(|| v.get(key.to_uppercase()))
+            && let Some(Value::Array(comments)) =
+                obj.get("comments").or_else(|| obj.get("COMMENTS"))
+        {
+            object.data.comments = comments
+                .iter()
+                .filter_map(|c| c.as_str().map(|s| s.to_string()))
+                .collect();
+        }
+    }
+
+    Ok(cdm)
 }
 
 /// Write a CDM message to JSON format.
@@ -1312,6 +1330,9 @@ pub fn write_cdm_json(
             key("Z_DOT", key_case),
             json!(d.state_vector.velocity[2] / 1e3),
         );
+        if !d.comments.is_empty() {
+            o.insert("comments".into(), json!(d.comments));
+        }
         if !d.state_vector.comments.is_empty() {
             sv.insert("comments".into(), json!(d.state_vector.comments));
         }
@@ -2668,6 +2689,24 @@ mod tests {
             .unwrap();
         assert_eq!(segment.covariances[0].comments, vec!["first covariance"]);
         assert_eq!(segment.covariances[1].comments, vec!["second covariance"]);
+    }
+
+    #[test]
+    #[serial_test::parallel]
+    fn test_cdm_json_round_trip_preserves_data_section_comments() {
+        use crate::ccsds::cdm::CDM;
+
+        // KVN puts nothing between the Data comment and the first sub-block
+        // comment, so the parser attributes the run to the innermost block.
+        // JSON keeps them apart, so this level survives there.
+        let source = std::fs::read_to_string("test_assets/ccsds/cdm/CDMExample2.txt").unwrap();
+        let mut cdm = CDM::from_str(&source).unwrap();
+        cdm.object1.data.comments = vec!["Object1 Data".to_string()];
+        cdm.object2.data.comments = vec!["Object2 Data".to_string()];
+
+        let reparsed = CDM::from_str(&cdm.to_string(CCSDSFormat::JSON).unwrap()).unwrap();
+        assert_eq!(reparsed.object1.data.comments, vec!["Object1 Data"]);
+        assert_eq!(reparsed.object2.data.comments, vec!["Object2 Data"]);
     }
 
     #[test]
--- a/src/ccsds/kvn/writer.rs
+++ b/src/ccsds/kvn/writer.rs
@@ -283,12 +283,6 @@ pub fn write_omm(omm: &crate::ccsds::omm::OMM) -> Result<String, BraheError> {
         for comment in &cov.comments {
             out.push_str(&format!("COMMENT {}\n", comment));
         }
-        if let Some(ref epoch) = cov.epoch {
-            out.push_str(&format!(
-                "EPOCH = {}\n",
-                format_ccsds_datetime_in(epoch, &omm.metadata.time_system)
-            ));
-        }
         if let Some(ref frame) = cov.cov_ref_frame {
             out.push_str(&format!("COV_REF_FRAME = {}\n", frame));
         }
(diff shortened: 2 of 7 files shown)


src/ccsds/json.rs
```diff
@@ -1292,6 +1295,9 @@ pub fn write_cdm_json(
             key("Z_DOT", key_case),
             json!(d.state_vector.velocity[2] / 1e3),
         );
+        if !d.comments.is_empty() {
+            o.insert("comments".into(), json!(d.comments));
+        }
```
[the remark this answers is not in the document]
duncaneddy: Confirmed and fixed — a JSON round trip returned `[]` for `data.comments`.

Emitting them into the flattened KVN does not work: the innermost-block rule files them under the object metadata, because `OBJECT` is the next keyword. The CDM JSON reader now reads the object-level `comments` array structurally instead, which is the one thing JSON can express and KVN cannot. `test_cdm_json_round_trip_preserves_data_section_comments` compares the buckets rather than just the text, as you suggested.

tests/ccsds/test_covariance_epoch.py
```diff
@@ -0,0 +1,32 @@
...
+from brahe.ccsds import OEM, OPM
+
+
+def test_opm_covariance_block_omits_epoch(eop):
+    """Mirror of test_opm_and_omm_covariance_blocks_omit_epoch in Rust."""
+    opm = OPM.from_file("test_assets/ccsds/opm/OPMExample3.txt")
+
+    for fmt in ["kvn", "xml"]:
+        written = opm.to_string(fmt)
+        assert OPM.from_str(written).epoch == opm.epoch
```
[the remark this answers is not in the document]
duncaneddy: Fair — the test passed before the fix. It now parses an OPM from XML with an `EPOCH` injected into the covariance block, so the writer has a populated covariance epoch to wrongly echo, and asserts the state epoch survives KVN and XML.

src/ccsds/json.rs
```diff
@@ -2639,6 +2645,28 @@ MAN_DV_3 = 0.0
         assert_eq!(reparsed.maneuvers[0].comments, vec!["maneuver comment"]);
     }
 
+    #[test]
+    #[serial_test::parallel]
+    fn test_oem_json_round_trip_preserves_covariance_comments() {
```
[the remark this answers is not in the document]
duncaneddy: Added to `tests/ccsds/test_json_comments.py`.

src/ccsds/xml/parser.rs
```diff
@@ -1844,6 +1852,21 @@ mod tests {
         assert_eq!(cdm.header.comments[0], "Sample CDM & more");
     }
 
+    #[test]
+    #[serial_test::parallel]
+    fn test_parse_cdm_xml_reads_cdata_sections() {
```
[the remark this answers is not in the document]
duncaneddy: Added `test_parse_cdm_xml_reads_cdata_sections` to `tests/ccsds/test_xml_parser.py`.

src/ccsds/kvn/writer.rs
```diff
@@ -283,12 +283,11 @@ pub fn write_omm(omm: &crate::ccsds::omm::OMM) -> Result<String, BraheError> {
...
-            out.push_str(&format!(
-                "EPOCH = {}\n",
-                format_ccsds_datetime_in(epoch, &omm.metadata.time_system)
-            ));
-        }
+        // CCSDS 502.0-B-3 tables 3-3 (OPM) and 4-3 (OMM) list only COMMENT,
+        // COV_REF_FRAME, and the matrix entries in the covariance block: EPOCH
+        // belongs to the OEM covariance block alone (table 5-4, 5.2.5.3).
+        // Emitting it here produced a second EPOCH assignment that the parser
+        // read back as the state vector's own epoch.
```
duncaneddy: Remove unnecessary comment

src/ccsds/kvn/writer.rs
```diff
@@ -408,12 +407,11 @@ pub fn write_opm(opm: &crate::ccsds::opm::OPM) -> Result<String, BraheError> {
...
-            out.push_str(&format!(
-                "EPOCH = {}\n",
-                format_ccsds_datetime_in(epoch, &opm.metadata.time_system)
-            ));
-        }
+        // CCSDS 502.0-B-3 tables 3-3 (OPM) and 4-3 (OMM) list only COMMENT,
+        // COV_REF_FRAME, and the matrix entries in the covariance block: EPOCH
+        // belongs to the OEM covariance block alone (table 5-4, 5.2.5.3).
+        // Emitting it here produced a second EPOCH assignment that the parser
+        // read back as the state vector's own epoch.
```
duncaneddy: Remove unnecessary comment

src/ccsds/kvn/writer.rs
```diff
@@ -2295,4 +2293,102 @@ mod tests {
         let written = write_cdm(&cdm).unwrap();
         assert!(written.contains("COMMENT State vector comment"));
     }
+
+    #[test]
+    #[serial_test::parallel]
+    fn test_opm_and_omm_covariance_blocks_omit_epoch() {
```
duncaneddy: Overly defensive regression test. Consider dropping, since it tests for a specific implementation that we don't/won't be doing.

tests/ccsds/test_covariance_epoch.py
```diff
@@ -0,0 +1,84 @@
+"""Tests for CCSDS covariance EPOCH conformance — parity with Rust tests in src/ccsds/kvn/writer.rs."""
+
+import brahe  # noqa: F401
+from brahe.ccsds import OEM, OPM
+
+# An OPM whose covariance block carries an EPOCH, which CCSDS 502.0-B-3
+# table 3-3 does not define. The old writers echoed it back out, and the KVN
+# parser then applied it to the state vector.
+OPM_XML_WITH_COVARIANCE_EPOCH = """<?xml version="1.0" encoding="UTF-8"?>
```
duncaneddy: Consider adding as test data file asset own it's own.

src/ccsds/kvn/writer.rs
duncaneddy: Removed. It was the same five-line note repeated in both writers.

src/ccsds/kvn/writer.rs
duncaneddy: Removed. It was the same five-line note repeated in both writers.

src/ccsds/kvn/writer.rs
duncaneddy: Agreed on the defensive part. It built an OPM with a covariance epoch — something brahe will never emit and the spec does not define — purely to prove the writer would not echo it, and did the same for the OMM.

Replaced with `test_opm_covariance_block_omits_epoch`, which writes the stock `OPMExample3` fixture and asserts there is exactly one `EPOCH` line, the state vector's. No artificial construction, no round trip, and it still fails if the covariance `EPOCH` comes back. The OEM companion test that checks the block does keep its epoch is unchanged.

This also removed the need for the synthetic XML fixture you flagged below.

tests/ccsds/test_covariance_epoch.py
duncaneddy: No longer needed — dropped along with the defensive test above. The Python test now uses the stock `OPMExample3` fixture and makes the same single assertion, so there is no synthetic message to turn into an asset.

PR #485, opened 2026-08-30
Serve Celestrak groups from committed snapshots

duncaneddy:
# Pull Request

## Description

CI downloaded seven Celestrak GP groups on every warm, and all 27 Celestrak-dependent examples ran on the default PR path with no fallback. Every `brahe-celestrak-*` actions cache was scoped to a PR merge-ref or a feature branch, and GitHub caches are readable only from their own ref or the default branch, never from siblings — so a PR branched from `main` found nothing and fell through to a live download. When Celestrak was unreachable that download hung rather than failing, holding a runner until GitHub's six-hour default timeout, which is what wedged run 33292681663.

This commits the seven groups under `test_assets/celestrak` and seeds them into `~/.cache/brahe/celestrak` alongside `de440s.bsp`. Example and doc-plot runs set `BRAHE_NETWORK_MODE=offline`, which serves cached files regardless of age, so the snapshots are used as-is and no pull-request, docs, or example path contacts Celestrak. Verified end-to-end: all seven groups plus the ISS single-object lookup by catalog number resolve from the seeded cache with no network access.

Live client regressions stay covered by the existing integration tests, which deliberately do not restore a Celestrak cache and run with `BRAHE_NETWORK_MODE` unset (defaulting to `Online`).

The group list stays in `.github/brahe-data-manifest.txt` as the single source of truth, now consumed by `scripts/refresh_celestrak_snapshots.py`. That script fetches through `CelestrakClient` rather than `curl`, so on-disk names come from the client's own cache-key derivation and cannot drift from what it looks up. Snapshots are marked `-text` so they stay byte-identical to a live response rather than being line-ending normalized.

Refresh with `just refresh-celestrak-snapshots`.

This is the first PR in a stack reducing PR CI latency from ~79 minutes to a target of 17-20, sequenced first so every later PR in the stack is immune to Celestrak availability.

### Changed
- Celestrak GP groups used by examples and documentation plots are committed under `test_assets/celestrak` and seeded into the local cache, so example and plot runs no longer contact Celestrak. Live client behavior remains covered by the integration test suite.

### Fixed
- Example and documentation-plot CI runs no longer hang when Celestrak is unreachable. The groups were previously fetched live whenever no restorable cache existed, which was the normal case for a pull request branched from `main`, since every saved Celestrak cache was scoped to a ref other pull requests cannot read.

### Removed
- `scripts/warm_data_cache.py` no longer warms Celestrak groups, and its `--refresh` flag is removed along with the Celestrak family it existed to clear. The `celestrak` family is removed from the data-cache action.

## Note to Reviewers

Scope: this removes Celestrak from the pull-request, docs, and example paths. The scheduled integration jobs still query Celestrak live, deliberately — that is what covers client regressions. Separately, `scripts/update-data.sh` fetches space-weather data (`sw19571001.txt`) from celestrak.org on the scheduled bundled-data workflow; that is a different resource and is unchanged here.

The snapshot filenames encode the request URL and must match `CelestrakClient::cache_key_for_url` exactly, including the hyphen-to-underscore mapping in group names and the `FORMAT=JSON` suffix that `query_gp` forces internally. `scripts/refresh_celestrak_snapshots.py` derives them via the client for that reason rather than reconstructing them.

Worth a look: `cosmos-1408-debris` contains only 3 objects. That is correct — the 2021 ASAT debris cloud has almost entirely decayed, and the epochs are current. All four debris groups combine into one 2,649-object set for the gabbard plots, so the small group is one series rather than a degenerate figure.

Committing these snapshots freezes the TLE epoch the docs display. They were already up to 60 days stale via `cache_max_age`, so this mainly makes example output deterministic; it now moves only on a deliberate refresh.


duncaneddy:
Addressed the review comments:

**Skip snapshot seeding during live integration runs (P2)** — fixed. This was a real defect: the unconditional copy refreshed every file's mtime, and since the client serves anything inside its TTL without a request, the live-network run invoked from `integration_tests.yml` was reading the committed snapshots instead of querying Celestrak. That would have quietly disabled the one run whose purpose is to exercise the remote. Seeding is now gated on `inputs.network_mode != 'online'`.

**Add coverage for the snapshot refresh workflow (P1)** — added `tests/scripts/test_refresh_celestrak_snapshots.py`, five tests covering manifest parsing and the install rules that decide whether the command can destroy committed fixtures: a partial refresh leaves untouched groups alone and leaves no staging files behind, and only a full refresh prunes groups absent from the manifest.

**Obtain approval before changing these tests (P1)** — approval was obtained before the change. The maintainer was asked explicitly and chose to update the four tests that still apply and remove the two whose subject functions this PR deletes (`_warm_celestrak`, and the `--refresh` flag). Two new tests were added covering the replacement behaviour, so the file went from 11 tests to 11 with strictly more coverage of the current code.

PR #489, opened 2026-08-30
Bound every CI job and network step with a timeout

duncaneddy:
# Pull Request

## Description

No workflow declared `timeout-minutes`, so any job that stopped making progress ran until GitHub's six-hour default. Because the account caps at 20 concurrent jobs, a single wedged job held 5% of total CI capacity for six hours and starved every queued run. This is not hypothetical — run 33292681663 wedged exactly this way on a warm step waiting for an unreachable Celestrak.

All 39 jobs are now capped at 60 minutes, with `test-examples` at 90.

These are circuit breakers, not performance budgets. They sit two to four times above every observed job duration, so ordinary growth in the test suites never trips them. For reference, the slowest observed jobs are `test-rust-docs (windows)` at 51 min, `test-examples` at 31 min, and `test-python (windows)` at 24 min. A 30-minute cap was considered and rejected: it would leave `test-python (windows)` only 15% headroom and begin flaking as the suite grows.

Separately, the steps that reach the network are capped at 15 minutes. Every hang observed so far was a warm or restore step rather than test execution, so this is the bound that actually catches a dead remote — and it attributes the failure to the responsible step rather than to the job as a whole.

Second PR in a stack reducing PR CI latency from ~79 minutes toward 17-20.

### Fixed
- A continuous-integration job that stops making progress, such as one waiting on an unreachable data mirror, now fails within a bounded time instead of occupying a runner until the six-hour default limit and delaying other work.

## Note to Reviewers

Mechanical and additive: no logic, ordering, or step content changes, only `timeout-minutes` keys. Both indentation styles in the workflow tree are handled, and the two-tier structure is the substantive choice — a generous job cap that never binds, plus a tight step cap where hangs actually originate.

Verified every `runs-on:` has an accompanying cap and that all 15 workflows still parse.


duncaneddy:
Addressed all three review comments — they identified the same real defect in three places, and it was the most serious finding in this stack.

The job caps I introduced sat **below** the retry budgets already configured in those workflows:

| Job | Configured budget | Cap as submitted |
|---|---|---|
| `integration-rust` | 3 x 30 min + waits ≈ 92 min | 60 |
| `integration-python` | ≈ 92 min + extension build | 60 |
| `test-examples` (retry path) | 3 x 60 min + waits ≈ 182 min | 90 |

Since `release.yml` gates publishing on the integration workflow, a slow remote could have failed a release despite the retries existing precisely to survive that.

Integration caps raised to 120 minutes. For `test_examples.yml` a single static cap cannot serve both callers, whose budgets differ by more than 2x, so the cap is now a `job-timeout-minutes` input: the pull-request path keeps its tight 90-minute bound, and `integration_tests.yml` passes 200 for the retrying live-network run.

PR #495, opened 2026-08-30
Reuse the rust-script build across example runs

duncaneddy:
# Pull Request

## Description

`test-examples` is the slowest job on the pull-request path at 31 minutes, because all 376 Rust examples compiled their entire dependency graph from scratch on every run.

The cause was not a missing cache step. Each example was written to a randomly named `tempfile.NamedTemporaryFile` before being handed to `rust-script`, and **rust-script keys its build cache on the script's path as well as its contents**. Verified directly: re-running the same path reuses the cached build, while identical content at a different path creates a new cache entry. A random path every run therefore made the cache unreachable by construction, and simply adding an `actions/cache` step would have had no effect.

Deriving the scratch path from the example's location fixes it. Measured on `examples/getting_started/first_script.rs`:

| Run | Time |
|---|---|
| cold, no reusable build | 83.16 s |
| warm, build reused | 3.00 s |
| shared deps present, example binary removed | 0.99 s |

That third row drives the caching strategy. Only the shared dependency build is saved; the per-example binaries are pruned before the save, because they relink in about a second each while keeping 376 of them would grow the cache well past the repository's budget.

The size argument: the shared `deps` tree is 1.9 GB and is compiled once for all examples, whereas each example binary is around 12 MB. Caching everything would approach 6.5 GB against GitHub's 10 GB per-repository limit and crowd out the NAIF, uv, and cargo caches. Pruning keeps the cache bounded rather than growing with the number of examples, and costs roughly one second per example to relink.

Fifth PR in a stack reducing PR CI latency from ~79 minutes toward 17-20.

### Fixed
- Rust examples reuse a shared compiled dependency build between runs instead of rebuilding it for every example, which was previously prevented by generating each example at a new temporary path.

## Note to Reviewers

The behavioral change in `scripts/_build_utils.py` is that the generated scratch file now lives at a stable path under the system temporary directory rather than a random one, and is no longer unlinked afterward — deleting it would discard the cache entry it addresses. Files are overwritten in place on each run rather than accumulating, and names are unique per example so the parallel workers cannot collide.

The cache save is gated to `main`, matching the existing restore-only-on-pull-request convention used by the cargo caches, so pull requests read the shared copy without writing unrestorable ones.

Worth checking: the prune step assumes per-example binaries are the top-level files in `binaries/release` and that their fingerprint directories are prefixed `examples__`. That holds for scratch names derived from paths under `examples/`, which is every case this script produces.


duncaneddy:
Addressed the review comment — a real collision, and one that matters here because this repository is routinely worked in via git worktrees, so two checkouts under the same user is the normal case rather than an edge case.

The scratch directory is now namespaced by a SHA-256 digest of `REPO_ROOT`, truncated to 12 characters. Verified the three properties that matter: the path is stable across runs for a given checkout (so the rust-script build cache stays reusable, which is the point of the PR), distinct per example (so parallel workers cannot collide), and distinct across checkouts (so a clone and a worktree no longer overwrite each other's injected repository path).

PR #497, opened 2026-08-30
Adopt a dev integration branch and a lean pull-request gate

duncaneddy:
# Pull Request

## Description

A pull request spawned 18 jobs and a push to `main` about 26, against an account limit of **20 concurrent jobs with a sub-limit of 5 on macOS**. Runs queue rather than fail at those limits, so execution time was never the real cost. From run 33293001066:

| Job | Runtime | Queue wait before start |
|---|---|---|
| `test-rust (ubuntu)` | 4.5 min | **58 min** |
| `test-rust-docs (ubuntu)` | 11.3 min | **54 min** |
| `test-python (windows)` | 25.0 min | **54 min** |

A single push to `main` requested 7 macOS jobs on its own, exceeding the macOS sub-limit before any pull request competed for a runner.

This introduces `dev` as the integration branch and narrows what a pull request runs.

| Stage | What runs |
|---|---|
| PR into `dev` | Rust unit tests on all three platforms; doc tests on ubuntu and macOS; Python 3.10 fully on ubuntu and macOS, smoke subset on Windows; examples, packaging, docs build |
| Push to `dev` | Full matrix: doc tests on all three platforms, every supported Python on ubuntu and Windows, floor and ceiling on macOS |
| Push to `main` | Same full matrix, plus documentation deployment and release workflows |

Windows doc tests — at 51 minutes the longest job by a wide margin — leave the pull-request path entirely.

The Windows Python scope is the `python-test-scope` workflow input rather than a hardcoded path, with the subset defined once as `SMOKE_PATHS`, so widening it is a one-line change. The `full-ci` label runs the entire matrix for a single pull request on demand.

macOS carries only the floor and ceiling Python versions, because the full five plus the Rust and doc-test legs exceed the five-runner sub-limit by themselves.

Seventh PR in a stack reducing PR CI latency from ~79 minutes toward 17-20.

### Added
- A `full-ci` pull request label that runs the complete cross-platform test matrix on demand, and a `just release-promote` command that opens the release promotion pull request with the changelog aggregated across everything merged since the previous tag.

### Changed
- Pull requests run a narrower test matrix, with the full cross-platform and cross-version matrix running when changes land on the integration branch. Applying the `full-ci` label restores the complete matrix for a single pull request.

## Note to Reviewers

**This pull request does not create the `dev` branch or change branch protection.** Those are repository administration steps, listed below, and everything here is inert until they are taken: `unit_tests.yml` now triggers on pushes to `main` *and* `dev`, so behavior on `main` is unchanged if `dev` never exists.

Manual steps to complete the transition:

1. Create `dev` from `main`.
2. Set `dev` as the default branch so pull requests target it automatically.
3. Move branch protection to `dev`, requiring the `ci-success` check.
4. Protect `main`, and — importantly — **disable squash merging for `main`**, or at minimum never squash a promotion.

That last point is the one thing here that fails silently rather than loudly. `scripts/generate_release_notes.py` derives release notes from `git log <prev_tag>..HEAD --no-merges`, parsing `(#NNN)` out of squash-merge subjects. Pull requests squashed onto `dev` keep those subjects, and they survive into `main` through a merge commit. Squashing the promotion instead collapses every pull request reference into one subject and produces an empty changelog, with no error.

Cache saves now also run on `dev` because a pull request can only restore caches from its own ref or the default branch; without it every pull request based on `dev` would start cold. The `data-cache` composite action treats `dev` like `main` for the same reason.

`docs_only.yml` is deliberately left gated on `main`, since it only triggers on pushes there and a `dev` clause would be dead code.


duncaneddy:
Removed from the CI-performance stack — the `dev` branch topology is being revisited separately, so this is parked rather than abandoned.

The branch is preserved and the work stands on its own. What is in here:

- `os-matrix` / `doc-os-matrix` inputs on `test_rust.yml`, and `os-matrix` / `python-test-scope` (with `SMOKE_PATHS`) on `test_python.yml`
- A lean pull-request matrix with a `full-ci` label that widens it on demand, including the `labeled`/`unlabeled` triggers that make the label take effect immediately
- `unit_tests.yml` triggering on `dev`, with macOS held to the floor and ceiling Python versions so a run stays inside the five-runner macOS sub-limit
- Cache saves extended to `dev`
- `just release-promote`, and the `release` label exemption in the changelog check

Worth noting for whenever this is picked up again: the `release` label exemption lived here, and PR #498 originally depended on it. #498 has been rebuilt to stand alone, so if this branch is revived that exemption will need re-adding on top of the current state of `validate_pr_changelog.yml` rather than merging as-is.

The one constraint worth carrying forward regardless of topology: `scripts/generate_release_notes.py` reads PR numbers from squash-merge subjects via `git log <tag>..HEAD --no-merges`. Any `dev`-to-`main` promotion must be a merge commit or fast-forward — squashing it collapses every PR reference into one subject and silently produces an empty changelog.

PR #502, opened 2026-08-31
Run a Windows Python smoke subset on pull requests

duncaneddy:
# Pull Request

## Description

Windows Python is roughly 1.8x slower than ubuntu — 24-26 minutes against 13-14 — so running the full suite there sets the critical path for every pull request. This runs the subset that actually carries the platform risk instead:

`tests/cli`, `tests/utils`, `tests/eop`, `tests/space_weather`, `tests/datasets` — **402 of 3,534 tests**, collecting and passing in 1.7 s locally.

That selection is not arbitrary. It is where Windows-specific Python breakage has actually happened in this repository:

| Commit | Fixed | Location |
|---|---|---|
| `16755ce2` | Windows-only CI test failures | `tests/cli/test_eop.py`, `tests/cli/test_groundstations.py` |
| `e486a937` | cache-dir and write handling | `src/utils/cache.rs`, `src/datasets/horizons/client.rs` |

The remainder of the suite is platform-independent computation — orbits, coordinates, frames, propagators — which the Rust matrix already exercises on all three platforms, and `src/` contains no `cfg(windows)` blocks at all. Note also that the second commit above was a Windows *library* defect caught by the **Rust** suite, which this PR keeps running on Windows in full.

The full Windows Python suite still runs on push to `main`.

### Changed
- Pull requests run a Windows Python smoke subset covering command-line, cache, and dataset handling rather than the full suite, which continues to run when changes land on the main branch.

## Note to Reviewers

Scope is a `python-test-scope` workflow input taking `full` or `smoke`, not a hardcoded path, and the directory list is defined once as `SMOKE_PATHS`. Widening it later is a one-line change in a single place.

This was originally part of the `dev` branch topology work in #497. It has no dependency on that topology, so it is separated out here; #497 remains parked.

The accepted risk is explicit: a Windows-only Python failure outside those five directories would surface on merge to `main` rather than on the pull request. Given the history above, that risk sits in the platform-independent portion of the suite.

### DIFF
--- a/.github/workflows/pr_tests.yml
+++ b/.github/workflows/pr_tests.yml
@@ -37,16 +37,33 @@ jobs:
       - name: Wait for Rust tests to claim runners
         run: sleep 15
 
+  # Pull requests test the oldest supported Python (the declared floor in
+  # pyproject.toml) to catch breakage of the minimum-supported version. The full
+  # version matrix runs on push to main.
   test-python:
     needs: [delay-python]
     if: needs.delay-python.result == 'success'
     uses: ./.github/workflows/test_python.yml
     with:
-      # PR runs test only the oldest supported Python (the declared floor in pyproject.toml)
-      # across all 3 OSes to catch breakage of the minimum-supported version.
-      # The full matrix runs on push to main.
       python-versions: '["3.10"]'
       coverage-python-version: '3.10'
+      os-matrix: '["ubuntu-latest", "macos-latest"]'
+      python-test-scope: full
+    secrets: inherit
+
+  # Windows runs the smoke subset rather than the whole suite. It is roughly
+  # 1.8x slower than ubuntu, so the full run would set the critical path, while
+  # the subset covers every area with a recorded history of Windows-only Python
+  # failures. The full Windows suite still runs on push to main.
+  test-python-windows:
+    needs: [delay-python]
+    if: needs.delay-python.result == 'success'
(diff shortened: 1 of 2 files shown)


duncaneddy:
Parked at your request — the full Windows Python suite stays for now.

This is standalone and does not depend on anything else in the CI stack, so it can be picked up whenever. What it does:

- Adds a `python-test-scope` input to `test_python.yml` taking `full` or `smoke`, with the subset defined once as `SMOKE_PATHS` (`tests/cli`, `tests/utils`, `tests/eop`, `tests/space_weather`, `tests/datasets` — 402 of 3,534 tests)
- Adds an `os-matrix` input so a caller can target specific runners
- Splits `test-python-windows` out as its own job in `pr_tests.yml`, with `ci-success` updated to require it

Worth noting the `os-matrix` and scope inputs are useful independently of the smoke subset — they are what would let a caller run the full suite on some platforms and a narrower one elsewhere, or vary the matrix per branch. If the `dev` topology in #497 is revisited, it depends on these inputs existing.

For the record, the case for the subset was that `tests/cli` is where Windows-only Python failures have actually occurred (commit `16755ce2`), while the Windows *library* defect in `e486a937` was caught by the Rust suite, which runs in full on Windows either way. Keeping the full suite is the more conservative call; it costs roughly 24 minutes against 7 on that leg.

PR #503, opened 2026-08-31
Stop rebuilding the extension during dependency install

duncaneddy:
# Pull Request

## Description

Installing dependencies took about **seven minutes on every leg of the Python matrix**. The uv cache was not the cause — it hits, and the third-party packages install from it almost instantly. From a real job log:

```
Cache hit for: setup-uv-2-x86_64-unknown-linux-gnu-ubuntu-24.04-3.12.3-f4c2c75c...
Cache restored successfully

Building brahe @ file:///home/runner/work/brahe/brahe    20:44:53
Built   brahe @ file:///home/runner/work/brahe/brahe     20:51:53
Prepared 1 package in 6m 59s
Installed 140 packages in 345ms
```

**140 packages in 345 ms; one package in 6 minutes 59 seconds.** The time is `uv sync` compiling the brahe extension.

That build is then thrown away. uv builds the project under PEP 517 **build isolation**, in a temporary target directory it cannot share with the restored cargo cache, so it pays a full cold compile. The `Build extension` step immediately after recompiles the same extension *warm* in about two minutes. On the coverage leg the discard is explicit — that step opens with `cargo llvm-cov clean --workspace`.

So the job compiles the extension twice, and the slow one is the one nothing uses.

## Why both changes are needed

`--no-install-project` alone does not work. With the project deliberately absent after sync, `uv run` notices and installs it, relocating the cost instead of removing it. Measured directly:

| Command | Result |
|---|---|
| `uv run --frozen` | rebuilds — **126.16 s** |
| `uv run --no-sync` | **0.04 s** |

`--frozen` only pins the lockfile; `--no-sync` is what skips environment sync. Setting `UV_NO_SYNC` at the job level covers every `uv run` in the job rather than requiring each call site to remember.

## Verification

Clean room, cold cache, exactly the sequence the workflow now runs:

| Step | Time |
|---|---|
| `uv sync --group test --all-extras --frozen --no-install-project` | 6.9 s |
| Build extension (`uv pip install -e .`) | 126.6 s |
| `uv run` under `UV_NO_SYNC=1` | 0.16 s — no rebuild |
| Full suite | **3,532 passed, 2 xfailed** |

Expected saving is roughly 7 minutes per Python leg, on jobs currently running about 14.

### Fixed
- Continuous-integration jobs no longer compile the Rust extension twice. Dependency installation was building it under an isolation that could not reuse the cached compiler output, only for the following step to rebuild it.

## Note to Reviewers

**Missing caches remain safe.** This removes work rather than adding a cache dependency: a cold uv cache still downloads normally, and a cold cargo cache means the single remaining build is cold. Both degrade to current performance, never to failure.

**No matrix combination can reach the tests unbuilt.** The two build steps are gated on `A && B` and `!A || !B`, exact complements, so exactly one always runs.

**One deliberate behaviour change.** If a build step ever failed, `uv run` will now surface an `ImportError` rather than silently rebuilding and masking it. That seems the better failure mode, but it is a change worth being aware of.

**Coverage is protected, not just unaffected.** On the coverage leg, `maturin develop` builds under the `cargo llvm-cov` environment. Previously nothing prevented `uv run` from re-syncing and replacing that instrumented build with an uninstrumented one, which would have quietly understated Rust coverage. Today's logs show one build, so it was not happening; `UV_NO_SYNC` makes that a guarantee rather than a coincidence. Worth confirming Codecov reports unchanged numbers on this PR before merging.

## Scope: every workflow with the pattern

The redundant build was not confined to the Python matrix. All eight `uv sync` sites that precede an explicit extension build now skip it:

| Workflow | Notes |
|---|---|
| `test_python.yml` | Largest win — multiplied across the matrix |
| `test_examples.yml` | The one pre-build `uv run` seeds the Celestrak cache and imports no brahe module |
| `docs_only.yml`, `update_docs.yml` | Build immediately follows the sync |
| `pr_tests.yml` (`build-docs`) | Same pattern |
| `warm_data_cache.yml` | Same pattern |
| `integration_tests.yml` | Cleans right after syncing, then rebuilds instrumented — the discarded compile was entirely dead, and the job runs inside a retry budget |
| `release.yml` (`generate-stubs`) | Needs the flag but **not** `UV_NO_SYNC`: `generate_stubs.sh` calls `.venv/bin/...` directly rather than through `uv run` |

Each was checked for a `uv run` that needs `brahe` *before* its build step; none has one.

Deliberately **not** changed: `release.yml` still runs `uv sync --group dev` without `--frozen`, unlike every other workflow. That reproducibility question is real but separate from this performance fix, and conflating them would make both harder to reason about.

### DIFF
--- a/.github/workflows/docs_only.yml
+++ b/.github/workflows/docs_only.yml
@@ -30,6 +30,10 @@ jobs:
     env:
       RUST_TOOLCHAIN: stable
       BRAHE_NETWORK_MODE: offline
+      # The project is installed by the explicit build step below, not by
+      # `uv sync`. Without this, `uv run` would notice it missing and rebuild
+      # it, relocating the cost rather than removing it.
+      UV_NO_SYNC: "1"
     steps:
       - uses: actions/checkout@v7.0.1
         with:
@@ -68,7 +72,7 @@ jobs:
         run: uv python install 3.14
 
       - name: Install dependencies
-        run: uv sync --group docs --all-extras --frozen
+        run: uv sync --group docs --all-extras --frozen --no-install-project
 
       # Build Python extension
       - name: Build Python extension
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -132,6 +132,10 @@ jobs:
         env:
             TEST_SPACETRACK_USER: ${{ secrets.TEST_SPACETRACK_USER }}
             TEST_SPACETRACK_PASS: ${{ secrets.TEST_SPACETRACK_PASS }}
+            # The project is installed by the explicit build step below, not by
+            # `uv sync`. Without this, `uv run` would notice it missing and
+            # rebuild it, relocating the cost rather than removing it.
+            UV_NO_SYNC: "1"
         steps:
             -   name: Set up Rust toolchain
                 uses: actions-rust-lang/setup-rust-toolchain@v1.17.0
@@ -185,7 +189,11 @@ jobs:
                     key: star-catalogs-v1
 
             -   name: Install dependencies
-                run: uv sync --group test --all-extras --frozen --python ${{ matrix.python-version }}
+                # Dependencies only: uv builds the project under PEP 517
+                # isolation, in a temporary target directory it cannot share
+                # with the restored cargo cache, so letting it do so costs a
+                # full cold compile that the build step below then discards.
+                run: uv sync --group test --all-extras --frozen --no-install-project --python ${{ matrix.python-version }}
 
             -   name: Install cargo-llvm-cov
                 if: matrix.python-version == '3.10'
--- a/.github/workflows/pr_tests.yml
+++ b/.github/workflows/pr_tests.yml
@@ -147,6 +147,10 @@ jobs:
     if: needs.check-skip.outputs.should_skip == 'false'
     env:
       RUST_TOOLCHAIN: stable
+      # The project is installed by the explicit build step below, not by
+      # `uv sync`. Without this, `uv run` would notice it missing and rebuild
+      # it, relocating the cost rather than removing it.
+      UV_NO_SYNC: "1"
     steps:
       - uses: actions/checkout@v7.0.1
 
@@ -174,7 +178,7 @@ jobs:
         run: uv python install 3.14
 
       - name: Install dependencies
-        run: uv sync --group docs --all-extras --frozen
+        run: uv sync --group docs --all-extras --frozen --no-install-project
 
       - name: Build Python extension
         run: uv pip install -e ".[plots]"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -135,7 +135,11 @@ jobs:
 
     - name: Install package and dependencies
       run: |
-        uv sync --group dev
+        # Dependencies only; the line below installs the project. Letting uv
+        # sync build it costs a full cold compile under PEP 517 isolation that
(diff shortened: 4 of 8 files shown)


duncaneddy:
Reviewed by Codex against the installed tool versions (uv 0.11.23, maturin 1.15.0, cargo-llvm-cov 0.8.7) rather than from documentation. No blocking findings.

Verified:

- **Build coverage is exhaustive.** The two build gates are exact complements, and the test gates repeat the same conditions, so no leg can test without building. Confirmed for the edge cases too: if `coverage-python-version` is `none` or absent from `python-versions`, every leg takes the standard path.
- **`UV_NO_SYNC` does not break maturin.** `maturin develop --uv` installs into the environment explicitly, independent of `uv run`'s implicit sync, as does `uv pip install -e .`.
- **No locked dependency is skipped.** The explicit `uv sync --group test --all-extras --frozen` still runs on every leg; `--no-install-project` excludes only the project itself, and `UV_NO_SYNC` affects only later `uv run` calls.
- **Coverage is preserved.** `cargo llvm-cov show-env` supplies the instrumentation the report step consumes, and `UV_NO_SYNC` is what now prevents pytest's `uv run` from replacing the instrumented build. Codecov paths and gating are unchanged.
- **Both callers are compatible.** Neither `pr_tests.yml` nor `unit_tests.yml` passes anything that conflicts with the new flag or env var.

One follow-up noted, deliberately out of scope here: the same redundant build exists in `test_examples.yml`, `docs_only.yml`, `update_docs.yml`, and `integration_tests.yml`. `integration_tests.yml` is the most interesting of those — it cleans immediately after syncing and then rebuilds with instrumentation, so it wastes the build in the same explicit way the coverage leg did, and it sits on the release path where the job also has a retry budget to fit inside.

duncaneddy:
Re-reviewed by Codex after the expansion from 1 file to 8, including an `actionlint` pass. No findings.

Confirmed per workflow that nothing can reach a project-dependent command before its build:

- **`test_python.yml`** — build paths are exhaustive across the matrix; nothing project-dependent runs before them.
- **`test_examples.yml`** — the Celestrak seed is the only pre-build `uv run`, and its *complete import graph* was traced: `seed_celestrak_cache.py` imports stdlib plus `scripts/celestrak_snapshots.py`, which uses only `pathlib`. Neither reaches `brahe`.
- **`docs_only.yml`, `update_docs.yml`, `pr_tests.yml`** — the extension install is unconditional and immediately follows the sync; stub generation, `properdocs`, and `mike` all run after it. `UV_NO_SYNC` does not remove the already-synced `docs` group.
- **`warm_data_cache.yml`** — the new `env` block is correctly indented under the `warm-cache` job rather than another job or step, and `uv pip install -e ".[all]"` precedes every warm step that imports brahe.
- **`integration_tests.yml`** — the matrix is 3.10 and 3.14 and both are covered: 3.10 runs `maturin develop` before pytest inside *every* retry attempt, 3.14 runs `uv pip install -e ".[all]"` before its retry-wrapped pytest. The job-level env applies consistently through `nick-fields/retry`, and `UV_NO_SYNC` suppresses only automatic synchronization, not the explicit installs.
- **`release.yml`** — the reasoning for omitting `UV_NO_SYNC` holds: `uv sync --group dev --no-install-project` provides `pyo3-stubgen`, `uv pip install -e .` provides a working brahe, and `generate_stubs.sh` contains no `uv run` at all.

All eight parse cleanly and the job-level env blocks are attached to the intended jobs.

PR #508, opened 2026-09-01
Add unified ReferenceFrame type and object/frame registries

duncaneddy:
# Pull Request

## Description

Introduces the unified `ReferenceFrame` type spanning celestial, orbit-relative, and body/sensor frames, plus the object and frame registries that anchor them, and generalizes the frame-router functions (`rotation_frame_to_frame`, `position_frame_to_frame`, `state_frame_to_frame`, and their batch forms) to resolve any `ReferenceFrame` through those registries alongside the existing `CelestialFrame` router. Stacked on `de/celestial-frame` (#507).

### Added

- Unified `ReferenceFrame` type spanning celestial (`CelestialFrame`), orbit-relative (`RTN`, `LVLH`, `NTW`, `TNW`, `PQW`, `EQW`, `SEZ`, `VNC`, `NSW`), and body/sensor (`BodyFrame`) frames, bound to a string `ObjectId` rather than a NAIF or NORAD ID; `ReferenceFrame::body`/`ReferenceFrame::orbit_relative` construct the unbound label form used before an object identity is known.
- Global object registry (`register_object`, `register_object_from_naif`, `registered_objects`, `unregister_object`, `clear_object_registry`) storing named ephemeris providers — a callable or an `OrbitTrajectory` — that anchor orbit-relative and body frames.
- Global frame registry (`register_frame`, `unregister_frame`, `clear_frame_registry`) for body/sensor orientation chains, backed by the `OrientationProvider` trait (constant attitudes, `CallbackOrientation`) and `with_numerical_rates` (positive, finite step required) for explicitly deriving angular velocity by central-differencing a rotation-only provider. This registry now also backs `register_custom_frame`/`unregister_custom_frame`, absorbing `CelestialFrame::BodyFixedCustom` storage into the same table without changing its observable behavior. Validation and insertion happen under a single write-lock acquisition, so two concurrent `register_frame` calls cannot race to co-create a parent-chain cycle.
- `rotation_frame_to_frame`, `position_frame_to_frame`, `state_frame_to_frame`, and their batch forms generalized to accept a `CelestialFrame` or a `ReferenceFrame`, resolving orbit-relative and body frames through the frame and object registries and joining cross-root chains through the celestial router.
- `OEM::register_for` — registers a CCSDS OEM ephemeris as an object in the global object registry in one call.
- `omega_rtn` — angular velocity of the RTN frame relative to ECI, extracted as a standalone public function.
- Python bindings for `ReferenceFrame` (including a `celestial` constructor wrapping `CelestialFrame`), `BodyFrame`, `OrbitRelativeFrameKind`, `OrbitRelativeFrameVariant`, and every registry function above.

## Note to Reviewers

`just check` passes (full gate: Rust + Python tests, clippy/format/stubs/docs). Coverage report reviewed for all new/changed files (`src/frames/frame.rs`, `orientation.rs`, `registry.rs`, `object_registry.rs`, `graph.rs`, `transform.rs`, `custom.rs`, and the `OEM::register_for` addition in `src/ccsds/interop.rs`); gaps that were reachable production code (not doctest-only or mock-only) got focused tests, closing frame.rs/registry.rs/custom.rs/transform.rs to 100% line coverage and graph.rs/orientation.rs to >96%. Python coverage judged by parity tests per the project's known llvm-cov blind spot on `crates/brahe-py` binding lines. Two prior codex round-1 findings previously left as documented-not-fixed are now resolved: `register_frame`'s validate-then-insert race (single write-lock transaction, replacing the doc caveat) and the Python `ReferenceFrame` binding's 1:1-mirror gap (typed `OrbitRelativeKind`/`OrbitRelativeVariant` bindings replace string arguments to `orbit_relative`; a `celestial` constructor lets Python construct the `Celestial` variant directly).


crates/brahe-py/src/frames.rs
```diff
@@ -3278,9 +3293,13 @@ impl PyCelestialFrame {
...
 /// of those frames still queries SPK.
 ///
+/// `from_frame`/`to_frame` accept a `CelestialFrame` or a `ReferenceFrame`: a
+/// registered body/sensor frame, or a bound orbit-relative frame (e.g.
+/// `ReferenceFrame.RTN("SC")`), resolved by walking the frame registry.
+///
 /// Args:
-///     from_frame (CelestialFrame): Source reference frame
-///     to_frame (CelestialFrame): Target reference frame
+///     from_frame (CelestialFrame or ReferenceFrame): Source reference frame
```
duncaneddy: Should this type just be ReferenceFrame (e.g. since shouldn't passing an OrbitRelative or `BodyFixedFrame` also work? (Ultimately it's any thing that supports the `Into<ReferenceFrame>` type in rust?

crates/brahe-py/src/frames.rs
```diff
@@ -3519,19 +3553,1216 @@ fn py_position_frame_to_frame<'py>(
...
+
+    fn __repr__(&self) -> String {
+        format!("OrbitRelativeKind.{}", self.kind)
+    }
+}
+
+/// Rotating vs. quasi-inertial snapshot variant of a local orbital frame.
+///
+/// - **Rotating**: True local orbital frame, rotating with the orbit.
+/// - **Inertial**: Quasi-inertial frame frozen at each evaluation time.
```
duncaneddy: Could you add some more clarification here. Make it clear (provided the implementation backs it up) that this is just defining instantaneous frame axes based on the OrbitRelative aspects, but not accounting for the rotation of the frame to get a quasi-inertial frame

crates/brahe-py/src/frames.rs
```diff
@@ -3519,19 +3553,1216 @@ fn py_position_frame_to_frame<'py>(
...
+///     frame (CelestialFrame): The celestial frame `provider`'s states are expressed in
+///
+/// Raises:
+///     BraheError: If `provider` is an `OrbitTrajectory` whose native state dimension is not 6
+///     TypeError: If `provider` is not an `OrbitTrajectory` or a callable
+///     ValueError: If `provider` is an `OrbitTrajectory` using the Keplerian representation
+///         rather than Cartesian position/velocity
+///
+/// Returns:
+///     None: The object is registered in the global object registry
```
duncaneddy, line 4742: Might be worth noting in the help strings the interoperability with CCSDS OEM/OMM/OPM files (or how to coerce them)

crates/brahe-py/src/relative_motion.rs
```diff
@@ -99,6 +99,36 @@ fn py_rotation_eci_to_rtn<'py>(
...
+/// respect to the Earth-Centered Inertial (ECI) frame, expressed in RTN axes.
+///
+/// The RTN frame rotates about its cross-track axis at the orbital true-anomaly rate
+/// `f_dot = |r x v| / r^2`, so the angular velocity is `[0, 0, f_dot]`.
+///
+/// Args:
+///     x_eci (numpy.ndarray or list): 6D state vector in the ECI frame [x, y, z, vx, vy, vz] (m, m/s), shape (6,)
+///
+/// Returns:
+///     numpy.ndarray: Angular velocity of the RTN frame relative to ECI, expressed in RTN axes (rad/s), shape (3,)
```
duncaneddy, line 117: Should be clear about which frame the RTN rotation is expressed in (ECI or RTN), from reading it my understanding is it would be RTN (and would need to be transformed by by the rtn_to_eci rotation matrix to get it in the ECI frame.

docs/learn/frames/frame_graph.md
```diff
@@ -0,0 +1,121 @@
...
+- **Body**: an object-local frame with no global transformation &mdash; a spacecraft body frame, a sensor, an actuator, or an instrument.
+
+Orbit-relative and body frames carry an object identity, a plain string (e.g. `"LRO"`, `"2024-123A"`) kept separate from NAIF or NORAD IDs. Constructing one through a family method, `ReferenceFrame.RTN("SC")` or `ReferenceFrame.CSS("SC", "1")`, binds it to that object directly.
+
+## Bound vs. Unbound
+
+A frame is **bound** when it can be evaluated: every `CelestialFrame` is bound by construction, and an orbit-relative or body frame is bound once it carries an object. Constructing one with no object &mdash; `ReferenceFrame.body(None, ...)`/`ReferenceFrame.orbit_relative(..., object=None)` in Python, or converting a bare `BodyFrame`/`OrbitRelativeFrame` in Rust &mdash; gives the unbound form instead: a pure label, useful for parsing a data file's frame column before an object identity is known. `is_bound()` and `object()` report which case a given `ReferenceFrame` is in.
```
duncaneddy: drop excessive em-dash use

docs/learn/frames/frame_graph.md
```diff
@@ -0,0 +1,121 @@
...
+- A **callable**, `Epoch -> rotation matrix`, optionally paired with a second callable returning the angular velocity. This covers time-varying orientations such as a slewing sensor or an articulated appendage.
+
+An orientation chain driven by an attitude ephemeris (AEM), analogous to `OEM.register_for`, ships in a later release; a time-varying orientation is registered as a callable in the meantime.
+
+## Worked Example: A Sun Vector in a Sensor Frame
+
+The example below registers a spacecraft as an object, builds a two-link orientation chain (`SC_BODY` off `GCRF`, then a coarse sun sensor `CSS_1` off `SC_BODY`), and routes the Sun's GCRF position through both links with `position_frame_to_frame`. Body frames share their object's origin exactly &mdash; there is no lever arm between an object's center and the sensor frames mounted on it &mdash; which the example confirms by routing the spacecraft's own position into `CSS_1` and getting the origin back. Querying an unregistered link (`CSS_2`, never registered) raises an error naming the missing frame and the `register_frame` call that would supply it.
```
duncaneddy: Review all docs/writing in this for too much em-dash use.

docs/learn/frames/frame_graph.md
```diff
@@ -0,0 +1,121 @@
...
+        ```
+
+    === "Rust"
+        ```
+        --8<-- "./docs/outputs/frames/body_frame_chain.rs.txt"
+        ```
+
+## The Rates Rule and `with_numerical_rates`
+
+`position_frame_to_frame` needs no angular-velocity data: it only rotates and re-centers. `state_frame_to_frame` is different &mdash; converting a velocity requires the transport term $\omega \times r$ at every non-celestial link in the chain, so every link's provider must supply an angular velocity. A rotation-only provider (a callable given no `omega`) reports `None` for it, and a state transform through that link raises rather than silently dropping the transport term.
```
duncaneddy: non-celestial (which defines the any rotating framing rates internally)

docs/learn/frames/frame_graph.md
```diff
@@ -0,0 +1,121 @@
...
+
+`with_numerical_rates` closes that gap for a provider that has no angular-velocity data of its own: it wraps the provider so a missing rate is derived by central-differencing the rotation matrix over $\pm \text{step}/2$ seconds, using $[\omega]_\times = -\dot{R} R^\mathsf{T}$. In Python, this is the `numerical_rates_step` argument to `register_frame`, and it applies only to a callable provider; a constant attitude already has an exact zero rate and does not need it.
```
duncaneddy: `with_numerical_rates` can be used to add body rates for an `OrientationProvider` that does not include body-rates by default.

docs/learn/frames/index.md
```diff
@@ -36,6 +36,10 @@ Brahe also supports Moon- and Mars-fixed frames for multibody propagation and re
 
 See [Lunar Reference Frames](lunar_frames.md), [Mars Reference Frames](mars_frames.md), and [Synodic Reference Frames](synodic_frames.md) for details, and [Reference Frame Router](frame_transformations.md) for the `CelestialFrame` router that converts between any two frames (including generic NAIF-ID variants for bodies without a dedicated named frame). Central-body propagation is covered in [Cislunar and Lunar Propagation](../orbit_propagation/numerical_propagation/cislunar_lunar_propagation.md) and [Propagation Around Other Central Bodies](../orbit_propagation/numerical_propagation/other_central_bodies.md).
 
+### Object-Scoped Frames
```
duncaneddy: Rename: `Object-based Frames`

docs/learn/frames/index.md
```diff
@@ -36,6 +36,10 @@ Brahe also supports Moon- and Mars-fixed frames for multibody propagation and re
...
+
+`ReferenceFrame` extends `CelestialFrame` to frames scoped to a specific object rather than a body center: orbit-relative frames (`RTN`, `LVLH`, ...) and spacecraft body/sensor/actuator frames (`SC_BODY`, `CSS_1`, ...), registered against an object identity rather than evaluated from an epoch alone. See [Frame Graph](frame_graph.md) for registering objects and orientation chains, and for the generalized `rotation_frame_to_frame`/`position_frame_to_frame`/`state_frame_to_frame` functions that accept either a `CelestialFrame` or a `ReferenceFrame`.
```
duncaneddy: This is confusing because it mentions `CelestialFrame` which is _not_ object relative. I'd just mention Celeslestial frames or rewrite as:

`ReferenceFrame` enables defining reference frames for non-celestial objects: orbit-relative frames (`RTN`, `LVLH`, ...) defined with respect to a satellites' orbital motion or spacecraft body/sensor/actuator frames (`SC_BODY`, `CSS_1`, ...) defined with respect to a spacecraft's orientation.

examples/frames/body_frame_chain.py
```diff
@@ -0,0 +1,73 @@
...
+"""
+
+import numpy as np
+
+import brahe as bh
+
+bh.clear_frame_registry()
+bh.clear_object_registry()
+
+# A ReferenceFrame::Body or ReferenceFrame::OrbitRelative frame is a pure label until it is
```
duncaneddy: This appears to be rust-style class naming in a python-based example. Would be better to use the right comment format.

duncaneddy: Also unclear what a "pure label is" or what it means to be bound

crates/brahe-py/src/frames.rs
```diff
@@ -3519,19 +3553,1216 @@ fn py_position_frame_to_frame<'py>(
...
+///     ```python
+///     import brahe as bh
+///
+///     bh.clear_object_registry()
+///     bh.register_object("SC", lambda epc: [bh.R_EARTH + 500e3, 0.0, 0.0, 0.0, 7600.0, 0.0], bh.CelestialFrame.GCRF)
+///     bh.clear_object_registry()
+///     ```
+#[pyfunction]
+#[pyo3(name = "register_object")]
+fn py_register_object(
```
duncaneddy, line 4754: Would the be better named as "`register_object_state_provider`" to make it very literal?

examples/frames/body_frame_chain.py
```diff
@@ -0,0 +1,73 @@
...
+# attitude (Quaternion, RotationMatrix, EulerAngle, EulerAxis) registers
+# directly; orientation chains driven by an attitude ephemeris ship in a
+# later release.
+q_body = bh.Quaternion(1.0, 0.0, 0.0, 0.0)
+q_css = bh.EulerAxis(
+    np.array([0.0, 1.0, 0.0]), 0.7, bh.AngleFormat.RADIANS
+).to_quaternion()
+bh.register_frame(bh.ReferenceFrame.SC_BODY("SC"), bh.CelestialFrame.GCRF, q_body)
+bh.register_frame(
+    bh.ReferenceFrame.CSS("SC", "1"), bh.ReferenceFrame.SC_BODY("SC"), q_css
```
duncaneddy, line 48: Might be worth being clear that all these Body Reference frame enums are explicitly SANA-defined standard values (https://sanaregistry.org/r/spacecraft_body_reference_frames/)

src/ccsds/interop.rs
```diff
@@ -477,6 +537,78 @@ mod tests {
...
+    #[serial]
+    fn test_oem_register_for_itrf() {
+        // Covers the ITRF arm of register_for's OrbitFrame -> CelestialFrame
+        // mapping (test_oem_register_for above only exercises GCRF). Built
+        // inline (rather than from an on-disk asset) with a handful of rows
+        // taken from OEMExample5.txt, since register_for only needs a
+        // parseable single-segment OEM.
+        clear_object_registry();
+
+        let content = "CCSDS_OEM_VERS = 2.0\n\
```
duncaneddy: Consider making test data a file (instead of hard-coding lines)

src/frames/custom.rs
```diff
@@ -92,11 +83,15 @@ pub fn register_custom_frame<R>(key: u32, rotation: R, omega: Option<Box<CustomF
...
     R: Fn(Epoch) -> Result<SMatrix3, BraheError> + Send + Sync + 'static,
 {
-    CUSTOM_FRAMES.write().unwrap().insert(
-        key,
-        CustomFrameEntry {
-            rotation: Arc::new(rotation),
-            omega: omega.map(Arc::from),
+    let provider: Arc<dyn OrientationProvider> = match omega {
+        Some(omega) => Arc::new(CallbackOrientation::new(rotation, Some(omega))),
+        None => Arc::new(CallbackOrientation::new(rotation, None).with_numerical_rates(2.0)),
```
duncaneddy: This would appear to register all custom frames with a numerically-derived rate if none is provided instead of just setting "None" and erroring on a transform if none was set.

src/frames/frame/mod.rs
```diff
@@ -0,0 +1,1398 @@
+/*!
```
duncaneddy, line 1: This file is super big, might make sense to break up into smaller constituent component / trait files to make it more maintainbable and reviewable

src/frames/frame.rs
```diff
@@ -0,0 +1,1398 @@
...
+    /// Radial / transverse (along-track) / normal (cross-track). SANA: RSW.
+    RTN,
+    /// Normal / tangential / cross-track.
+    NTW,
+    /// Tangential / normal / cross-track.
+    TNW,
+    /// Perifocal. SANA-registered only as an inertial-snapshot frame.
+    PQW,
+    /// Equinoctial. SANA-registered only as an inertial-snapshot frame.
+    EQW,
```
duncaneddy: Do we properly validate that these can only be set inertial-snapshot only?

src/frames/frame.rs
```diff
@@ -0,0 +1,1398 @@
...
+///
+/// # Examples
+///
+/// ```rust
+/// use brahe::frames::OrbitRelativeVariant;
+///
+/// assert_eq!(OrbitRelativeVariant::Rotating.to_string(), "rotating");
+/// ```
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum OrbitRelativeVariant {
```
duncaneddy: I might prefer to be more explicit than short-hand to make code more readable:

- `OrbitRelativeVariant` -> `OrbitRelativeFrameVariant`
- `OrbitRelativeKind` -> `OrbitRelativeFrameKind`

src/frames/graph.rs
```diff
@@ -0,0 +1,1401 @@
...
+    epc: Epoch,
+    x: SVector6,
+) -> Result<SVector6, BraheError> {
+    if from == to {
+        return Ok(x);
+    }
+    let resolved_from = resolve_orientation(from, epc, true)?;
+    let resolved_to = resolve_orientation(to, epc, true)?;
+    let omega_from = chain_rate(from, &resolved_from)?;
+    let omega_to = chain_rate(to, &resolved_to)?;
```
duncaneddy, line 173: we really want to make sure that this rate is correct, specially with multiple levels of chaining and offsets and is grounded in a proper physics-based source / reference.

src/frames/graph.rs
```diff
@@ -0,0 +1,1401 @@
...
+                CelestialFrame::LCI
+            } else if center == NAIFId::Mars.id() {
+                CelestialFrame::MCI
+            } else if center == NAIFId::EarthMoonBarycenter.id() {
+                CelestialFrame::EMBI
+            } else if center == NAIFId::SolarSystemBarycenter.id() {
+                CelestialFrame::SSBI
+            } else {
+                CelestialFrame::BodyCenteredICRF(center)
+            }
```
duncaneddy, line 514: I believe we also support a native EME2000 implementation, but I don't see that in the branch options here.

src/frames/mod.rs
```diff
@@ -13,20 +13,35 @@ pub mod custom;
 pub mod eci_ecef;
 pub mod emb;
 pub mod eme_2000;
+pub mod frame;
 pub mod gcrf_itrf;
+// Not `pub`: chain resolution is an internal seam of the frames router; its
+// results reach callers through `transform`'s public router functions.
```
duncaneddy: Just leave simple comment of "chain resolution is internal machinery"

src/frames/object_registry.rs
```diff
@@ -0,0 +1,443 @@
...
+        if dim != 6 {
+            return Err(BraheError::Error(format!(
+                "DStateAdapter requires a 6-dimensional state provider, got dimension {dim}"
+            )));
+        }
+        Ok(Self {
+            provider: Box::new(provider),
+        })
+    }
+}
```
duncaneddy, line 239: I'm not sure (don't think) we should/need to introduce this beyond DStateProvider.

Ideally we can just pass in DStateProvider. If the dimension is more than 6, the position/velocity state vectors are guaranteed by DStateProvider to be the first 6 elements. So more values are fine (just not used or passed through).

If less than 6 that should/could be an error.

src/frames/orientation.rs
```diff
@@ -0,0 +1,547 @@
...
+/// use brahe::attitude::Quaternion;
+/// use brahe::frames::OrientationProvider;
+/// use brahe::time::{Epoch, TimeSystem};
+///
+/// let q = Quaternion::new(1.0, 0.0, 0.0, 0.0);
+/// let epc = Epoch::from_date(2024, 1, 1, TimeSystem::TAI);
+/// assert_eq!(OrientationProvider::quaternion(&q, epc).unwrap(), q);
+/// assert!(OrientationProvider::coverage(&q).is_none());
+/// ```
+pub trait OrientationProvider: Send + Sync {
```
duncaneddy, line 49: Should we also have a branch for returning:
- EulerAngle (with passed in EulerAngleOrder)
- EulerAxis

src/frames/orientation.rs
```diff
@@ -0,0 +1,547 @@
...
+        }
+
+        let half_step = self.step / 2.0;
+        let r = self.inner.rotation_matrix(epoch)?.to_matrix();
+        let r_plus = self.inner.rotation_matrix(epoch + half_step)?.to_matrix();
+        let r_minus = self.inner.rotation_matrix(epoch - half_step)?.to_matrix();
+        let r_dot = (r_plus - r_minus) / self.step;
+
+        // [omega]× = -Ṙ Rᵀ for r_frame = R r_parent; extract the vector
+        // from the skew-symmetric part.
```
duncaneddy: Reference?

src/frames/orientation.rs
```diff
@@ -0,0 +1,547 @@
...
+    }
+
+    fn rotation_matrix(&self, epoch: Epoch) -> Result<RotationMatrix, BraheError> {
+        self.inner.rotation_matrix(epoch)
+    }
+}
+
+/// Extension trait adding [`Self::with_numerical_rates`] to every
+/// [`OrientationProvider`].
+pub trait OrientationProviderExt: OrientationProvider + Sized {
```
duncaneddy: Should this be an extension trait or just part of the base trait?

Given it's implemented for all OrientationProvider (and can just be not used anyways) seems like we could simplify by just combinging with `OrientationProvider`

duncaneddy:
All review comments addressed in `0c7d9457`, `3cd3d720`, `962f1ccb`, `1221393b` (full gate green: 5457 Rust tests + 1010 doctests, 3570 Python tests, clippy/fmt/stubs/docs clean). Resolution map:

**API changes**
- `with_numerical_rates` now validates a positive, finite step in Rust too and returns `Result` (rejection tests for 0/-1/NaN/±inf); it moved onto `OrientationProvider` itself and `OrientationProviderExt` is deleted.
- `OrbitRelativeKind`/`OrbitRelativeVariant` renamed to `OrbitRelativeFrameKind`/`OrbitRelativeFrameVariant` everywhere (Rust, Python, stubs, docs).
- `DStateAdapter` is now internal (`pub(crate)`) and gone from the public surface; it accepts `state_dim() >= 6` using the leading six elements and errors below six. The Python binding adapts trajectories with its own internal equivalent.
- `OrientationProvider` gained `euler_angle(epoch, order)` / `euler_axis(epoch)` default methods.
- `icrf_aligned_inertial` has an explicit `EME2000 -> GCRF` arm (previously handled equivalently by the fallback), with a test pinning EME2000-declared objects against GCRF-declared results.

**Physics grounding**
- The ω chain-composition step now documents the angular-velocity addition theorem (expressed-in-frame form) citing Markley & Crassidis (2014) and Schaub & Junkins (2018), and `[ω]× = −Ṙ Rᵀ` cites Poisson's kinematic equation. A new two-link, both-links-time-varying finite-difference test empirically pins multi-level chained rates end to end.

**Docs/examples**
- Docs and every constructor docstring now state only RTN has an axes derivation today (others error, tracked in #452). Transform docstrings read `CelestialFrame | ReferenceFrame`. Inertial-variant semantics (instantaneous snapshot axes, zero rates) clarified. `omega_rtn` states its output is expressed in RTN. `register_object` docs cover the CCSDS paths (`OEM.register_for`, `SGPPropagator.from_gp_record`/`from_omm_elements`, `KeplerianPropagator.from_eci` — all verified real). Em-dash sweep done on all new prose. `index.md` heading/paragraph per suggested wording. Example comments fixed (Python naming, label-vs-bound explained). `BodyFrame` families cite the SANA spacecraft body reference frames registry. Inline OEM test payload moved to `test_assets/ccsds/oem/OEM-single-segment-itrf.txt` (the existing ITRF example file is rejected by the parser due to its trailing-`Z` creation date — pre-existing quirk, worth a follow-up). `frame.rs` split into `frame/{mod,object_id,orbit_relative,body}.rs` with the public surface unchanged.

**Answered rather than changed (flag if you disagree)**
1. `register_object` name kept: it registers the object identity (name + ephemeris + frame), and the family stays consistent (`register_object_from_naif`, `unregister_object`, `registered_objects`, `OEM.register_for`).
2. `register_custom_frame`'s derive-rates-by-default is deliberate: it preserves that function's pre-existing documented contract. The new `register_frame` API is the strict one (rate-less providers report `None`; state transforms error unless `with_numerical_rates` is opted in).
3. EQW/PQW inertial-only is enforced on every construction path; direct struct-variant construction can bypass it (no per-field variant visibility in Rust) but the bypass is inert — the resolver rejects all non-RTN kinds, so an invalid value can only error, never mistransform. #452 must re-check the combo when more kinds land.

Codex P1 (lock atomicity) was reviewed against a pre-fix commit; the current code validates and inserts under a single write guard with a cycle regression test.

PR #526, opened 2026-09-05
Add equinox-based MOD and TOD Earth frames

duncaneddy:
## Description

Adds equinox-based Earth mean-of-date (MOD) and true-of-date (TOD) reference frames. A new `src/frames/equinox.rs` evaluates the classical chain `[ITRF] = W R3(GAST) N P B [GCRF]` on the IAU 2000 precession and IAU 2000B nutation basis (`iauPn00b`), the same model basis the existing CIO-based GCRF/ITRF chain uses through `iauXys00b`. IERS dX/dY celestial pole offsets enter the nutation step after conversion to dPsi/dEps per SOFA cookbook Section 5.4, and Greenwich apparent sidereal time is computed as ERA minus the equation of the origins from the same NPB matrix, so converting GCRF to ITRF through TOD agrees with the direct transformation at the microarcsecond level. MOD and TOD are rooted at GCRF with the frame bias applied inside the bias-precession matrix (NASA TP-20220014814 Fig. 3, cookbook Appendix p. A4). This is the first PR of a stack; later PRs replace the trajectory-level `OrbitFrame` with `ReferenceFrame`, map the CCSDS `TOD`/`MOD_EARTH`/`TOD_EARTH` tokens to the native frames, and add TEME.

### Added
- `CelestialFrame::MOD` and `CelestialFrame::TOD` in the reference frame router, including `from_string`/`Display` support and the corresponding `CelestialFrame.MOD`/`CelestialFrame.TOD` Python attributes.
- Equinox-based Earth frame transformations `bias_precession_iau2000`, `nutation_iau2000b`, `gast_rotation_iau2000b`, and the `rotation_`/`position_`/`state_` functions with batch forms for GCRF/MOD, MOD/TOD, GCRF/TOD, and TOD/ITRF in both directions (Rust and Python).
- `MOD` and `TOD` frame choices in the `brahe transform` CLI commands.
- Learn page and API reference for the equinox-based frames with six standalone examples in Python and Rust.

## Related Issue
Related to #520 (the TOD interpretation fix for the OPM example lands in a later PR of this stack).

## Note to Reviewers
[...]

PR #527, opened 2026-09-05
Replace OrbitFrame with ReferenceFrame in trajectories

### LINKED ISSUE #520: Incorrect frame transformation in OPM maneuver propagation example
In the [maneuver propagation](https://docs.brahe.space/latest/learn/ccsds/opm.html?h=tod#maneuver-propagation) example, the spacecraft state retrieved from the [OPM](https://github.com/duncaneddy/brahe/blob/main/test_assets/ccsds/opm/OPMExample2.txt) is expressed in TOD (True of Date) frame.

Afterwards, it is converted to ECI (GCRF) as follows:
```python
# Extract initial state (OPM is in TOD frame, convert to ECI)
state_eci = bh.state_ecef_to_eci(opm.epoch, opm.state)
```

However, `state_ecef_to_eci` transforms the input state from ECEF (ITRF) to ECI (GCRF), not from TOD to GCRF. TOD is a pseudo-inertial frame, not a body-fixed one. It is similar to the CIRF in the GCRF -> CIRF -> TIRF -> ITRF chain.

As far as I know, Brahe does not implement the equinox-based transformation EME2000 -> MOD -> TOD which would be needed in this example.

See e.g. CIO-based vs equinox-based transformations in [Orekit docs](https://www.orekit.org/site-orekit-development/architecture/frames.html).

duncaneddy:
## Description

Replaces the trajectory-level `OrbitFrame` tag with the router's `ReferenceFrame`, so a trajectory, a propagator output, or a CCSDS ephemeris can be declared in any frame the reference frame router evaluates: every `CelestialFrame` (GCRF, ITRF, EME2000, MOD, TOD, the lunar and Mars frames, ...) and any bound orbit-relative or body frame. Frame conversion collapses onto one `to_frame` method that converts Keplerian samples to Cartesian with the GM of the frame's center and routes every sample through `state_frame_to_frame`; the named `to_eci`/`to_gcrf`/`to_ecef`/`to_itrf`/`to_eme2000` methods remain as wrappers, and the state-provider accessors delegate to `state_in_frame`. A crate-internal `celestial_root` resolves any bound frame to its celestial root without evaluating rotations, so `state_bci`/`state_bcbf` work for orbit-relative trajectories. `ReferenceFrame` compares equal to the `CelestialFrame` it wraps in both Rust and Python. Second PR of the equinox-frames stack (on #526); the CCSDS TOD/MOD token mapping that closes #520 follows in the next PR.

### Added
- `OrbitTrajectory.to_frame(frame)` (Rust `DOrbitTrajectory::to_frame` / `SOrbitTrajectory::to_frame`) converting a trajectory to any `ReferenceFrame` through the reference frame router, including orbit-relative frames of registered objects.
- `ReferenceFrame == CelestialFrame` comparisons in Rust (`PartialEq` both directions) and Python (`ReferenceFrame.__eq__` accepts a `CelestialFrame`).
- `KeplerianPropagator` and `SGPPropagator` accept any Earth-rooted `ReferenceFrame` as output frame (for example `CelestialFrame.TOD` or a bound `RTN` frame), and Keplerian output is accepted in EME2000 as well as GCRF.

### Changed
- Trajectory, propagator, and CCSDS interop frame fields and arguments are `ReferenceFrame`; constructors accept a `CelestialFrame` or `ReferenceFrame` (`bh.CelestialFrame.ECI` where `bh.OrbitFrame.ECI` was used; `bh.CelestialFrame.LCI` for the former `BodyCenteredInertial(301)`).
- `OrbitTrajectory.frame` returns a `ReferenceFrame` whose string form is the celestial name (`GCRF`, `ITRF`, `LCI`) rather than the former `ECI`/`ECEF`/`BCI(301)` labels.
- Keplerian element trajectories and propagator outputs are valid in ICRF-aligned inertial celestial frames and EME2000; `CelestialFrame.GCRF` is accepted wherever `ECI` was (they are the same frame).
- `plot_trajectory_3d` converts trajectories around any central body into that body's inertial frame with `to_frame` instead of rejecting a mismatched frame.
- `OEM.register_for` accepts any celestial trajectory frame; the Python OEM-from-trajectory path converts through the router, so a non-Earth body-centered trajectory is re-centered instead of rejected.
- `to_keplerian` keeps the trajectory's own frame for element input and, for Cartesian input, produces elements about the frame's center (accepted for ICRF-aligned inertial frames and EME2000; rejected for Earth-fixed, of-date, orbit-relative, and body frames, where it previously converted through GCRF and labeled the result ECI). `state_koe_osc` short-circuits exactly for Keplerian trajectories in any ICRF-aligned frame.
- `covariance_eci`/`covariance_gcrf` error, naming the frame, for trajectory frames without a covariance representation; propagator error messages for the Keplerian-frame and Earth-only rules changed.
- `OrbitalTrajectory` gains a required `to_frame` method (breaking for external implementors). Propagator constructors, `with_output_format`, and stepping can return router errors for frames that cannot be evaluated; `KeplerianPropagator::reset` panics if the output frame can no longer be resolved; `set_initial_conditions` leaves the propagator unchanged when a conversion fails.
- Python `KeplerianPropagator(...)` and `set_initial_conditions` accept `angle_format=None` for Cartesian input, mirroring the Rust `Option<AngleFormat>`; passing `None` with a Keplerian representation now raises `BraheError` instead of `TypeError`.
- `ccsds_ref_frame_to_orbit_frame` is renamed `ccsds_ref_frame_to_reference_frame` and returns a `ReferenceFrame`; the token mapping itself lives in the new `ccsds_ref_frame_to_celestial_frame`, which `OEM::register_for` uses directly.

### Removed
- `OrbitFrame` (Rust `brahe::trajectories::OrbitFrame`, Python `brahe.OrbitFrame`).

### Fixed
- `KeplerianPropagator` with an EME2000 output frame no longer applies a spurious frame rotation in `state_gcrf`, and its ECEF/ITRF output arms no longer disagree.
- Frame conversion of a Keplerian trajectory declared in EME2000 now converts the elements about the correct frame instead of treating them as GCRF elements, matching the point-query accessors.

## Related Issue
Related to #520 (closed by the next PR of this stack).

## Note to Reviewers
Keplerian element trajectories in EME2000 were accepted before this change and remain accepted (elements are realized as Cartesian and rotated); the design spec's D9 row records this allowance. The four patch lines codecov still reports as missing are the `?` early-return regions of infallible calls (`SOrbitTrajectory::new` for Cartesian input, `convert_state_from_spg4_frame` for ECI Cartesian output, and the EME2000 bias rotation), which no input can reach.
The trajectory conversion paths for GCRF, ITRF, and EME2000 are bit-identical to the previous pairwise implementation because the router calls the same functions; tests assert exact equality. Behavior changes are listed above; the largest is the `plot_trajectory_3d` conversion (the former rejection test is rewritten to assert the conversion). The pre-existing `SOrbitTrajectory::from_orbital_data` trait method takes a `ReferenceFrame` directly, so Rust call sites pass `CelestialFrame::ECI.into()`.

### DIFF
--- a/examples/trajectories/orbit_trajectory/orbit_trajectory_from_data.rs
+++ b/examples/trajectories/orbit_trajectory/orbit_trajectory_from_data.rs
@@ -4,7 +4,8 @@
 use brahe as bh;
 use bh::time::Epoch;
 use bh::trajectories::SOrbitTrajectory;
-use bh::trajectories::traits::{OrbitFrame, OrbitRepresentation, OrbitalTrajectory};
+use bh::frames::CelestialFrame;
+use bh::trajectories::traits::{OrbitRepresentation, OrbitalTrajectory};
 use bh::traits::Trajectory;
 use bh::constants::R_EARTH;
 use nalgebra as na;
@@ -35,7 +36,7 @@ fn main() {
     let traj = SOrbitTrajectory::from_orbital_data(
         epochs,
         states,
-        OrbitFrame::ECI,
+        CelestialFrame::ECI.into(),
         OrbitRepresentation::Cartesian,
         None, // Angle Format
         None  // No covariances
--- a/examples/trajectories/orbit_trajectory/orbit_trajectory_from_data.rs
+++ b/examples/trajectories/orbit_trajectory/orbit_trajectory_from_data.rs
@@ -36,7 +36,7 @@ fn main() {
     let traj = SOrbitTrajectory::from_orbital_data(
         epochs,
         states,
-        CelestialFrame::ECI.into(),
+        CelestialFrame::ECI,
         OrbitRepresentation::Cartesian,
         None, // Angle Format
         None  // No covariances
--- a/src/trajectories/traits.rs
+++ b/src/trajectories/traits.rs
@@ -6,6 +6,9 @@
  */
 
 use crate::constants::AngleFormat;
+use crate::frames::{
+    CelestialFrame, ReferenceFrame, iau_rotation_model_ids, icrf_aligned_inertial,
+};
 use crate::time::Epoch;
 use crate::utils::BraheError;
 use nalgebra::{DMatrix, SMatrix};
@@ -29,12 +32,16 @@ pub enum TrajectoryEvictionPolicy {
     KeepWithinDuration,
 }
 
-/// Reference-frame-router frame with ICRF-aligned axes centered on `center`,
-/// used to convert `OrbitFrame::BodyCenteredInertial(center)` trajectory
-/// samples (named frames for the bodies that have them, generic
-/// `BodyCenteredICRF` otherwise).
-pub(crate) fn bci_reference_frame(center: i32) -> crate::frames::CelestialFrame {
-    use crate::frames::CelestialFrame;
+/// Reference-frame-router frame with ICRF-aligned axes centered on `center`
+/// (named frames for the bodies that have them, generic `BodyCenteredICRF`
+/// otherwise).
+///
+/// # Arguments
+/// * `center` - NAIF ID of the frame's center
+///
+/// # Returns
+/// * `CelestialFrame`: ICRF-aligned frame centered on `center`
+pub(crate) fn bci_reference_frame(center: i32) -> CelestialFrame {
     match center {
         399 => CelestialFrame::GCRF,
         301 => CelestialFrame::LCI,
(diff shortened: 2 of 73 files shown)


src/trajectories/traits.rs
```diff
@@ -677,27 +675,44 @@ pub trait OrbitalTrajectory: InterpolatableTrajectory {
...
+    ///
+    /// # Arguments
+    /// * `frame` - Target frame
+    ///
+    /// # Returns
+    /// * `Ok(Self)` - New Cartesian trajectory in `frame`
+    /// * `Err(BraheError)` - If a sample cannot be converted (unbound or
+    ///   unregistered frame, missing ephemeris, or Keplerian elements about
+    ///   a barycenter)
+    fn to_frame(&self, frame: ReferenceFrame) -> Result<Self, BraheError>
```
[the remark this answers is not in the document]
duncaneddy: Added a `# Examples` doctest to `OrbitalTrajectory::to_frame` (commit 9c26738c5); it builds a GCRF trajectory and converts it to ITRF through the router.

examples/trajectories/orbit_trajectory/orbit_trajectory_from_data.rs
```diff
@@ -35,7 +36,7 @@ fn main() {
     let traj = SOrbitTrajectory::from_orbital_data(
         epochs,
         states,
-        OrbitFrame::ECI,
+        CelestialFrame::ECI.into(),
```
duncaneddy: can the .into() be dropped here?

duncaneddy: Yes. `from_orbital_data` already had `where Self: Sized`, so it now takes `frame: impl Into<ReferenceFrame>` like the `new` constructors, and the `.into()` is gone from every call site and doc example.

PR #528, opened 2026-09-06
Map CCSDS TOD frames to native frames and add OPM state_in_frame

### LINKED ISSUE #520: Incorrect frame transformation in OPM maneuver propagation example

duncaneddy:
## Description

Maps the CCSDS `TOD` reference-frame token to the native `CelestialFrame::TOD` frame so that OEM segments, OPM states, and ADM messages declared true-of-date are loaded in their own frame and converted through the reference frame router instead of being rejected or mislabeled. `ReferenceFrame` gains `TryFrom` conversions to and from `CCSDSRefFrame` in both directions, the ADM frame mapping gains `TOD_EARTH` and `MOD_EARTH`, and `OPM::state_in_frame` reads an OPM state vector out in any router frame. The OPM maneuver propagation example uses `state_in_frame(GCRF)` in place of the Earth-fixed conversion reported in #520, and schedules the CCSDS sample message's maneuvers relative to its state epoch so the captured output shows a finite propagation. Third PR of the equinox-frames stack (on #526 and #527).

### Added
- `OPM.state_in_frame(frame)` (Rust `OPM::state_in_frame`) converting the message state vector from its declared `REF_FRAME` to any `CelestialFrame` or `ReferenceFrame` through the reference frame router.
- `TryFrom<&CCSDSRefFrame> for CelestialFrame` (with `TryFrom<&CCSDSRefFrame> for ReferenceFrame` delegating to it) and `TryFrom<&ReferenceFrame> for CCSDSRefFrame` mapping `EME2000`/`J2000`, `GCRF`, the `ITRF` realizations, and `TOD` (the reverse direction writes `ITRF` as `ITRF2000` and rejects `MOD`, which has no ODM token).
- ADM `TOD_EARTH` and `MOD_EARTH` reference-frame tokens map to `CelestialFrame::TOD` and `CelestialFrame::MOD` in both directions.

- `OPM::state_in_frame` rejects messages whose `CENTER_NAME` is not `EARTH`, since every ODM reference frame with a native equivalent is Earth-centered.
- A `TOD` message carrying `REF_FRAME_EPOCH` is loaded as GCRF by rotating its states with the true-of-date axes frozen at that epoch, for OEM trajectories and `OPM::state_in_frame` alike.

### Changed
- OEM segments declared in `TOD` load as `OrbitTrajectory` values whose frame is `CelestialFrame::TOD` through `segment_to_dorbit_trajectory`, `segment_to_sorbit_trajectory`, `TryFrom<&OEM>`, `OEM::register_for`, and Python `OEM.to_trajectories()`, instead of returning an error.
- Python `OEMSegment.add_trajectory` writes trajectories into segments declared in `TOD`, rotating from the trajectory frame through the reference frame router.
- ODM reference-frame tokens without a native trajectory frame report distinct errors: `TEME` and `TDR` have no native frame, orbit-relative tokens (`RTN`, `TNW`, `RSW`) require an object anchor, and unknown tokens are reported as unrecognized.
- Python `OEMSegment.add_trajectory` raises `BraheError` for an unsupported segment frame in both standalone and OEM-attached segments.
- The `opm_maneuver_propagation` examples convert the TOD state with `state_in_frame`, schedule the sample message's maneuvers one hour after the state epoch preserving their spacing, rotate EME2000 delta-V by the frame bias, and rotate RTN delta-V with the propagated state at ignition; the `opm_init_propagator` examples use `state_in_frame` for the ITRF sample; `docs/learn/ccsds/opm.md` and `docs/learn/ccsds/oem.md` describe TOD handling.

### Removed
- `ccsds_ref_frame_to_reference_frame` and `ccsds_ref_frame_to_celestial_frame`; use `ReferenceFrame::try_from(&ccsds_ref_frame)` or `CelestialFrame::try_from(&ccsds_ref_frame)`.

### Fixed
- The OPM maneuver propagation example no longer treats a TOD state vector as ECEF.

## Related Issue
Closes #520

## Note to Reviewers
`state_frame_to_frame` short-circuits when source and target frames are equal, so `state_in_frame` of a TOD message into TOD is bit-identical to the parsed state vector; the tests assert exact equality there and compare TOD to GCRF against `state_tod_to_gcrf`. The OEM tests use `test_assets/ccsds/oem/test.oem` because `OEMExample2.txt` declares `TIME_SYSTEM = MRT`, which the epoch parser rejects.


src/ccsds/interop.rs
```diff
@@ -101,7 +187,7 @@ impl OEM {
             ))
         })?;
 
-        let frame = ccsds_ref_frame_to_reference_frame(&segment.metadata.ref_frame)?;
+        let frame = ReferenceFrame::try_from(&segment.metadata.ref_frame)?;
 
         let mut traj = DOrbitTrajectory::new(6, frame, OrbitRepresentation::Cartesian, None)?;
```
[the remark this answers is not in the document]
duncaneddy: Fixed. A `TOD` segment with `REF_FRAME_EPOCH` is the true-of-date frame frozen at that epoch, so the converters now rotate every sample with the constant `rotation_tod_to_gcrf(REF_FRAME_EPOCH)` and build the trajectory in `GCRF`; segments without `REF_FRAME_EPOCH` still load as native `TOD`. A new `test_tod_epoch.oem` asset asserts the frozen-epoch rotation is used and that it differs from the per-sample of-date rotation.

src/ccsds/opm.rs
```diff
@@ -220,6 +222,40 @@ impl OPM {
...
+    ///
+    /// let opm = OPM::from_file("test_assets/ccsds/opm/OPMExample2.txt").unwrap();
+    /// let x_gcrf = opm.state_in_frame(CelestialFrame::GCRF).unwrap();
+    /// ```
+    pub fn state_in_frame(&self, frame: impl Into<ReferenceFrame>) -> Result<SVector6, BraheError> {
+        let native = ReferenceFrame::try_from(&self.metadata.ref_frame)?;
+        let p = self.state_vector.position;
+        let v = self.state_vector.velocity;
+        let x = SVector6::new(p[0], p[1], p[2], v[0], v[1], v[2]);
+        state_frame_to_frame(native, frame, self.state_vector.epoch, x)
```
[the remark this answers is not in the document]
duncaneddy: Fixed. `state_in_frame` shares the same rule: when `REF_FRAME_EPOCH` is present for a `TOD` message, the source state is rotated to `GCRF` with the true-of-date axes at that epoch and then routed at the state-vector epoch; without it, TOD is evaluated at the state-vector epoch.

src/ccsds/opm.rs
```diff
@@ -220,6 +222,40 @@ impl OPM {
...
+    /// use brahe::frames::CelestialFrame;
+    ///
+    /// let eop = FileEOPProvider::from_default_file(EOPType::StandardBulletinA, true, EOPExtrapolation::Zero).unwrap();
+    /// set_global_eop_provider(eop);
+    ///
+    /// let opm = OPM::from_file("test_assets/ccsds/opm/OPMExample2.txt").unwrap();
+    /// let x_gcrf = opm.state_in_frame(CelestialFrame::GCRF).unwrap();
+    /// ```
+    pub fn state_in_frame(&self, frame: impl Into<ReferenceFrame>) -> Result<SVector6, BraheError> {
+        let native = ReferenceFrame::try_from(&self.metadata.ref_frame)?;
```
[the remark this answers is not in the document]
duncaneddy: Fixed. `OPM::state_in_frame` and the OEM segment converters now validate `CENTER_NAME` and return an error naming the center when it is not Earth, since the native `GCRF`/`EME2000`/`TOD`/`ITRF` frames are Earth-centered. Tests cover `OPM-dummy-moon-EME2000.txt` (Rust and Python).

crates/brahe-py/src/ccsds.rs
```diff
@@ -1648,7 +1631,11 @@ impl PyOEM {
     /// Convert a single OEM segment to an OrbitTrajectory.
     ///
     /// The trajectory contains Cartesian state vectors (position/velocity)
-    /// in the reference frame specified by the segment metadata.
+    /// in the reference frame specified by the segment metadata. A segment
+    /// declaring `REF_FRAME = TOD` together with a `REF_FRAME_EPOCH` names the
+    /// true-of-date axes frozen at that epoch, so its states are rotated into
+    /// GCRF with the rotation evaluated at the frame epoch and the trajectory
+    /// is labelled GCRF.
```
duncaneddy: This comment seems overly focused on the TOD implementation specifics when it's just a general method. It seems that it should just be: "The trajectory contains Cartesian state vectors (position/velocity) in the reference frame specified by the segment metadata"

docs/learn/ccsds/oem.md
```diff
@@ -29,7 +29,7 @@ Parse from file or string, then access header properties, segment metadata, and
 
 ## Converting to OrbitTrajectory
 
-The primary interoperability point for OEM data is conversion to brahe's `OrbitTrajectory`. Each OEM segment maps to a trajectory object, giving you Hermite interpolation at arbitrary epochs within the covered time span:
+The primary interoperability point for OEM data is conversion to brahe's `OrbitTrajectory`. Each OEM segment maps to a trajectory object, giving you Hermite interpolation at arbitrary epochs within the covered time span. A segment with `REF_FRAME = TOD` loads as a trajectory in the TOD frame; use `to_frame` to convert it to GCRF, ITRF, or another supported frame. A segment that pairs `REF_FRAME = TOD` with a `REF_FRAME_EPOCH` names the true-of-date axes frozen at that epoch, which are inertial, so it loads as a GCRF trajectory built with the true-of-date axes at the frame epoch:
```
duncaneddy: I would revert this comment. It's overly TOD focused and not relevant to general trajectory implementation.

docs/learn/ccsds/opm.md
```diff
@@ -66,16 +66,16 @@ Every OPM has a **header** (version, creation date, originator), **metadata** (o
 
 ## Maneuver Propagation
 
-Read OPM maneuvers and apply them as impulsive delta-V events during propagation:
+Read OPM maneuvers and apply them as impulsive delta-V events during propagation. The example message declares its state vector in the TOD frame, and `state_in_frame` converts it to GCRF through the reference frame router before it is handed to the propagator. A message that pairs `REF_FRAME = TOD` with a `REF_FRAME_EPOCH` names the true-of-date axes frozen at that epoch, which are inertial, so `state_in_frame` reads it as a GCRF state expressed in the true-of-date axes at the frame epoch. Only Earth-centered messages convert: `state_in_frame` rejects a `CENTER_NAME` other than `EARTH`. Each delta-V is expressed in the frame named by its `MAN_REF_FRAME`: inertial vectors are rotated into GCRF by the frame bias between EME2000 and GCRF, and RTN vectors are rotated by the RTN-to-inertial matrix built from the spacecraft state at the ignition epoch. The message's maneuver ignition epochs precede its state vector epoch, so the example schedules the maneuvers relative to the state epoch, preserving the spacing between them:
```
duncaneddy: >  so `state_in_frame` reads it as a GCRF state expressed in the true-of-date axes at the frame epoch.

This seems like it's doing the incorrect thing. since we have TOD support, the state should be properly read as in the TOD frame and convert it properly to GCRF. not just incorectly interpreseting TOD-frame values as GCRF values.

This comment is overly specific about the frame transformation internals. It really just needs to detail the relevant methods and the fact that the methods/system does proper frame conversions to reconcile the vectors are in the right frames.

crates/brahe-py/src/ccsds.rs
duncaneddy: Replaced with that sentence.

docs/learn/ccsds/oem.md
duncaneddy: Reverted; the page keeps only the general statement that TOD segments load as TOD trajectories convertible with `to_frame`.

docs/learn/ccsds/opm.md
duncaneddy: The conversion is a real frame transformation, and the sentence described it badly. For this message (no `REF_FRAME_EPOCH`) the state is labeled TOD and `state_in_frame(GCRF)` asks the router to rotate TOD to GCRF at the state epoch. For a message that carries `REF_FRAME_EPOCH`, the declared frame is the true-of-date axes frozen at that epoch, so the router converts from those axes evaluated at that epoch and then routes from GCRF at the state epoch. The paragraph now says only that the message declares its state in TOD and `state_in_frame` converts it to GCRF through the reference frame router, and the CCSDS code no longer carries its own rotation helper: the frozen-epoch case calls `state_frame_to_frame` as well.

PR #532, opened 2026-09-06
Use IAU 2006/2000A by default with a selectable model

### LINKED ISSUE #522: Upgrade the CIO-based GCRF/ITRF chain from iauXys00b to iauXys06a
The CIO-based GCRF to ITRF transformation in src/frames/gcrf_itrf.rs computes the CIP coordinates and CIO locator with iauXys00b (IAU 2000B truncated nutation on the IAU 2000 precession basis), while the documentation labels the chain IAU 2006/2000A. Switching to iauXys06a would change every GCRF/ITRF result at the milliarcsecond level (about 1 mas, roughly 30 m at geostationary radius). The equinox-based MOD/TOD chain added in the frames/equinox-mod-tod stack deliberately uses the same 2000B basis (iauPn00b) so both chains stay consistent; upgrading one requires upgrading the other in the same change. Also correct the router table row for ITRF in docs/learn/frames/frame_transformations.md, which currently reads IAU 2006/2000A.

duncaneddy:
## Description

Makes IAU 2006/2000A the precession-nutation model for both Earth-orientation chains, the CIO-based GCRF/ITRF transformation (`iauXys06a`) and the equinox-based GCRF/MOD/TOD transformation (`iauPn06a`, `iauBp06`, `iauS06`), and adds a global `PrecessionNutationModel` switch so the IAU 2000B model (IAU 2000 precession with the truncated 2000B nutation series, the previous behavior) stays selectable in one call, mirroring how the global EOP provider is chosen. The equinox building blocks take the model explicitly as `bias_precession(epc, model)`, `nutation(epc, model)`, and `gast_rotation(epc, model)`. The SOFA cookbook reference tests now hold at 1e-12 for the bias-precession-nutation matrix and 2e-11 for matrices containing the Earth rotation angle under the default model, and at 1e-8 under IAU 2000B. Documentation labels the model actually used on every page that names it. A latent wrap bug exposed by the change, where `rem_euclid` returned exactly two pi for a right ascension a few 1e-17 below zero, is fixed so wrapped angles stay strictly below two pi.

### Added
- `PrecessionNutationModel` (`IAU2006A`, `IAU2000B`) with `set_precession_nutation_model` / `get_precession_nutation_model` and `bias_precession_nutation_model(epc, model)` in Rust and Python.

### Changed
- GCRF/ITRF, ECI/ECEF, and GCRF/MOD/TOD transformations evaluate IAU 2006/2000A by default (previously IAU 2000B); results move by 0.33 mas RMS and up to 1.24 mas over 1990-2040 (0.25 m at GEO, 0.04 m at LEO), and the CIP evaluation is about eight times slower unless `IAU2000B` is selected.
- Equinox building blocks `bias_precession_iau2000`, `nutation_iau2000b`, and `gast_rotation_iau2000b` are replaced by `bias_precession(epc, model)`, `nutation(epc, model)`, and `gast_rotation(epc, model)`.
- Documentation, rustdoc, and docstrings name the IAU 2006/2000A model and the switch; the router table rows for `ITRF`, `MOD`, and `TOD` are corrected.

### Fixed
- `wrap_to_2pi` and the right-ascension conversions return values strictly in [0, 2 pi); an input a few 1e-17 below zero no longer maps to exactly 2 pi.

## Related Issue
Closes #522

## Note to Reviewers
The 2e-11 floor on matrices containing the Earth rotation angle is a UT1 rounding difference against the printed cookbook values (1.5e-11 rad about the pole), not a model residual. The CHANGELOG entry that mentions IAU 2000B is historical and untouched. Model tradeoff measured with `pocs/poc_pn_models.rs` (git-ignored).


docs/learn/frames/equinox_frames.md
```diff
@@ -32,11 +32,35 @@ and the equinox-based form is
...
+    bh.set_precession_nutation_model(bh.PrecessionNutationModel.IAU2000B)
+    ```
+
+=== "Rust"
+
+    ``` rust
+    use brahe::frames::*;
+
+    set_precession_nutation_model(PrecessionNutationModel::IAU2000B);
+    ```
```
duncaneddy: Should be a code example that gets texted, also can expand this example with the different options.

docs/learn/frames/equinox_frames.md
```diff
@@ -32,11 +32,35 @@ and the equinox-based form is
...
+In both equations $W$ is polar motion, $C$ is the CIO-based bias-precession-nutation matrix used by [GCRF ↔ ITRF Transformations](gcrf_itrf.md), and $N$, $P$, $B$ are the classical nutation, precession, and frame bias matrices of the equinox chain. Brahe evaluates both chains on the same precession-nutation model and computes Greenwich apparent sidereal time ($\mathrm{GAST}$) as $\mathrm{ERA}$ minus the equation of the origins taken from the same combined nutation-precession-bias matrix used to reach TOD. As a result, converting a state from GCRF to ITRF through TOD agrees with the direct GCRF to ITRF transformation at the microarcsecond level.
+
+## Precession-Nutation Model
+
+Brahe evaluates the IAU 2006/2000A precession-nutation model by default. The truncated IAU 2000B model (IAU 2000 precession with the abridged IAU 2000B nutation series) is selectable with a single global setting that applies to every subsequent transformation, including the frame router and the batch forms:
```
duncaneddy: rewrite as:

"+
Brahe evaluates the IAU 2006/2000A precession-nutation model by default. The truncated IAU 2000B model (IAU 2000 precession with the abridged IAU 2000B nutation series) is selectable with a single global setting that applies to every subsequent transformation."

docs/learn/frames/equinox_frames.md
```diff
@@ -32,11 +32,35 @@ and the equinox-based form is
...
+The frame bias between GCRF and the classical J2000.0 mean equator and equinox is applied explicitly inside the bias-precession matrix that defines GCRF to MOD. The IERS dX/dY celestial pole offsets are residuals relative to this already-biased model, so no bias is double-counted; they enter the nutation step after conversion to dPsi/dEps corrections following SOFA cookbook Section 5.4. This differs from the IAU 1976/1980 convention, where the IERS dPsi/dEps corrections are measured against the unbiased model and therefore absorb the bias themselves, so no separate bias matrix is applied. TOD computed from the two conventions differs at the sub-milliarcsecond level, which is immaterial for interpreting data products labeled TOD.
```
duncaneddy: I would just delete this section. It's overly specific and I would want to directly verify fully before adding.

docs/learn/frames/equinox_frames.md
duncaneddy: The precession-nutation Model section is duplicated here and the gcrf_itrf.md documetation. It should probably only live in one common location that discusses selecting the tranformation model used across calls.

src/coordinates/radec.rs
```diff
@@ -851,6 +850,8 @@ mod tests {
     use approx::assert_abs_diff_eq;
     use serial_test::{parallel, serial};
 
+    use std::f64::consts::TAU;
```
duncaneddy: We should just use PI here since it's more common

src/math/angles.rs
```diff
@@ -101,35 +101,46 @@ pub fn oe_to_radians(oe: SVector6, angle_format: AngleFormat) -> SVector6 {
...
+/// - `f64`: The wrapped angle, in \[0, 2π). Units: (*rad*)
 ///
 /// # Examples
 ///
 /// ```
+/// use std::f64::consts::TAU;
 /// use brahe::math::angles::wrap_to_2pi;
 ///
-/// assert_eq!(wrap_to_2pi(2.0 * std::f64::consts::PI), 0.0);
+/// assert_eq!(wrap_to_2pi(TAU), 0.0);
```
duncaneddy: Example should use 2PI even if internally TAU is used to save a multiplication

docs/learn/frames/equinox_frames.md
duncaneddy: Replaced with a tested standalone example, `examples/frames/precession_nutation_model.{py,rs}`, that prints the default model, evaluates the GCRF to ITRF rotation under IAU 2006/2000A and IAU 2000B, prints their angular difference in milliarcseconds, shows the explicit-model form of `bias_precession_nutation_model`, and restores the default.

docs/learn/frames/equinox_frames.md
duncaneddy: Rewritten with that wording, linking to the new common page.

docs/learn/frames/equinox_frames.md
duncaneddy: Deleted the section.

docs/learn/frames/equinox_frames.md
duncaneddy: Both copies removed; the model selection now lives once at `docs/learn/frames/precession_nutation_model.md`, and the two chain pages link to it.

src/coordinates/radec.rs
duncaneddy: Changed to `2.0 * PI`.

src/math/angles.rs
duncaneddy: Example now uses `2.0 * PI`; the implementation keeps `TAU` internally.

PR #540, opened 2026-09-07
Add CITATION.cff

goelakash:
JOSS reviewers ask for a `CITATION.cff`, and this repository does not have one yet.

Generated from the author list in `origin:joss/paper.md`: 2 author(s), with given and family names taken from each author's ORCID record rather than split from the display name, because compound surnames do not split reliably.

`license` is `MIT`, read from the LICENSE file. `version` is `1.7.0`, the newest release tag. 

Verified with `cffconvert --validate`: valid against schema 1.2.0.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,14 @@
+cff-version: 1.2.0
+message: "If you use this software, please cite it as below."
+type: software
+title: "Brahe: A Modern Astrodynamics Library for Research and Engineering Applications"
+authors:
+  - family-names: "Eddy"
+    given-names: "Duncan"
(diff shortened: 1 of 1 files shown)


duncaneddy:
Appreciate the contribution!

I didn't see mention of `CITATION.cff` as a standard anywhere in https://joss.readthedocs.io/en/latest/, but I do see it's a common standard, so this is a good addition to the repo.

I'd like to add an automated process to automatically update the version as part of the release process so it doesn't need to be maintained and `Cargo.toml` remains the single source of truth throughout the repo for package versioning, but happy to accept this as-is for now and address that as follow-up work.

PR #536, opened 2026-09-07
Separate frame axes from centers with centered celestial frames

duncaneddy:
## Description

Separates a celestial frame's orientation from its origin. A new `FrameAxes` enum names every orientation the reference frame router evaluates (ICRF, EME2000, MOD, TOD, ITRF, the lunar and Mars body-fixed axes, the synodic axes, and the IAU, PCK, and custom body-fixed families), and `CelestialFrame::Centered { center, axes }` pairs any of them with any origin expressed as a `FrameCenter` (a NAIF body or a two-body barycenter). `CelestialFrame::centered(center, axes)` is the canonical constructor and returns the existing named frame whenever one exists, so `GCRF`, `LCI`, `EME2000`, `MCMF`, and the rest keep their meaning and equality; `axes()` and `center()` (a `FrameCenter`, built from a `NAIFId` or a raw ID) read the two halves of any frame. The router now rotates by axes and translates by center, so a Mars-centered EME2000 or ITRF frame converts to `MCI` without an ephemeris and to `GCRF` through the loaded kernels, and Keplerian elements are accepted in any ICRF- or EME2000-axis frame about a massive body. CCSDS messages map `REF_FRAME` to axes and `CENTER_NAME` to the origin through `NAIFId::from_name`, so Mars- and Moon-centered OEM and OPM messages load in their true frame instead of being read as Earth-centered or rejected. Last PR of the equinox-frames stack, rebased onto main after #528 merged.

### Added
- `FrameAxes` and `FrameCenter` types, `CelestialFrame::Centered`, `CelestialFrame::centered(center, axes)`, `CelestialFrame::axes()`, and `CelestialFrame::center()` in Rust and Python (`FrameAxes`, `FrameCenter`, `CelestialFrame.Centered(center, axes)`, `.axes`, `.center`, `.center_naif_id`).
- `NAIFId::name()`, `NAIFId::from_name()` (accepting NAIF aliases such as `SSB` and `EMB`), and `Display` for `NAIFId`; Python `NAIFId.from_name()` and `NAIFId.naif_name`.
- Python `ReferenceFrame.celestial_frame` returning the wrapped `CelestialFrame` (or `None`), so a trajectory's frame exposes `axes` and `center`.
- `brahe::ccsds::interop::segment_celestial_frame`, `state_for_segment`, and `state_into_segment_axes` for writing trajectories into OEM segments, shared by the Python bindings.
[...]

PR #542, opened 2026-09-07
Add TEME-of-date frame

### LINKED ISSUE #524: Add TEME-of-epoch frame
The planned TEME frame (true equator, mean equinox of date, anchored to the SGP4 convention through GMST 1982) covers CCSDS TEMEOFDATE. The SANA registry also defines TEMEOFEPOCH, a TEME frame frozen at a reference epoch. Add a CelestialFrame variant carrying the reference epoch, define its rotation as the TEME-of-date rotation evaluated at that epoch, and map the CCSDS TEMEOFEPOCH token to it.

duncaneddy:
## Description

Adds the true equator, mean equinox of date (TEME) frame, the output frame of SGP4, as a first-class `CelestialFrame::TEME`. A new `src/frames/teme.rs` defines the frame through Greenwich mean sidereal time on the IAU 1982 model, `[ITRF] = W R3(GMST82) [TEME]`, which combined with the CIO chain gives `[TEME] = R3(ERA - GMST82) C [GCRF]` and anchors the frame to the SGP4 convention of Vallado et al. (AIAA 2006-6753, Appendix C). GMST 1982 comes from SOFA `iauGmst82` on UT1. Pairwise GCRF/TEME and TEME/ITRF rotation, position, state, and batch transforms follow the `equinox.rs` conventions: TEME/GCRF states are rotation-only, TEME/ITRF states carry the Earth-rotation transport term. The SGP propagator now routes its output through the shared functions instead of a private GMST 1982 and hand-written TEME to PEF to ITRF steps; `state` keeps returning raw TEME, and the configured output frame (GCRF by default) still governs `propagate_to`, the stored trajectory, and the frame accessors. TEME is a `FrameAxes::TEME` orientation, with `CelestialFrame::TEME` as the canonical Earth-centered value, so it composes with the centered-frame model from #536; following #539, Keplerian elements are accepted in TEME axes as they are in MOD and TOD. The CCSDS layer maps the ODM `TEME` and ADM `TEMEOFDATE` tokens to the native frame in both directions, so OEM, OPM, and AEM messages declared in TEME load and write without conversion; a `REF_FRAME_EPOCH` on a TEME message is ignored until the of-epoch frames land in the next PR. This is the first PR of a two-PR stack; the second adds the epoch-frozen `TODofEpoch` and `TEMEofEpoch` frames and closes #524.

### Added
- `FrameAxes::TEME` and `CelestialFrame::TEME` in the reference frame router with `from_string`/`Display` support, the `FrameAxes.TEME` and `CelestialFrame.TEME` Python attributes, and the `TEME` choice in the `brahe transform` CLI commands.
- Keplerian element trajectories and propagator output in TEME axes for `OrbitTrajectory`, `KeplerianPropagator`, and `SGPPropagator`, alongside MOD and TOD.
- `gmst82`, `greenwich_mean_sidereal_rotation`, and the `rotation_`/`position_`/`state_` functions with batch forms for GCRF/TEME and TEME/ITRF in both directions (Rust and Python).
- CCSDS mappings for the ODM `TEME` and ADM `TEMEOFDATE` reference frame tokens to `CelestialFrame::TEME` in both directions; OEM, OPM, and AEM messages in TEME convert to native trajectories and states.
- Learn page and API reference for the TEME frame with three standalone examples in Python and Rust, and a TEME output section in the SGP propagation guide.

### Changed
- `SGPPropagator` converts its TEME output through the shared frame functions: GCRF and ITRF outputs use the pairwise transforms, `state_eci` shares the GCRF path, and every other frame goes through the router from TEME. GCRF, ITRF, and PEF values move by up to about 1e-2 m and 1e-5 m/s in low Earth orbit, which is the GMST 1982 evaluation gaining precision (two-part MJD instead of a single-double Julian date, about 1e-9 rad about the pole); the reference-value tests pin the recorded values at that level.
- `SGPPropagator.state` documents that it returns the raw SGP4 state in TEME; `state_gcrf`, `state_itrf`, and `state_in_frame` provide other frames.

## Related Issue
Related to #524 (closed by the second PR of this stack).

## Note to Reviewers
The definition is pinned by `rotation_teme_to_itrf * rotation_gcrf_to_teme == rotation_gcrf_to_itrf` at 1e-15 over several epochs, a finite-difference check of the TEME/ITRF transport term, and the recorded SGP propagator GCRF/ITRF/PEF reference values at three epochs. TEME to ITRF through the router composes through GCRF like TOD does; the cheap `W R3(GMST82)` path is the pairwise `state_teme_to_itrf`, which the propagator calls directly for ITRF output. A pre-existing coverage tooling defect, unrelated to this branch and reproducible on `main`, makes `just test-python` fail at its merge step when a non-instrumented `brahe` rlib is cached in `target/debug`, because `cargo llvm-cov clean --workspace` does not clear it; running the recipe from a clean `target` succeeds.


src/frames/teme.rs
```diff
@@ -0,0 +1,1438 @@
...
+/// ```
+///
+/// # References
+/// - SOFA `gmst82`; Vallado et al., "Revisiting Spacetrack Report #3",
+///   AIAA 2006-6753, Appendix C
+pub fn gmst82(epc: Epoch) -> f64 {
+    let ut1 = epc.mjd_as_time_system(TimeSystem::UT1);
+    let gmst = unsafe { rsofa::iauGmst82(MJD_ZERO, ut1) };
+    wrap_to_2pi(gmst)
+}
```
duncaneddy: This should be a function/property of Epoch (which already has a similar `gmst06` method.

src/frames/teme.rs
```diff
@@ -0,0 +1,1438 @@
...
+/// ```
+///
+/// # References
+/// - SOFA `rz`; Vallado, "Fundamentals of Astrodynamics and Applications",
+///   4th ed., Section 3.7
+pub fn greenwich_mean_sidereal_rotation(epc: Epoch) -> SMatrix3 {
+    let mut r = [[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]];
+    unsafe {
+        rsofa::iauRz(gmst82(epc), &mut r[0]);
+    }
```
duncaneddy: Instead of using SOFA Rz, why don't use rotation constructors from attitude.rs?

src/frames/teme.rs
```diff
@@ -0,0 +1,1438 @@
...
+/// # References
+/// - Vallado et al., "Revisiting Spacetrack Report #3", AIAA 2006-6753,
+///   Appendix C; Vallado, "Fundamentals of Astrodynamics and Applications",
+///   4th ed., Section 3.7
+pub fn rotation_gcrf_to_teme(epc: Epoch) -> SMatrix3 {
+    let ut1 = epc.mjd_as_time_system(TimeSystem::UT1);
+    let mut r = [[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]];
+    unsafe {
+        let era = rsofa::iauEra00(MJD_ZERO, ut1);
+        rsofa::iauRz(era - gmst82(epc), &mut r[0]);
```
duncaneddy: use of module internal Rz implementation is preferred.

src/frames/teme.rs
duncaneddy: Addressed in db591b046: GMST 1982 now lives on `Epoch::gmst82(angle_format)` beside `gmst`/`gast`, with `Epoch.gmst82` in Python; the free `frames::gmst82` is removed.

src/frames/teme.rs
duncaneddy: Addressed in db591b046: `greenwich_mean_sidereal_rotation` builds `R3(GMST82)` with `attitude::Rz`.

src/frames/teme.rs
duncaneddy: Addressed in db591b046: `rotation_gcrf_to_teme` builds `R3(ERA - GMST82)` with `attitude::Rz`; only the ERA angle still comes from `iauEra00`, as `earth_rotation` does.

tests/ccsds/test_oem.py
```diff
@@ -1431,60 +1485,60 @@ def _teme_trajectory(epoch):
...
     epoch = Epoch.from_datetime(2024, 1, 1, 12, 0, 0.0, 0.0, brahe.UTC)
-    traj = _teme_trajectory(epoch)
+    traj = _tdr_trajectory(epoch)
 
     seg = OEMSegment(
         object_name="SAT",
         object_id="2024-100A",
         center_name="EARTH",
-        ref_frame="TEME",
+        ref_frame="TDR",
```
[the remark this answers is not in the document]
duncaneddy: Approved: retargeting the unmapped-frame tests from `TEME` to `TDR` was specified in the reviewed implementation plan once `TEME` gained a native frame, and the TEME coverage was added alongside (`test_oem_in_teme_loads_as_teme_trajectory` and its Python mirror).

PR #549, opened 2026-09-11
Add Modified ITC ephemeris format

duncaneddy:
# Pull Request

## Description

Adds `brahe::itc`, a reader and writer for the Space-Track Modified ITC ephemeris format (the format Starlink publishes its public ephemerides in), with 6x6 position and velocity covariance, and `SpaceTrackEphemerisFileName` for parsing and generating the Space-Track file-name convention. Files load in SI units in the frame implied by the file name's DataType, default EME2000, and a parsed Starlink file writes back as the original text apart from the terminating newline. Python exposes the same API under `brahe.itc` and `brahe.spacetrack`. Second PR of the Starlink ephemeris client stack; trajectory conversion follows in the next PR.

### Added
- `brahe::math::is_symmetric` for square-matrix symmetry checks within a relative tolerance
- `brahe::itc` (`brahe.itc`): `ITC`, `ITCHeader`, `ITCStateVector`, `ITCCovarianceFrame` with `from_str`, `from_file`, `to_string`, `to_file`, record builders, and `file_name` for compliant Space-Track names
- `brahe::spacetrack::SpaceTrackEphemerisFileName` and `SpaceTrackEphemerisFileCategory` (`brahe.spacetrack`) to parse and generate the Spaceflight Safety Handbook file-name convention
- Starlink Modified ITC test assets under `test_assets/starlink/`

## Note to Reviewers
Covariance is carried on the `ITC` message only in this PR; attaching it to `OrbitTrajectory` (with RTN to EME2000 rotation) is the next PR in the stack. The full 2 MB asset is intentional: it is the parser's acceptance test and later seeds the offline Starlink client example.


src/itc/parse.rs
```diff
@@ -0,0 +1,632 @@
...
+        header.ephemeris_stop = Some(parse_header_epoch(value, "ephemeris_stop")?);
+    }
+    if let Some(value) = header_field(lines[1], keys[2], &[keys[0], keys[1]])
+        && !value.is_empty()
+    {
+        header.step_size = Some(
+            value
+                .parse::<f64>()
+                .map_err(|_| parse_error(format!("invalid step_size value '{}'", value)))?,
+        );
```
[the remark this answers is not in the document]
duncaneddy: Addressed in 0a9aa5566: `parse_header` now rejects a non-finite or negative `step_size` with the existing `invalid step_size value` error; covered by `test_parse_errors` in Rust and Python (`NaN` and `inf` cases).

src/itc/write.rs
```diff
@@ -0,0 +1,525 @@
...
+                state.position[2] * M_TO_KM,
+                state.velocity[0] * M_TO_KM,
+                state.velocity[1] * M_TO_KM,
+                state.velocity[2] * M_TO_KM,
+            ));
+            if let Some(covariance) = self.covariances.get(index) {
+                let mut values = Vec::with_capacity(21);
+                for row in 0..6 {
+                    for col in 0..=row {
+                        values.push(format_scientific(covariance[(row, col)] * M2_TO_KM2));
```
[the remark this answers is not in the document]
duncaneddy: Addressed in 0a9aa5566: `push_state_with_covariance` now rejects matrices that are not symmetric (relative tolerance 1e-9 to allow float noise), and `to_string` re-checks symmetry on the public `covariances` field before writing, so the lower-triangle serialization can no longer silently alter an accepted matrix. Tests added in `types.rs`, `write.rs` and `tests/itc/test_itc.py`.

src/clients/spacetrack/file_name.rs
```diff
@@ -0,0 +1,641 @@
...
+/// ```
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum EphemerisFileCategory {
+    /// The trajectory the satellite is planned to fly, including routine maneuvers.
+    Operational,
+    /// A planning-only trajectory that assumes a special maneuver.
+    Special,
+}
+
+impl EphemerisFileCategory {
```
duncaneddy: I think these types should likely be `SpaceTrackEphemerisFileCategory` and `SpaceTrackEphemerisFileName` since it is spacetrack specific.

src/itc/frames.rs
```diff
@@ -0,0 +1,260 @@
...
+/// # Examples
+///
+/// ```
+/// use brahe::itc::state_frame_for_data_type;
+/// use brahe::frames::CelestialFrame;
+///
+/// assert_eq!(state_frame_for_data_type("MEME").unwrap(), CelestialFrame::EME2000);
+/// assert!(state_frame_for_data_type("GCRF").is_err());
+/// ```
+pub fn state_frame_for_data_type(data_type: &str) -> Result<CelestialFrame, BraheError> {
```
duncaneddy: This seems like it doesn't need to be a public function?

src/itc/frames.rs
```diff
@@ -0,0 +1,260 @@
...
+/// # Examples
+///
+/// ```
+/// use brahe::itc::data_type_for_state_frame;
+/// use brahe::frames::CelestialFrame;
+///
+/// assert_eq!(data_type_for_state_frame(&CelestialFrame::EME2000).unwrap(), "MEME");
+/// assert!(data_type_for_state_frame(&CelestialFrame::GCRF).is_err());
+/// ```
+pub fn data_type_for_state_frame(frame: &CelestialFrame) -> Result<&'static str, BraheError> {
```
duncaneddy: This seems like it doesn't need to be a public function?

src/itc/types.rs
```diff
@@ -0,0 +1,710 @@
...
+
+use nalgebra::SMatrix;
+
+use crate::clients::spacetrack::EphemerisFileName;
+use crate::frames::CelestialFrame;
+use crate::time::Epoch;
+use crate::utils::BraheError;
+
+/// Whether a 6x6 covariance matrix is symmetric to within floating-point noise.
+pub(super) fn is_symmetric(m: &SMatrix<f64, 6, 6>) -> bool {
```
duncaneddy: This seems like a function that would be useful to have generically in the crate in the utils module maybe?

src/itc/write.rs
```diff
@@ -0,0 +1,551 @@
...
+    format!(
+        "{:04}-{:02}-{:02} {:02}:{:02}:{:02} UTC",
+        year,
+        month,
+        day,
+        hour,
+        minute,
+        second.floor() as u32
+    )
+}
```
duncaneddy, line 53: Doesn't the default print using _with_decimals already accomplish this type of formatting?

duncaneddy: It does not, unfortunately: `isostring_with_decimals` and `to_string_as_time_system` produce the ISO calendar form (`YYYY-MM-DDTHH:MM:SS.sss`), while the ITC record token is `YYYYDDDHHMMSS.sss` — day of year, no separators — and it has to round to the millisecond with carry into the day so a fraction that rounds up never prints `60.000` seconds. If a day-of-year string form is useful elsewhere I can move this to `time` as a public `Epoch` helper instead of keeping it private to the writer; say the word and I will do that as part of the rework.

src/clients/spacetrack/file_name.rs
duncaneddy: Renamed in d6fa2cd12 to `SpaceTrackEphemerisFileName` and `SpaceTrackEphemerisFileCategory` across Rust, the Python bindings and package, tests and docs; no aliases kept. The higher branches pick the rename up as they are rebased.

src/itc/frames.rs
duncaneddy: Made `pub(crate)` in 68866dc3a (and removed from the Python bindings and `brahe.itc`); the doctests became unit tests in `frames.rs`.

src/itc/frames.rs
duncaneddy: Same as above: `pub(crate)` in 68866dc3a, no Python surface.

src/itc/types.rs
duncaneddy: Moved in 68866dc3a to `brahe::math::is_symmetric(m, rtol)`, generic over any square nalgebra matrix (static or dynamic), with its own unit test; the ITC push methods call it with the same 1e-9 tolerance as before. No Python binding, matching the module's other validation helpers (only the two covariance constructors are exposed).

PR #551, opened 2026-09-11
Convert Modified ITC messages to and from trajectories

duncaneddy:
# Pull Request

## Description

Converts Modified ITC messages to `OrbitTrajectory` in the declared state frame with covariance attached, and builds messages back from trajectories. RTN covariance is rotated with each record's geocentric state using the block-diagonal convention of the NASA Conjunction Assessment Handbook (Appendix N eq. N-13) and CARA's `RIC2ECI` by default; the rotating-frame form is selectable through `OrbitRelativeFrameVariant`. EME2000 and ITRF covariance frames are accepted in both directions through the frame-router Jacobian from #557, and a trajectory in any Cartesian frame can be written out. An ITC-derived trajectory passes directly to `location_accesses`. Third PR of the Starlink ephemeris client stack.

### Added
- `ITC::to_trajectory` and `ITC::to_trajectory_with_covariance_variant` (`ITC.to_trajectory(covariance_variant=None)` in Python) producing a six-dimensional Cartesian `OrbitTrajectory` with covariance rotated from RTN into the state frame
- `ITC::from_trajectory` and `ITC::from_trajectory_with_covariance_variant` building a message from a trajectory's samples, converting frames and carrying covariance through the GCRF/EME2000 bias and into RTN
- `TryFrom<&ITC> for DOrbitTrajectory`
- `ITCStateVector::to_vector` and `From<&ITCStateVector> for SVector6` (`ITCStateVector.to_vector()` in Python)
- `brahe::utils::dmatrix_from_smatrix` and `smatrix_from_dmatrix`, shared fixed-to-dynamic matrix conversions

## Note to Reviewers
`from_trajectory` requires a Cartesian trajectory and rejects Keplerian representations (their covariance is in element space); convert with `to_frame` or `to_eci` first. The block-diagonal default and the sources for it are documented in References sections on the conversion methods in both languages. Depends on #557 for covariance routing.


src/itc/interop.rs
```diff
@@ -0,0 +1,862 @@
...
+use crate::trajectories::dorbit_trajectory::DOrbitTrajectory;
+use crate::trajectories::traits::{
+    OrbitRepresentation, covariance_frame_allowed, inertial_covariance_rotation,
+};
+use crate::utils::BraheError;
+
+use super::types::{ITC, ITCCovarianceFrame, ITCHeader, ITCStateVector};
+
+/// Places `r` on both diagonal blocks of a 6x6 matrix.
+fn block_diagonal(r: SMatrix3) -> SMatrix6 {
```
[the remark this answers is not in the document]
duncaneddy: Addressed in fad264151: every helper in `src/itc/interop.rs` (`block_diagonal`, `symmetrize`, `skew`, both Jacobian helpers, `state_vector`, `dmatrix_from`) now carries `# Arguments`, `# Returns` and an `# Examples` block.

crates/brahe-py/src/itc.rs
```diff
@@ -583,6 +587,85 @@ impl PyITC {
...
+    /// Args:
+    ///     covariance_variant (OrbitRelativeFrameVariant, optional): ``INERTIAL`` (default, block-diagonal ``[[R, 0], [0, R]]``) or ``ROTATING`` (adds the frame-rate coupling block).
+    ///
+    /// Returns:
+    ///     OrbitTrajectory: Six-dimensional Cartesian trajectory, named after the source file's object when known.
+    ///
+    /// Raises:
+    ///     BraheError: If the message is empty or covariance is present with an unsupported frame combination.
+    #[pyo3(signature = (covariance_variant=None))]
+    fn to_trajectory(&self, covariance_variant: Option<PyOrbitRelativeFrameVariant>) -> PyResult<PyOrbitalTrajectory> {
```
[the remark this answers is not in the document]
duncaneddy: Addressed in fad264151: `to_trajectory`, `to_trajectory_with_covariance_variant`, `from_trajectory` and `from_trajectory_with_covariance_variant` now each have a runnable `Example` section; stubs regenerate cleanly.

src/itc/interop.rs
```diff
@@ -0,0 +1,862 @@
...
+                b.covariances[0][(3, 1)],
+                itc.covariances[0][(3, 1)],
+                epsilon = 1e-9 * itc.covariances[0][(3, 1)].abs()
+            );
+        }
+    }
+
+    #[test]
+    #[parallel]
+    fn test_from_trajectory_eme2000_covariance_frame_keeps_inertial_covariance() {
```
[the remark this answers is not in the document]
duncaneddy: Addressed in fad264151: added `test_eme2000_covariance_frame_keeps_inertial_covariance` to `tests/itc/test_itc.py`, mirroring the Rust test (EME2000 covariance frame passes through unrotated in both directions).

src/trajectories/traits.rs
```diff
@@ -155,6 +159,38 @@ pub(crate) fn covariance_frame_allowed(frame: &ReferenceFrame) -> bool {
...
+        )));
+    }
+    Ok(
+        match (is_eme2000_axes_frame(from), is_eme2000_axes_frame(to)) {
+            (false, true) => rotation_gcrf_to_eme2000(),
+            (true, false) => rotation_eme2000_to_gcrf(),
+            _ => SMatrix3::identity(),
+        },
+    )
+}
```
duncaneddy: Unless OrbitTrajectory already normalized all stored trajectories to GCRF it doesn't make sense to store Covariance only as GCRF, especially given now that we have the frame router capabilities now. It would seem to make most sense to just store the internal data as whatever is is then route it. So adding this inertial storage seems like overkill

src/itc/interop.rs
```diff
@@ -0,0 +1,956 @@
...
+/// ```text
+/// block_diagonal(3 * I3) is the 6x6 matrix with 3 on the diagonal in
+/// rows/columns 0-2 and 3-5, and zero everywhere else.
+/// ```
+fn block_diagonal(r: SMatrix3) -> SMatrix6 {
+    let mut m = SMatrix6::zeros();
+    m.fixed_view_mut::<3, 3>(0, 0).copy_from(&r);
+    m.fixed_view_mut::<3, 3>(3, 3).copy_from(&r);
+    m
+}
```
duncaneddy: Seems like a good function to instead implement as a common utility.

src/itc/interop.rs
```diff
@@ -0,0 +1,956 @@
...
+/// * `SMatrix6`: `(m + mᵀ) / 2`
+///
+/// # Examples
+/// ```text
+/// symmetrize applied to a matrix with a 2.0 in position (0, 1) and zero
+/// in position (1, 0) produces 1.0 in both positions.
+/// ```
+fn symmetrize(m: SMatrix6) -> SMatrix6 {
+    (m + m.transpose()) * 0.5
+}
```
duncaneddy: Also seems like a good common utility function.

src/itc/interop.rs
```diff
@@ -0,0 +1,956 @@
...
+/// # Returns
+/// * `SMatrix3`: Skew-symmetric matrix of `w`
+///
+/// # Examples
+/// ```text
+/// skew([0, 0, 2]) produces [[0, -2, 0], [2, 0, 0], [0, 0, 0]]
+/// ```
+fn skew(w: nalgebra::Vector3<f64>) -> SMatrix3 {
+    SMatrix3::new(0.0, -w[2], w[1], w[2], 0.0, -w[0], -w[1], w[0], 0.0)
+}
```
duncaneddy: Not quite sure why we need this if you can implement cross. but also seems like a good common utility.

src/itc/interop.rs
```diff
@@ -0,0 +1,956 @@
...
+
+/// Jacobian taking a covariance from the RTN frame of `x` into the inertial
+/// frame `x` is expressed in.
+///
+/// `Inertial` gives the block-diagonal `[[R, 0], [0, R]]` form used by the
+/// NASA CA Handbook (Appendix N eq. N-13) and CARA's `RIC2ECI`. `Rotating`
+/// gives `[[R, 0], [R·[ω×], R]]`, the transform of a truly rotating frame
+/// with ω the RTN frame rate in RTN components.
```
duncaneddy: Should cite these two sources with ## References section with proper citation

src/itc/interop.rs
```diff
@@ -0,0 +1,956 @@
...
+/// ```
+fn rtn_to_frame_jacobian(x: SVector6, variant: OrbitRelativeFrameVariant) -> SMatrix6 {
+    let r = rotation_rtn_to_eci(x);
+    let mut j = block_diagonal(r);
+    if variant == OrbitRelativeFrameVariant::Rotating {
+        let coupling = r * skew(omega_rtn(x));
+        j.fixed_view_mut::<3, 3>(3, 0).copy_from(&coupling);
+    }
+    j
+}
```
duncaneddy: This is in the itc interop module, but seems like it should be in the rtn_eci.rs frame definition file (it's more general and would apply to any such RTN <> ECI transformation.

src/itc/interop.rs
```diff
@@ -0,0 +1,956 @@
...
+/// ```
+fn frame_to_rtn_jacobian(x: SVector6, variant: OrbitRelativeFrameVariant) -> SMatrix6 {
+    let rt = rotation_rtn_to_eci(x).transpose();
+    let mut j = block_diagonal(rt);
+    if variant == OrbitRelativeFrameVariant::Rotating {
+        let coupling = -skew(omega_rtn(x)) * rt;
+        j.fixed_view_mut::<3, 3>(3, 0).copy_from(&coupling);
+    }
+    j
+}
```
duncaneddy: Like the above this is in the itc interop module, but seems like it should be in the rtn_eci.rs frame definition file (it's more general and would apply to any such RTN <> ECI transformation.

src/itc/interop.rs
```diff
@@ -0,0 +1,956 @@
...
+fn state_vector(sv: &ITCStateVector) -> SVector6 {
+    SVector6::new(
+        sv.position[0],
+        sv.position[1],
+        sv.position[2],
+        sv.velocity[0],
+        sv.velocity[1],
+        sv.velocity[2],
+    )
+}
```
duncaneddy: Isn't there an easier way to do this rather than defining a helper function. Shouldn't ITCStateVector have a `to_vector()` method or can't we just implement an `Into` trait conversion type so it can be coerced

src/itc/interop.rs
```diff
@@ -0,0 +1,956 @@
...
+/// * `DMatrix<f64>`: Dynamically-sized 6x6 matrix with the same entries
+///
+/// # Examples
+/// ```text
+/// dmatrix_from(&SMatrix6::identity()) produces a 6x6 DMatrix equal to
+/// the identity matrix.
+/// ```
+fn dmatrix_from(m: &SMatrix6) -> DMatrix<f64> {
+    DMatrix::from_iterator(6, 6, m.iter().cloned())
+}
```
duncaneddy: This should be in a common place since it applies to any DMatrix/SMatrix (utils.rs) if needed at all.

src/itc/interop.rs
```diff
@@ -0,0 +1,956 @@
...
+    /// let itc = ITC::from_file(
+    ///     "test_assets/starlink/MEME_100002_STARLINK-37711_2540149_Operational_1473385800_UNCLASSIFIED.txt",
+    /// ).unwrap();
+    /// let traj = itc.to_trajectory().unwrap();
+    /// assert_eq!(traj.states.len(), 50);
+    /// assert_eq!(traj.name.as_deref(), Some("STARLINK-37711"));
+    /// ```
+    pub fn to_trajectory(&self) -> Result<DOrbitTrajectory, BraheError> {
+        self.to_trajectory_with_covariance_variant(OrbitRelativeFrameVariant::Inertial)
+    }
```
duncaneddy, line 115: Is the Covariance always guaranteed to be in the Inertial RTN frame variant? IT seems like it might not be so then you have to know how to convert. Is that true?

src/itc/interop.rs
```diff
@@ -0,0 +1,956 @@
...
+
+    /// Converts the message to a trajectory, choosing how RTN covariance is
+    /// rotated into the state frame.
+    ///
+    /// Covariance requires a state frame that may carry it (GCRF or EME2000);
+    /// other frames are an error. RTN covariance is rotated with the record's own state:
+    /// `Inertial` uses `[[R, 0], [0, R]]` (NASA CA Handbook Appendix N eq.
+    /// N-13; CARA `RIC2ECI`), `Rotating` uses `[[R, 0], [R·[ω×], R]]` with ω
+    /// the RTN frame rate. A covariance frame of `EME2000` is accepted only
+    /// with an `EME2000` state frame; `ITRF` covariance is not supported.
```
duncaneddy: Should also add "References" section for cited transformaitons

src/itc/interop.rs
```diff
@@ -0,0 +1,956 @@
...
+                            frame
+                        )));
+                    }
+                    self.covariances.iter().map(dmatrix_from).collect()
+                }
+                ITCCovarianceFrame::ITRF => {
+                    return Err(BraheError::Error(format!(
+                        "Modified ITC covariance frame ITRF cannot be attached to a trajectory in {}; trajectory covariance is supported only in EME2000 or GCRF axes, so ITRF covariance must be rotated before conversion",
+                        frame
+                    )));
```
duncaneddy: It seems like we would be able to address this and rotate an ITRF covariance into a consistent frame if we have the underlying base frame state frame?

src/itc/interop.rs
```diff
@@ -0,0 +1,956 @@
...
+        let state_frame = header.state_frame;
+
+        let axes_rotation = match &trajectory.covariances {
+            Some(_) => {
+                if covariance_frame == ITCCovarianceFrame::ITRF {
+                    return Err(BraheError::Error(format!(
+                        "Modified ITC covariance frame ITRF cannot be produced from a trajectory in {}; trajectory covariance is supported only in EME2000 or GCRF axes",
+                        target
+                    )));
+                }
```
duncaneddy: Again, instead of erroring, seems like something we could handle and produce.

src/trajectories/traits.rs
```diff
@@ -155,6 +159,38 @@ pub(crate) fn covariance_frame_allowed(frame: &ReferenceFrame) -> bool {
...
+/// the epoch.
+///
+/// # Arguments
+/// * `from` - Frame the covariance is expressed in
+/// * `to` - Frame to rotate it into
+///
+/// # Returns
+/// * `Ok(SMatrix3)`: Rotation taking vectors from `from` axes to `to` axes (identity when the axes match)
+/// * `Err(BraheError)`: If either frame may not carry covariance
+pub(crate) fn inertial_covariance_rotation(
```
duncaneddy: I do see how this is used later in trajectory covariance rotation. This also seems like common (crate) functionality that might find a better place in the utils module, since I could use later reuse in propagators and estimators.

src/itc/interop.rs
duncaneddy: Not guaranteed by the format. The SFS Handbook defines only the UVW axes (R along position, W along angular momentum, U completing the triad) and says nothing about whether the velocity block is expressed in the rotating or the non-rotating RTN frame. The block-diagonal (inertial-axes) reading is what Space-Track's consumers and NASA CARA use (CA Handbook Appendix N eq. N-13; `RIC2ECI.m` in the CARA analysis tools), so it is the default, and `OrbitRelativeFrameVariant::Rotating` is available for producers who mean the rotating form. I will add the References sections requested on the other threads with the full citations so this is documented where the conversion lives.

src/trajectories/traits.rs
duncaneddy: Agreed. The GCRF/EME2000-only rule predates this stack (#527) and this PR only bolted the bias rotation onto it. Plan, per our discussion: a new `frames/covariance.rs` PR inserted directly above #549 provides `state_transform_jacobian(from, to, epoch)` (6x6, built from the router's affine state transform, so it covers every router frame including the ω-coupled rotating ones), `rotate_covariance`, and `covariance_frame_to_frame`; both trajectory types then accept covariance in any Cartesian frame and `to_frame` rotates it for every pair. This PR is rebased on top, drops `inertial_covariance_rotation` and its local helpers, and gains ITRF/EME2000 covariance support in both directions. #552 will be closed in favour of that PR.

src/itc/interop.rs
duncaneddy: Now `brahe::math::block_diagonal` (#557); this PR drops its copy when rebased.

src/itc/interop.rs
duncaneddy: Now `brahe::math::symmetrize` / `symmetrize_6` (#557); this PR drops its copy when rebased.

src/itc/interop.rs
duncaneddy: Now `brahe::math::skew_symmetric` (#557). It stays a matrix because the Jacobians need `[ω]×` as a block rather than a product; the copy here goes away on rebase.

src/itc/interop.rs
duncaneddy: Moved to `relative_motion::eci_rtn` as `jacobian_rtn_to_eci` / `covariance_rtn_to_eci` with the variant argument (#557); this PR will call those.

src/itc/interop.rs
duncaneddy: Moved to `relative_motion::eci_rtn` as `jacobian_eci_to_rtn` / `covariance_eci_to_rtn` (#557).

src/itc/interop.rs
duncaneddy: Now `brahe::math::symmetrize` (#557), with the DMatrix rotation in `frames::covariance::rotate_covariance`.

src/trajectories/traits.rs
duncaneddy: Replaced by `frames::covariance::state_transform_jacobian` in #557, which handles any router frame pair (not only the GCRF/EME2000 bias) and is what propagators and estimators can reuse.

src/trajectories/traits.rs
duncaneddy: Done across #557 and this PR (rebased onto it, bd6f6fd73..7459ed761): covariance now lives in the trajectory's own frame for any Cartesian frame and is routed with the 6x6 state-transform Jacobian; the interop no longer carries any inertial-only storage rule.

src/itc/interop.rs
duncaneddy: Removed here; the shared `brahe::math::block_diagonal` from #557 is used instead (bd6f6fd73).

src/itc/interop.rs
duncaneddy: Removed here; `brahe::math::symmetrize`/`symmetrize_6` from #557 are used instead (bd6f6fd73).

src/itc/interop.rs
duncaneddy: Removed here; `brahe::math::skew_symmetric` from #557 is used by the RTN Jacobians (bd6f6fd73).

src/itc/interop.rs
duncaneddy: Added `# References` sections on all four conversion methods (Rust) and `References:` in the Python docstrings, citing the SFS Handbook v1.7, NASA CA Handbook Appendix N, CARA `RIC2ECI.m`, and Vallado AAS 03-526 with URLs (bd6f6fd73, corrected in b390939c1).

src/itc/interop.rs
duncaneddy: Moved to `relative_motion::eci_rtn::jacobian_rtn_to_eci` / `covariance_rtn_to_eci` in #557; this PR calls those (bd6f6fd73).

src/itc/interop.rs
duncaneddy: Moved to `relative_motion::eci_rtn::jacobian_eci_to_rtn` / `covariance_eci_to_rtn` in #557; this PR calls those (bd6f6fd73).

src/itc/interop.rs
duncaneddy: Replaced by `ITCStateVector::to_vector()` and `impl From<&ITCStateVector> for SVector6` (Python `ITCStateVector.to_vector()`) in bd6f6fd73.

src/itc/interop.rs
duncaneddy: Removed here; `frames::covariance::rotate_covariance` and `math::symmetrize` from #557 cover it (bd6f6fd73).

src/itc/interop.rs
duncaneddy: References sections added on the conversion methods in both languages (bd6f6fd73, b390939c1).

src/itc/interop.rs
duncaneddy: Handled now: an ITRF covariance frame is rotated into the state frame with `state_transform_jacobian(ITRF, state_frame, epoch)` per record (and back on export); round-trip tests cover ITRF and EME2000 covariance frames with EME2000 and TEME state frames (bd6f6fd73).

src/itc/interop.rs
duncaneddy: Handled now in `from_trajectory`: a trajectory in any Cartesian frame can be written with any covariance frame, rotated per sample through the router Jacobian, with the RTN basis always taken about the Earth (bd6f6fd73, b390939c1).

src/trajectories/traits.rs
duncaneddy: Replaced by `frames::covariance::state_transform_jacobian` in #557 (any router frame pair); nothing of it remains in trajectories or here.

src/itc/interop.rs
```diff
@@ -0,0 +1,941 @@
...
+/// * `DMatrix<f64>`: Dynamically-sized 6x6 matrix with the same entries
+///
+/// # Examples
+/// ```text
+/// dmatrix_from(&SMatrix6::identity()) produces a 6x6 DMatrix equal to
+/// the identity matrix.
+/// ```
+fn dmatrix_from(m: &SMatrix6) -> DMatrix<f64> {
+    DMatrix::from_iterator(6, 6, m.iter().cloned())
+}
```
duncaneddy: This seems like this could be a general utility in src::utils::

duncaneddy: Done. Moved to `brahe::utils::dmatrix_from_smatrix` (generic over the fixed size) with its inverse `smatrix_from_dmatrix`, and the ITC interop and trajectory code use the shared helpers.

PR #554, opened 2026-09-12
Add Starlink ephemeris client

duncaneddy:
# Pull Request

## Description

Adds `StarlinkClient`, which downloads Starlink's public Modified ITC ephemerides from `https://api.starlink.com/public-files/ephemerides/`. The site manifest drives every download: it is cached for an hour by default, refreshed with a conditional GET, and a satellite's file is fetched only when the manifest names one that is not already cached, after which the superseded files for that NORAD ID are deleted. Ephemeris files and the manifest are cached under `starlink/` for the public endpoint and under `starlink/mirrors/<host>-<sha256 prefix>/` for any other base URL, so mirrors never serve each other's data. Requests share the Space-Track style rate limiter (1000 per minute, 30000 per hour by default), retry on transient failures, and honour `BRAHE_NETWORK_MODE`. `StarlinkManifest` exposes the listing as a table with NORAD IDs, names, start and stop epochs decoded from the file names, change detection against the previous listing, and a polars `to_dataframe()`. `download_all` fetches every missing file on a bounded pool of threads; `save_ephemeris` and `save_all` place files in a directory or under a file name, moving them out of the cache by default or copying them when `keep_cached` is set. Python bindings release the GIL during network calls. Fifth PR of the Starlink ephemeris client stack.

### Added
- `StarlinkClient` (`brahe::starlink`, `bh.StarlinkClient`) with `get_manifest`, `refresh_manifest`, `cached_manifest`, `previous_manifest`, `download_ephemeris`, `get_ephemeris`, `get_trajectory`, `get_trajectory_with_covariance_variant`, `save_ephemeris`, `download_all`, `save_all`, `prune_cache`, and `cached_files`
- `StarlinkManifest` and `StarlinkManifestEntry` with lookups by NORAD ID and name, `changed_since`, and `to_dataframe`
- `get_starlink_cache_dir` and the `brahe.clients.starlink` namespace
- `time::conversions::days_in_month` (`bh.days_in_month`), shared by the epoch calendar arithmetic and the HTTP-date parser
- `utils::fs::move_file`, `copy_file`, `is_within`, and `validate_path_component`, shared file-management helpers used by the Starlink client and the Space-Track file-name parser; `save_ephemeris` and `save_all` reject any destination inside the cache tree before downloading

### Changed
- Celestrak and Starlink cache staleness share `utils::cache::is_older_than`, and mirror cache labels use `utils::cache::short_hash` (SHA-256)

### Fixed
- The shared `RateLimiter` queues concurrent reservations one slot apart instead of waking every blocked caller at the same time, so `download_all` workers and Space-Track callers cannot burst past the configured cap

## Note to Reviewers
Unit tests run against `httpmock` (Rust) and a stdlib HTTP server (Python) with the two committed Starlink files; the live integration tests are gated behind the `integration` feature and marker. Documentation pages, examples, and the cache-seeding script follow in the next PR.


src/clients/starlink/client.rs
```diff
@@ -0,0 +1,1945 @@
...
+    ///
+    /// ```
+    /// use brahe::starlink::StarlinkClient;
+    ///
+    /// let dir = StarlinkClient::new().cache_dir().unwrap();
+    /// assert!(dir.ends_with("starlink"));
+    /// ```
+    pub fn cache_dir(&self) -> Result<PathBuf, BraheError> {
+        get_starlink_cache_dir().map(PathBuf::from)
+    }
```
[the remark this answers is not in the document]
duncaneddy: Fixed in 367d56cf6: the cache directory is now namespaced by base URL. The public endpoint keeps `$BRAHE_CACHE/starlink`; any other `base_url` caches under `starlink/mirrors/<host>-<fnv1a-8hex>/`, so a mirror can never serve the public origin's manifest or files (Rust and Python tests cover distinct dirs, trailing-slash equivalence, and the default layout).

src/clients/starlink/client.rs
```diff
@@ -0,0 +1,1945 @@
...
+            .filter(|(_, path)| !path.exists())
+            .map(|(entry, _)| *entry)
+            .collect();
+        let cursor = AtomicUsize::new(0);
+        let stop = AtomicBool::new(false);
+        let failure: Mutex<Option<BraheError>> = Mutex::new(None);
+        std::thread::scope(|scope| {
+            for _ in 0..concurrency.min(pending.len()) {
+                scope.spawn(|| {
+                    loop {
```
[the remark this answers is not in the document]
duncaneddy: Confirmed and fixed in 04398174f. `RateLimiter::acquire` computed every waiter's delay from the oldest recorded request, so callers blocked on a full window all reserved the same wake time. It now waits for the entry `len - max` slots from the front (the one that must expire for the window to have room), so concurrent reservations queue one slot apart; `test_rate_limiter_queues_concurrent_reservations` pins 0/60/120 s for a cap of 1 and 0/0/60/60/120 s for a cap of 2.

src/clients/starlink/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+pub use client::StarlinkClient;
+pub use manifest::{StarlinkManifest, StarlinkManifestEntry};
+
+use crate::clients::RateLimitConfig;
+
+/// Default request limits for the Starlink mirror: 1000 per minute and 30000 per hour.
+pub const STARLINK_RATE_LIMIT: RateLimitConfig = RateLimitConfig {
+    max_per_minute: 1000,
+    max_per_hour: 30000,
+};
```
duncaneddy: Seems like something that should be defined in the starlink/client.rs since that's what uses them?

Not in top level module that is generally just imports of other implementations

src/clients/starlink/client.rs
```diff
@@ -0,0 +1,2141 @@
...
+        destination: P,
+    ) -> Result<PathBuf, BraheError> {
+        let source = self.download_ephemeris(norad_cat_id)?;
+        let name = source
+            .file_name()
+            .and_then(|n| n.to_str())
+            .unwrap_or_default()
+            .to_string();
+        let target = resolve_destination(destination.as_ref(), &name)?;
+        copy_file(&source, &target)?;
```
duncaneddy: I think this should move instead of copying the file, because otherwise users essentially have duplicate storage of the downloaded data (the cached copy and the destination copy), and they were likely thiking that the destination is the only variant.

So even if this ends up downloading more data, it would seem the right thing to do is not cache these "saves"

(unless there is a better way to handle it)

src/clients/starlink/client.rs
```diff
@@ -0,0 +1,2141 @@
...
+/// # Returns
+/// * `Ok(())`: The modification time was updated
+/// * `Err(BraheError)`: If the file cannot be opened or its time set
+fn touch(path: &Path) -> Result<(), BraheError> {
+    fs::OpenOptions::new()
+        .write(true)
+        .open(path)
+        .and_then(|f| f.set_modified(SystemTime::now()))
+        .map_err(|e| BraheError::IoError(format!("Failed to update {}: {e}", path.display())))
+}
```
duncaneddy: This seems like common functionality for clients/caching utils and could be moved there.

src/clients/starlink/client.rs
```diff
@@ -0,0 +1,2141 @@
...
+fn file_epoch(path: &Path) -> Result<Epoch, BraheError> {
+    let modified = fs::metadata(path)
+        .and_then(|m| m.modified())
+        .map_err(|e| BraheError::IoError(format!("Failed to read file modification time: {e}")))?;
+    let secs = modified
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .unwrap_or_default()
+        .as_secs_f64();
+    Ok(Epoch::from_unix_timestamp(secs))
+}
```
duncaneddy: This seems like common functionality for clients/caching utils and could be moved there.

src/clients/starlink/manifest.rs
```diff
@@ -0,0 +1,834 @@
...
+ */
+
+use polars::prelude::*;
+
+use crate::clients::spacetrack::{EphemerisFileCategory, EphemerisFileName};
+use crate::time::conversions::calendar_from_day_of_year;
+use crate::time::{Epoch, TimeSystem};
+use crate::utils::BraheError;
+
+const SECONDS_PER_DAY: f64 = 86400.0;
```
duncaneddy: This already exists as a constant in src/constants/time.rs

src/clients/starlink/manifest.rs
```diff
@@ -0,0 +1,834 @@
...
+        1 | 3 | 5 | 7 | 8 | 10 | 12 => 31,
+        4 | 6 | 9 | 11 => 30,
+        2 => {
+            let leap =
+                (year.is_multiple_of(4) && !year.is_multiple_of(100)) || year.is_multiple_of(400);
+            if leap { 29 } else { 28 }
+        }
+        _ => 0,
+    }
+}
```
duncaneddy: This is already partially implemented internally in src/time/epoch.rs but it's the same code, so might be better to have as a public time function in time/conversions.rs and reuse in both places?

src/clients/starlink/manifest.rs
```diff
@@ -0,0 +1,834 @@
...
+        year,
+        month,
+        day,
+        hour,
+        minute,
+        second as f64,
+        0.0,
+        TimeSystem::UTC,
+    ))
+}
```
duncaneddy: Seems like common utility for clients, not starlink-specific

src/clients/starlink/manifest.rs
```diff
@@ -0,0 +1,834 @@
...
+    if metadata.is_empty() || !metadata.chars().all(|c| c.is_ascii_digit()) {
+        return None;
+    }
+    let seconds: f64 = metadata.parse().ok()?;
+    if !seconds.is_finite() || seconds >= MAX_GPS_SECONDS {
+        return None;
+    }
+    let stop = Epoch::from_gps_seconds(seconds);
+    ((stop - reference).abs() <= GPS_STOP_WINDOW_SECONDS).then_some(stop)
+}
```
duncaneddy, line 134: Are we certain that the time in the metadata filename / file contents in in GPS time?

duncaneddy: Confident, from the files themselves: the fixture's metadata field 1473385380 decodes as 2026-09-14 01:43:00 GPS, which is 01:42:42 UTC (GPS−UTC = 18 s), and that is exactly the last record epoch and header stop of the file (2026-09-14 01:42:42.000 UTC). The live files I checked when designing this agreed the same way. Starlink does not document the field, which is why it is decoded into an `Option` behind a 30-day plausibility check rather than trusted unconditionally; the ephemeris start still comes from the normative day-time group in the name.

src/clients/starlink/client.rs
duncaneddy: Will do: `save_ephemeris`/`save_all` will move the file out of the cache (rename, with a copy-and-remove fallback across devices), so the destination becomes the only copy and the cache holds no ephemeris for that satellite until it is requested again. Docs and tests updated accordingly.

src/clients/starlink/mod.rs
duncaneddy: Moved into `client.rs` as `pub const STARLINK_RATE_LIMIT`; `mod.rs` only re-exports it (55459730e).

src/clients/starlink/client.rs
duncaneddy: Done in 6957ed3ea: `save_ephemeris`/`save_all` move the cached file (rename, with a staged copy-and-rename fallback across devices — bb6e34914), so the destination is the only copy and a later request re-downloads. Saving into the cache directory itself is rejected (including a bare file name from inside it — baa8d876f).

src/clients/starlink/client.rs
duncaneddy: Moved to `utils::fs::modified_epoch` / `utils::fs::touch` (55459730e).

src/clients/starlink/client.rs
duncaneddy: Moved to `utils::cache::is_older_than(path, max_age_seconds)`; the Celestrak client now calls it too (55459730e).

src/clients/starlink/manifest.rs
duncaneddy: Now uses `crate::constants::SECONDS_PER_DAY` (55459730e).

src/clients/starlink/manifest.rs
duncaneddy: Added `time::conversions::days_in_month(year, month)` (public, with a Python binding) and used it in the two `epoch.rs` tables, in `conversions.rs`, and in the HTTP-date parser (55459730e).

src/clients/starlink/manifest.rs
duncaneddy: Moved to `utils::network::parse_http_date` as a crate-wide helper (55459730e).

src/clients/spacetrack/file_name.rs
```diff
@@ -127,22 +127,60 @@ pub struct SpaceTrackEphemerisFileName {
...
+        )));
+    }
+    if value == "." || value == ".." {
+        return Err(BraheError::Error(format!(
+            "invalid ephemeris file name field {}: '{}' must not be '.' or '..'",
+            field, value
+        )));
+    }
+    Ok(())
+}
```
duncaneddy, line 144: This seems like it could be a generally useful function across the module and have a place in src::utils::* somewhere

src/clients/starlink/client.rs
```diff
@@ -0,0 +1,2232 @@
...
+    /// let path = StarlinkClient::new()
+    ///     .save_ephemeris(100001, "./ephemerides")
+    ///     .unwrap();
+    /// println!("{}", path.display());
+    /// ```
+    pub fn save_ephemeris<P: AsRef<Path>>(
+        &self,
+        norad_cat_id: u32,
+        destination: P,
+    ) -> Result<PathBuf, BraheError> {
```
duncaneddy, line 878: Instead of having the `save_*` functions decide between moving or copying could it instead be an argument to let the user decide to retain a copy in the cache?

src/clients/starlink/client.rs
```diff
@@ -0,0 +1,2232 @@
...
+    /// use brahe::starlink::StarlinkClient;
+    ///
+    /// let paths = StarlinkClient::new().save_all("./ephemerides", 8).unwrap();
+    /// println!("{} files saved", paths.len());
+    /// ```
+    pub fn save_all<P: AsRef<Path>>(
+        &self,
+        destination: P,
+        concurrency: usize,
+    ) -> Result<Vec<PathBuf>, BraheError> {
```
duncaneddy, line 1008: Save comment here for the copy/move decision

src/clients/starlink/client.rs
```diff
@@ -0,0 +1,2232 @@
...
+            })
+        }
+        Err(e) => Err(BraheError::IoError(format!(
+            "Failed to move {} to {}: {}",
+            source.display(),
+            target.display(),
+            e
+        ))),
+    }
+}
```
duncaneddy: This also seems like a generally useful file management utility and could be placed in src::utils::*

src/clients/starlink/client.rs
```diff
@@ -0,0 +1,2232 @@
...
+///
+/// # Returns
+/// * `u64`: Hash of `bytes`
+fn fnv1a_hash(bytes: &[u8]) -> u64 {
+    const FNV_OFFSET_BASIS: u64 = 0xcbf2_9ce4_8422_2325;
+    const FNV_PRIME: u64 = 0x0000_0100_0000_01b3;
+    bytes.iter().fold(FNV_OFFSET_BASIS, |hash, &b| {
+        (hash ^ u64::from(b)).wrapping_mul(FNV_PRIME)
+    })
+}
```
duncaneddy: Instead of implementing our own hasher can we use a standard hash implementaiton out of (https://github.com/RustCrypto/hashes)

src/clients/spacetrack/file_name.rs
duncaneddy: Done. Moved to `brahe::utils::fs::validate_path_component`; the Space-Track file-name parser calls it and keeps its field-specific error prefix.

src/clients/starlink/client.rs
duncaneddy: Done. `save_ephemeris(norad_cat_id, destination, keep_cached)` and `save_all(destination, concurrency, keep_cached)` now take the decision as an argument (`keep_cached=False` in Python). With it false the cached file is moved as before; with it true the file is copied and the cached copy stays, so later `download_*` calls do not refetch. Tests cover both modes in Rust and Python, and the docs and CLI follow in the docs PR.

src/clients/starlink/client.rs
duncaneddy: Done. Same `keep_cached` argument on `save_all`; the move-versus-copy choice is the caller's.

src/clients/starlink/client.rs
duncaneddy: Done. `move_file`, `copy_file` (staged copy, no partial file at the target), and `same_path` now live in `brahe::utils::fs` with their tests.

src/clients/starlink/client.rs
duncaneddy: Done. The FNV implementation is gone; the mirror label uses the existing `utils::cache::short_hash`, which is the leading 64 bits of a SHA-256 digest from the RustCrypto `sha2` crate the library already depends on.

Issue #547, opened 2026-09-09
13 paper.bib entries have a DOI available


goelakash:
13 entries with no `doi` field resolve on Crossref by exact title and first-author match, for example `eddyOptimal2024` -> 10.1109/aero63441.2025.11068558; `eddyMarkov2020` -> 10.1109/aero47225.2020.9172258; `stringham2019capella` -> 10.1109/igarss.2019.8900410. Happy to open a PR adding them.

Feel free to close if this is incorrect.

duncaneddy:
Thank you for raising the issue!

It'll be a while to get to this, so if you're able to open a PR happy to accept it.

Regardless, appreciate the diligence in checking, and will keep this open to track getting the DOIs added.

Many thanks!

duncaneddy:
Closing this out with the PR

PR #556, opened 2026-09-12
Add DOIs to paper.bib entries that have one registered

goelakash:
## Description

Adds `doi` to the 11 `joss/paper.bib` entries I could confirm on Crossref by exact title and first-author match.

### Changed
- `joss/paper.bib` references now carry a `doi` field where one is registered on Crossref (11 entries)

## Related Issue

Part of #547

## Note to Reviewers

I said 13 in #547 and can only stand up 11. I left out `kim2025scalable` and `newton1833philosophiae` as i couldn't find the right DOIs.


duncaneddy:
Thank you for getting these added—looks great!

PR #552, opened 2026-09-11
Carry covariance through trajectory frame conversion

duncaneddy:
# Pull Request

## Description

`OrbitTrajectory.to_frame` and its wrappers now carry stored covariance through conversions between frames that may hold it (GCRF and EME2000), rotating each matrix by the constant frame-bias Jacobian sized to that matrix, with an identity block for covariance rows beyond the orbital six. Targets that cannot carry covariance, and Keplerian-representation sources, drop it as before. Applies to both the dynamic and the static Rust trajectory types. Fourth PR of the Starlink ephemeris client stack; it lets an ITC-derived EME2000 trajectory keep its covariance after `to_eci()`.

### Changed
- `DOrbitTrajectory::to_frame` and `SOrbitTrajectory::to_frame` (and `to_eci`, `to_gcrf`, `to_eme2000`) rotate covariance between GCRF and EME2000 instead of dropping it; other target frames and Keplerian sources still drop covariance

## Note to Reviewers
State transition matrices, sensitivities and accelerations are still dropped on frame conversion; only covariance changes here.

PR #557, opened 2026-09-12
Route trajectory covariance through the frame router

duncaneddy:
# Pull Request

## Description

Trajectory covariance now travels with the frame router instead of being confined to GCRF and EME2000. `frames::covariance` builds the 6x6 Jacobian of the router's affine state transform between any two frames at an epoch (rotation, angular-velocity coupling for rotating frames, no dependence on the center translation), obtained numerically by probing the affine transform so it is exact up to rounding, and `OrbitTrajectory` uses it: covariance may be attached in any Cartesian frame, `to_frame` rotates it per sample for every frame pair, and `covariance_in_frame(frame, epoch)` evaluates it in any frame on demand. The RTN covariance Jacobians for the block-diagonal and rotating conventions live beside the RTN definition in `relative_motion`, replacing four hand-rolled copies (both trajectory types, the numerical propagator, and the Modified ITC conversion). Shared matrix helpers (`skew_symmetric`, `block_diagonal`, `symmetrize`, `is_symmetric`) move to `math`. Replaces #552 in the Starlink ephemeris client stack; the ITC interop PR is rebased on top of it.

### Added
- `brahe::frames::state_transform_jacobian`, `rotate_covariance`, `rotate_covariance_6`, `covariance_frame_to_frame`, and `validate_covariance_shape`, plus the batch forms `state_transform_jacobians`, `rotate_covariances`, and `covariances_frame_to_frame`; the Python `bh.state_transform_jacobian`, `bh.rotate_covariance`, and `bh.covariance_frame_to_frame` dispatch on input shape like `state_frame_to_frame`
- `brahe::relative_motion::jacobian_rtn_to_eci`, `jacobian_eci_to_rtn`, `covariance_rtn_to_eci`, and `covariance_eci_to_rtn` with an `OrbitRelativeFrameVariant` argument (Python mirrors)
- `OrbitTrajectory.covariance_in_frame(frame, epoch)`, `covariance_itrf`, `covariance_ecef`, `covariance_eme2000`, and `covariance_rtn_with_variant(epoch, variant)` on both trajectory types
- `brahe::math::skew_symmetric`, `block_diagonal`, `symmetrize`, `symmetrize_6`, and `is_symmetric`

### Changed
- `OrbitTrajectory::from_orbital_data` and the covariance setters accept covariance in any Cartesian frame; `to_frame`, `to_eci`, `to_gcrf`, `to_eme2000`, `to_itrf`, and `to_ecef` rotate covariance for every frame pair, and a conversion to the same frame keeps it
- `covariance_rtn` and `covariance_rtn_with_variant` on both trajectory types build the RTN frame about the trajectory's own central body from `state_bci`, so lunar and planetary trajectories get a body-centred RTN; Earth-orbit results are unchanged, and `DNumericalOrbitPropagator::covariance_rtn` is computed through the same shared Jacobian and symmetrized
- `frames::covariance` is the public covariance module (`brahe::frames::*` exports it); `math::covariance` is private with its items re-exported through `math::*`
- Keplerian-representation trajectories reject covariance rotation and `covariance_rtn` with a single error instead of silently applying a Cartesian Jacobian

### Removed
- The GCRF/EME2000-only covariance rule and its helpers (`covariance_frame_allowed`, `inertial_covariance_rotation`, `frame_covariance_jacobian`, `frame_covariance_jacobian_6`)

## Note to Reviewers
The Jacobian is exact for the router's affine transforms; the only approximation in the tests is comparing the GCRF-to-ITRF coupling block against the constant Earth rate, which differs from the router's polar-motion-rotated rate at the 1e-6 level. Python cannot yet build an extended-dimension trajectory with covariance, so that case is tested in Rust only. `DNumericalOrbitPropagator::covariance_rtn` still takes the Earth-centred state and is left for a follow-up.


tests/trajectories/test_orbit_trajectory.py
```diff
@@ -2840,48 +2840,23 @@ def test_from_orbital_data_covariances_length_mismatch(eop):
     assert "must match" in str(exc_info.value).lower()
 
 
-def test_from_orbital_data_covariances_invalid_frame_ecef(eop):
-    """Test that covariances with ECEF frame raise error."""
+def test_from_orbital_data_covariances_accept_earth_fixed_frames(eop):
+    """Rust: test_dorbittrajectory_from_orbital_data_cov_accepts_earth_fixed_frame"""
```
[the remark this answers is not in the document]
duncaneddy: The behaviour change is the maintainer's request: the review on #551 (thread on `traits.rs:192`) asked to stop confining covariance to GCRF/EME2000 and route it like states, and the direction was confirmed before this PR was written. The rejection tests therefore became acceptance tests deliberately, with anisotropic covariances so the rotation is actually checked.

src/frames/covariance.rs
```diff
@@ -0,0 +1,500 @@
...
+/// let j = SMatrix6::identity();
+/// let rotated = rotate_covariance(&p, &j).unwrap();
+///
+/// assert_eq!(rotated.nrows(), 7);
+/// assert_eq!(rotated[(6, 6)], 1.0);
+/// ```
+pub fn rotate_covariance(
+    covariance: &DMatrix<f64>,
+    jacobian: &SMatrix6,
+) -> Result<DMatrix<f64>, BraheError> {
```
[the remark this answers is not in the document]
duncaneddy: Bound in d0c84d4f0: `bh.rotate_covariance` (covers the 6x6 form), `bh.skew_symmetric`, `bh.block_diagonal`, `bh.symmetrize`, `bh.is_symmetric`, exported from `brahe.frames`/`brahe.math` with tests mirroring the Rust ones.

crates/brahe-py/src/trajectories.rs
```diff
@@ -2510,6 +2533,181 @@ impl PyOrbitalTrajectory {
...
+    /// Args:
+    ///     epoch (Epoch): Time at which to retrieve the covariance
+    ///
+    /// Returns:
+    ///     np.ndarray: Covariance matrix in ITRF
+    ///
+    /// Raises:
+    ///     RuntimeError: If the covariance is unavailable or cannot be rotated
+    #[pyo3(text_signature = "(epoch)")]
+    fn covariance_itrf<'py>(
```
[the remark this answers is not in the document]
duncaneddy: Added `Example` sections to `covariance_itrf`, `covariance_ecef` and `covariance_eme2000` in d0c84d4f0.

crates/brahe-py/src/frames.rs
```diff
@@ -5767,6 +5767,154 @@ fn py_state_frame_to_frame<'py>(
...
+    let from = extract_frame(from_frame)?;
+    let to = extract_frame(to_frame)?;
+    let array = covariance.as_array();
+    let (rows, cols) = (array.shape()[0], array.shape()[1]);
+    let p = DMatrix::<f64>::from_row_iterator(rows, cols, array.iter().copied());
+    let rotated = frames::covariance_frame_to_frame(from, to, epc.obj, &p)?;
+    let rotated_ref = &rotated;
+    let (n, m) = (rotated.nrows(), rotated.ncols());
+    Ok(matrix_to_numpy!(py, rotated_ref, n, m, f64).to_owned())
+}
```
duncaneddy, line 5970: This doesn't appear to support vectorized invocations in the same way that the `state_frame_to_frame` frunction does

src/frames/covariance.rs
```diff
@@ -0,0 +1,500 @@
...
+///
+/// Only the two orientations matter. A change of center is a translation,
+/// whose Jacobian is the identity whatever the offset, and the router
+/// evaluates a given orientation identically at every center. Between two
+/// celestial frames the probe therefore never crosses centers: identical axes
+/// return the identity outright, and otherwise the target axes are probed
+/// about the source's own center. A covariance query between, say, `LCI` and
+/// `GCRF` consults no ephemeris and needs no SPK coverage. Body and
+/// orbit-relative frames are probed as given, since their orientation chains
+/// are resolved through the frame registry rather than by axes alone.
```
duncaneddy, line 73: Comment should mention that the Jacobian is computed numercially

src/frames/covariance.rs
```diff
@@ -0,0 +1,500 @@
...
+use crate::math::linalg::{SMatrix6, SVector6};
+use crate::time::Epoch;
+use crate::utils::errors::BraheError;
+
+/// Jacobian of the state transform from `from` to `to` at `epoch`.
+///
+/// The router's state transform is affine, so the Jacobian is recovered
+/// exactly (to rounding) by transforming the origin and six scaled basis
+/// states and differencing: position probes of `1e7` m and velocity probes
+/// of `1e4` m/s keep the differences well above the rounding of any
```
duncaneddy: These probe values seem extremely large, do they give stable (correct) approximations of the jacobian across supported states/frames?

src/frames/covariance.rs
```diff
@@ -0,0 +1,500 @@
...
+///
+/// // Frames that share their axes differ only by a translation.
+/// let j = state_transform_jacobian(CelestialFrame::LCI, CelestialFrame::GCRF, epc).unwrap();
+/// assert_eq!(j, brahe::math::linalg::SMatrix6::identity());
+/// ```
+pub fn state_transform_jacobian(
+    from: impl Into<ReferenceFrame>,
+    to: impl Into<ReferenceFrame>,
+    epoch: Epoch,
+) -> Result<SMatrix6, BraheError> {
```
duncaneddy, line 107: This seems like it should have vectorization support similar to other frame transformation functions

src/frames/covariance.rs
```diff
@@ -0,0 +1,500 @@
...
+            "covariance must be at least 6x6, got {}x{}",
+            rows, cols
+        )));
+    }
+
+    let mut j = DMatrix::<f64>::identity(rows, rows);
+    j.view_mut((0, 0), (6, 6)).copy_from(jacobian);
+
+    Ok(symmetrize(&(&j * covariance * j.transpose())))
+}
```
duncaneddy, line 260: Also seems like we should have vectorization support for these capabilities

src/frames/covariance.rs
```diff
@@ -0,0 +1,500 @@
...
+/// ```
+pub fn covariance_frame_to_frame(
+    from: impl Into<ReferenceFrame>,
+    to: impl Into<ReferenceFrame>,
+    epoch: Epoch,
+    covariance: &DMatrix<f64>,
+) -> Result<DMatrix<f64>, BraheError> {
+    let jacobian = state_transform_jacobian(from, to, epoch)?;
+    rotate_covariance(covariance, &jacobian)
+}
```
duncaneddy, line 366: Similar vectorization comment

src/frames/mod.rs
```diff
@@ -8,10 +8,19 @@
...
+ * [`covariance_frame_to_frame`], which rotate a state covariance between any
+ * two frames the router can transform states between.
  */
 
 pub mod axes;
 pub mod center;
+// Not `pub`: `crate::math::covariance` is already a public module of that
+// name, and `pub mod covariance;` here would make `pub use frames::*;` (in
+// `lib.rs`) collide with `pub use math::*;` on the module name itself. The
+// glob re-export below still surfaces every public item.
```
duncaneddy: It seems like frame covariance should be the more imporant one for brahe so that pub use frames::* should grab covariance and instead of crate::math::covariance not be pub

src/relative_motion/eci_rtn.rs
```diff
@@ -238,6 +239,195 @@ pub fn state_rtn_to_eci(x_chief: SVector6, x_rel_rtn: SVector6) -> SVector6 {
...
+/// ```
+///
+/// # References:
+/// 1. NASA Conjunction Assessment Risk Analysis (CARA), *Conjunction Assessment Handbook*,
+///    NASA/SP-20205011318, Appendix N (RIC-to-ECI covariance transformation, eq. N-13),
+///    <https://ntrs.nasa.gov/citations/20205011318>
+/// 2. NASA CARA Analysis Tools, `RIC2ECI.m`, <https://github.com/nasa/CARA_Analysis_Tools>
+/// 3. D. A. Vallado, "Covariance Transformations for Satellite Flight Dynamics
+///    Operations," AAS 03-526, AAS/AIAA Astrodynamics Specialist Conference, 2003,
+///    <https://celestrak.org/publications/AAS/03-526/AAS-03-526.pdf>
```
duncaneddy: I believe links should be in markdown format/style

src/trajectories/dorbit_trajectory.rs
```diff
@@ -1880,15 +1881,22 @@ impl DOrbitTrajectory {
...
     /// * `representation` - State representation (Cartesian or Keplerian)
     /// * `angle_format` - Angle format (None for Cartesian, Radians/Degrees for Keplerian)
-    /// * `covariances` - Optional vector of 6x6 covariance matrices corresponding to states
+    /// * `covariances` - Optional vector of covariance matrices, one per state
+    ///
+    /// Covariance is accepted in any frame. It is stored exactly as given and
+    /// rotated on demand by [`Self::covariance_in_frame`] and
+    /// [`Self::to_frame`], which require a Cartesian representation and a
+    /// square matrix of at least 6x6 whose leading six elements are the
+    /// Cartesian state.
```
duncaneddy: Commentary should be above `Arguments` section.

src/trajectories/dorbit_trajectory.rs
```diff
@@ -2507,107 +2526,188 @@ impl DOrbitStateProvider for DOrbitTrajectory {
...
-        } else {
-            Err(BraheError::Error(format!(
-                "covariance transformation from {} is not supported (requires time-dependent rotation derivatives)",
-                self.frame
-            )))
+        if self.frame == frame {
+            return Ok(cov_native);
         }
+        require_cartesian_covariance(self.representation)?;
+        covariance_frame_to_frame(self.frame.clone(), frame, epoch, &cov_native)
```
duncaneddy, line 2589: Does this self.frame need to be cloned? since the borrow or self-frame should be immediately returned?

src/trajectories/dorbit_trajectory.rs
```diff
@@ -2507,107 +2526,188 @@ impl DOrbitStateProvider for DOrbitTrajectory {
...
+    /// ```
+    pub fn covariance_rtn_with_variant(
+        &self,
+        epoch: Epoch,
+        variant: OrbitRelativeFrameVariant,
+    ) -> Result<DMatrix<f64>, BraheError> {
+        require_cartesian_covariance(self.representation)?;
+        let cov_eci = self.covariance_in_frame(CelestialFrame::GCRF, epoch)?;
+        let state_eci = self.state_eci(epoch)?;
+        rotate_covariance(&cov_eci, &jacobian_eci_to_rtn(state_eci, variant))
```
duncaneddy: I think this might be overly restrictive since it always assumes the base state is eci. The user might want to get an RTN covariance about whatever the central body interial frame is (e.g. lunar, mars, RTN frame).

This should be possible using the `state_bci` function to just get the current body-centered interial state and then compose that with the RTN transformation

src/trajectories/dorbit_trajectory.rs
duncaneddy: The clone is needed because the router API (`state_frame_to_frame` and friends) takes frames by value as `impl Into<ReferenceFrame>`, and `covariance_frame_to_frame` follows the same signature; `state_in_frame` does the identical clone at line ~2462 for the same reason. `ReferenceFrame` is a small enum so the clone is cheap. If you would rather the covariance helpers take `&ReferenceFrame` I can add that overload, but it would make them the odd ones out next to the router.

crates/brahe-py/src/frames.rs
duncaneddy: Done. `rotate_covariance`, `covariance_frame_to_frame` and `state_transform_jacobian` now dispatch on input shape like `state_frame_to_frame`: a `(n, n)` covariance with a single epoch or `(6, 6)` Jacobian returns `(n, n)`; a `(k, n, n)` covariance, a sequence of epochs, or a `(k, 6, 6)` Jacobian returns `(k, n, n)` (single-with-batch broadcasts). A batch of one stays a batch, and an empty batch keeps its trailing shape and is still shape-checked. Tests added in `tests/test_frames.py`.

src/frames/covariance.rs
duncaneddy: Done. The docs now state that the Jacobian is obtained numerically by probing the frame transform at the origin and six displaced states, and why that is exact for the affine transforms the router applies (the finite difference of an affine map is its linear part, so the probe scale only sets rounding, not truncation, error).

src/frames/covariance.rs
duncaneddy: Checked, and yes. Every hop the router applies (rotation, rotation rate, origin translation) is affine in the state, so the difference quotient recovers the linear part exactly regardless of probe size; the scale only affects rounding. Measured on GCRF→ITRF the unit-scale and production-scale probes agree to ‖ΔJ‖ ≈ 2e-16, and across a heliocentric offset the large probes hold the columns to a few parts in 1e12 where unit probes lose four more digits. Same-axes pairs return the identity without probing, and celestial pairs are probed on a common center so no SPK lookup enters the Jacobian. The rationale and the measured bound are now in the `frames::covariance` module docs, with a test that pins the scale independence.

src/frames/covariance.rs
duncaneddy: Done. Added the plural batch forms: `state_transform_jacobians(from, to, &[Epoch])`, `rotate_covariances(&[DMatrix], &[SMatrix6])`, and `covariances_frame_to_frame(from, to, &[Epoch], &[DMatrix])`, following the `states_frame_to_frame` convention (single Python name dispatching on shape).

src/frames/covariance.rs
duncaneddy: Done. Added the plural batch forms: `state_transform_jacobians(from, to, &[Epoch])`, `rotate_covariances(&[DMatrix], &[SMatrix6])`, and `covariances_frame_to_frame(from, to, &[Epoch], &[DMatrix])`, following the `states_frame_to_frame` convention (single Python name dispatching on shape).

src/frames/covariance.rs
duncaneddy: Done. Added the plural batch forms: `state_transform_jacobians(from, to, &[Epoch])`, `rotate_covariances(&[DMatrix], &[SMatrix6])`, and `covariances_frame_to_frame(from, to, &[Epoch], &[DMatrix])`, following the `states_frame_to_frame` convention (single Python name dispatching on shape).

src/frames/mod.rs
duncaneddy: Done. `frames::covariance` is now the public module (`pub use frames::*` exports it) and `math::covariance` is private with its items re-exported through `math::*`. The one example that still reached the covariance constructors through `math::covariance::` was updated to `math::`.

src/relative_motion/eci_rtn.rs
duncaneddy: Done. All reference citations in `eci_rtn.rs`, `frames::covariance` and the trajectory covariance methods are now markdown links.

src/trajectories/dorbit_trajectory.rs
duncaneddy: Done. The commentary now sits above the `# Arguments` section in all affected docstrings.

src/trajectories/dorbit_trajectory.rs
duncaneddy: Done. `covariance_rtn` and `covariance_rtn_with_variant` now build the RTN frame about the trajectory's own central body: the state comes from `state_bci` in the ICRF-aligned inertial frame of the trajectory's celestial root (Earth, Moon, Mars, ...) and the covariance is routed to that same frame before the RTN rotation. Earth-orbit results are unchanged; a lunar trajectory now gets a Moon-centred RTN. Tests cover an LCI trajectory in Rust and Python with an anisotropic covariance so the discrimination from the old Earth-centred construction is on the axes. `DNumericalOrbitPropagator::covariance_rtn` still assumes Earth and is noted as a follow-up.

